PR #18, opened 2017-06-06
Changing the min date from 50 when entering birthday year to 99 which…

blindly:
… is set for the standalone age.


djaiss:
Thanks so much for this PR. Unfortunately I already submitted a fix for that. But thanks, I really appreciate your effort.

PR #47, opened 2017-06-06
45 - set title for the profile pages to name of person you are looking at

ppanthony:
This fix allows each view to set a title variable. In this instance the title is set to the name of the profile you are currently looking at.


djaiss:
This is amazing. Thanks @ppanthony !!

Issue #52, opened 2017-06-06
Country Selection in Personal Information is broken


bimmlerd:
Country dropdown has 17010 (!) options, with lots of duplicates in the list.
This affects production, and is caused by the county seeding not truncating the table beforehand. PR created, untested but looks trivial.

djaiss:
Fixed by the PR.

PR #53, opened 2017-06-06
Truncate countries table before seeding

bimmlerd:
Because we reseed it, and truncate other tables.
This is likely the cause of #52


djaiss:
Thanks. Definitely makes sense.

PR #59, opened 2017-06-06
Remove ending periods

sikanderiqbal:
Remove ending periods to make data consistent with the rest of the language entries


djaiss:
Thanks so much for this, but I can't merge this as is. There are some sentences here and there that needs a "." at the end. I agree that perhaps for success messages we don't need them, or errors messages - but for sentences that appear for instance on blank pages, we need a dot at the end. Don't you agree?

sikanderiqbal:
@djaiss I suppose it should be consistent. Which do you prefer?

djaiss:
That's the thing. It should be consistent with the same nature of message.

* For success/errors: no dots
* For sentences that appear on blank states, or sentences for instances that are help text, a dot is necessary to end the sentence.

sikanderiqbal:
OK, let's leave as is then. I'll remove the PR. Cheers!

PR #60, opened 2017-06-06
Add import CSV Command for GMail export

mkaz:
Adds a Laravel Command to import a Google CSV file, one exported from GMail.

Usage: `php artisan import:csv USER FILE`

You will need to look up your user id and import, if you create a new account it is most likely 3. So for me I created a new user and downloaded the file as `google.csv` so the command is:

`php artisan import:csv 3 google.csv`

I also recommend merging any duplicates on Google/Gmail because there are data constraints on duplicate emails.


mkaz:
There was an issue if data wasn't complete, I fixed it so it won't set as empty strings so the database will get NULLs as expected.

djaiss:
Thanks so much @mkaz 

Can you indicate here how to get this CSV from Google? I'll update the README.

mkaz:
@djaiss - sure thing, I was going to make an animated GIF Screenshot, but too difficult with obscuring data. 

1. First go to Contacts by using the Mail drop down on the top left 

gmail-export-0

2. Next go More and select Export

gmail-export-1

3. Select what list you want to export and the Google CSV format

gmail-export-2

That will save the file to your system, likely in ~/Downloads/google.csv

You can then run it on the command-line using `php artisan import:csv 3 google.csv`

Replace 3 with your user id for the account you want to import to, I just peeked in the database to get this.

djaiss:
Thanks so much. I'll take of it tomorrow - it's too late for me now after this incredible day, but I truly appreciate the help here!

mkaz:
@djaiss - great tool, thanks for building and open sourcing it. I'm glad to help out. 
I'm adding a couple of extra checks so good it'll be tomorrow.
Cheers!

app/Console/Commands/ImportCSV.php
```diff
@@ -0,0 +1,121 @@
...
+                $row++;
+                $has_data = false;
+
+                // don't import the columns
+                if ( $row == 1 ) { continue; }
+
+                $contact = new Contact();
+                $contact->account_id = $user->id;
+
+                if ( ! empty( $data[1] ) ) {
```
djaiss: I think the first check should be to verify that either `$data[0]` or `$data[1]` contains a value (Name or given name) because these are the only data that we need to import. If we don't have one of these two values, we can't import the row, period (a contact without a name won't work in the current UI).
Can you make this check? Thanks a lot!

app/Console/Commands/ImportCSV.php
```diff
@@ -0,0 +1,121 @@
...
+                    $contact->last_name = $data[3];     // Family Name
+                    $has_data = true;
+                }
+
+                if ( ! empty( $data[14] ) ) {
+                    $contact->birthdate = date('Y-m-d', strtotime($data[14]));
+                }
+
+                // gender required - default to female
+                $contact->gender = ( substr($data[15], 0, 1 ) == 'm' ) ? 'male' : 'female';
```
djaiss: I've changed that today. You can now default to `none`.

djaiss:
Added some comments.

mkaz:
@djaiss updated with requests :+1:

djaiss:
@mkaz thanks!

I've tried it but all the contacts that I import are empty.

Instead of blank I should at least get a firstname... Any idea why? I took the CSV from Google Contact using the instructions you've provided.

image

erdmenchen:
Why do we need to search for a users id if we could also use the email address to identify a user for linking the imported data?
Just make the command usabe on the first run...dont make me search for my id...its easier to remember my email address...

yamakadi:
Wouldn't it be better to have a way to import and export vcards instead of limiting it to just google contacts? Not all providers use the same extended syntax for non-standard fields (such as custom labels), but standard fields shouldn't be an issue.

mkaz:
This was a quick and simple way to import some data in a format that I had it in - if it works for you then great. Ideally, an importer that can handle any format would be great with a front-end and not having to look up users or use the command-line.

erdmenchen:
Then why not trigger this import via the UI? 
For example inside the settings area.
And get the user id from the logged in user...

djaiss:
@mkaz do you know why I got these errors?

mkaz:
@djaiss I double checked my process and all worked fine for me. I'm not sure if Google exports them in different orders and uses the header to determine which is which. My implementation assumed that it was the same fixed order for every export, which would make sense.

If you import your csv file into Excel you can see which column maps to which field

PR #65, opened 2017-06-07
Remove number type from phone edit field

mkaz:
Phone numbers tend to be more than just numbers, they can include things like
ext 1234 for extensions, or dashes, hypens, parens to group numbers.
Also +1 for country code, etc...

Best to treat like a string


djaiss:

PR #66, opened 2017-06-07
Add Gravatar for Contact photo if avatar does not exist

mkaz:
Uses Gravatar https://en.gravatar.com/ for profile photo if an existing
avatar does not already exist and the user has an email address. One
thing is that it does not check Gravatar if it exists there, so if email
exists and no Gravatar, the default becomes the Gravatar default and not
the colored initials.

It is possible to check if the Gravatar exists by passing ?d=404 and
querying the headers to see if it returns a 404, but would be an extra
call on each page load, which isn't horrible since only one image per
page.


djaiss:
Thanks a lot! I've added a comment.

mkaz:
@djaiss updated with changes requested :+1:

djaiss:

PR #70, opened 2017-06-07
Make edit button more noticeable on Contact profile page

### LINKED ISSUE #46: Edit button is not obvious
edit

The edit button blends into the person page a bit too well. It was difficult to find. I suggest add a way to edit in the header perhaps with an icon ?

mkaz:
Adds class="btn" to the edit link which makes it larger and more
noticeable. I also added some spacing above. I tried to use the
bootstrap mt-5 class to add margin-top but doesn't seem to be there. I
don't have much bootstrap experience so not quite sure, so I simply
inlined the style. :-)

Fixes #46

### DIFF
--- a/resources/views/people/dashboard/people-information/index.blade.php
+++ b/resources/views/people/dashboard/people-information/index.blade.php
@@ -48,7 +48,7 @@
       </li>
     </ul>
 
-    <a href="/people/{{ $contact->id }}/edit" class="action-link">{{ trans('app.edit') }}</a>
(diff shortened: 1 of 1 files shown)


Kovah:
You may add the `mt-2` class to the edit button instead of the inlined CSS if it's available. `mt-2` is a helper class which adds a small margin to the top, see https://v4-alpha.getbootstrap.com/utilities/spacing/

djaiss:
Thanks a lot! However I had something else in mind for this. Let me open a new PR for this.

PR #78, opened 2017-06-07
Use the right variable

GhiMax:


djaiss:
Thanks. This was already fixed, but thanks a lot anyway for your effort!

Issue #56, opened 2017-06-06
Let the user disable new registrations


Kovah:
I would highly appreciate an option / setting / config parameter to completetly turn off new registrations. At the moment it seems that everyone that has access to the instance I set up can create an account and start adding contacts.
Even if the app separates all the data saved for users I would feel uncomfortable to know _anybody_ could use my app. Possible problem: security issues and access to an other users data via exploits. Not cool.

Such an option would just disable the register controller and remove the sign up button from the login page.

ocdtrekkie:
Duplicate of #16 and also fixed by https://github.com/monicahq/monica/commit/e9a776ca285f78dadd28e1c2a73a250fc61f03af

ocdtrekkie:
Oops. Commented on the wrong issue. :/

Kovah:
Pull request on the way.

degan6:
Maybe allow certain domains to create accounts or a whitelist of emails?

Kovah:
I just added a setting to completely disable the signup. Adding whitelists and only allow certain domains could be added later, however @djaiss should point out how to store these global configurations. The .env file may not be that practical to store lists with multiple entries.

PR #81, opened 2017-06-07
Let user disable the signup (#56)

Kovah:
Updates the RegisterController to check if the signup is disabled.
Removes signup link if needed.
Also adds a new generic 403 error page.

Solves #56


djaiss:
This is amazing.

Issue #84, opened 2017-06-07
Adding a debt bug


djaiss:
Adding a debt with a contact will always say "You owe $x" regardless if you select "You owe contact" or "Contact owes you"

(source: https://news.ycombinator.com/item?id=14504588)

PR #86, opened 2017-06-07
use correct translation if you're owed some debt

LePhil:
Should fix https://github.com/monicahq/monica/issues/84


djaiss:
Thanks a lot for this!

PR #87, opened 2017-06-07
Fix wrong word tense

vmitchell85:
The past tense should be "hung" not "hanged"


djaiss:
Thanks for fixing these kind of mistakes 👍

djaiss:
@vmitchell85 however if the names of the keys themselves change, we need to update the views that used those names as well.

vmitchell85:
@djaiss I searched the code base for "hanged" and didn't see any views... I'll check again...

vmitchell85:
Also... I caught the typo from the website screenshot: https://github.com/monicahq/monicahq.github.io/blob/master/assets/img/homepage-activities.png

Might want to take a new one...

djaiss:
@vmitchell85 you are right actually. This key is used programatically when building a view - this is why we don't find a view that contains it.

djaiss:
Thanks a lot!

However if we change the name of the keys, we'd need to reseed the production database, which I really don't want to do right now :-D

Can we simply change the translation and *not the key name themselves*? Thanks a lot!

vmitchell85:
I would suggest creating a new migration to fix the issue. Mainly because you're allowing people to self-host it. I can create the migration if you'd like.

vmitchell85:
Actually since it's a seeder and not a migration I wouldn't do a migration. I think the code base should be updated, but you can choose what to do with your hosted platform.

Since the activity types are a major part of the system I might have made them migrations.

djaiss:
@vmitchell85 I understand. The problem is for existing installations. If we do change the name without changing the DB field, it will create a bug. We need to run a migration no matter what to update the key.

vmitchell85:
@djaiss 

Just to make sure I understand. Even though new installs of Monica do not require the migration you would like a migration created to fix the text? I also just found another issue that the word "his" is used even when referencing a female (https://github.com/monicahq/monica/blob/master/database/seeds/ActivityTypesTableSeeder.php#L76).

Personally I think the best course of action would be to not add a migration to the github repo, but create a gist with instructions & code on fixing the database and add it to a special section of the readme

djaiss:
I'm ashamed of the time it took me to review it properly. Thanks a lot for your contribution.

PR #89, opened 2017-06-07
Shorten 'purpose' section

raboof:
"It's like an assistant (some call it spouse) you dream you had." might rub some people the wrong way, and the purpose is clearer and more to-the-point without it anyway.

(feel free to simply close this if you disagree, just my 2 cents ;) )


djaiss:
You are right. Thanks a lot!

PR #105, opened 2017-06-07
0 kid/reminder -> 0 kids/reminders

Qix-:
Zero of a collection still makes the following collection pluralized.


djaiss:
Thanks a lot!

PR #115, opened 2017-06-07
Add `Rather not say` gender

### LINKED ISSUE #44: Gender should be optional/allow other values
The gender field seems to be required on records wherever it is present, and only has male/female choices. This forces me to enter bad data where I don't know someone's gender or they identify as some other gender.

Ideally it would be nice to allow other values and 'unknown', but at a minimum making that field optional would at least avoid forcing the user to include known bad data.

(Given the subject matter, I really hope this doesn't become a flamewar...please feel free to lock/delete if it is in danger of doing so. And for the record, and perhaps in hope of averting that, I'll say that this is an *actual* issue that came up for me when trying the app with online contacts where I don't even know their gender for certain, and some friends who specifically identify as some non-binary option, not an "improve the world with social justice issue.")

djaiss:
People have complained that they want to add more genders than Male/Female. It makes sense. The quick workaround for this is to add a `rather not say` choice everywhere we mention the gender for people, which would be selected by default. In the future we'll be able to add more genders if needed.

This fixes #44

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -870,16 +870,16 @@
         },
         {
             "name": "jenssegers/date",
-            "version": "v3.2.10",
+            "version": "v3.2.11",
             "source": {
                 "type": "git",
                 "url": "https://github.com/jenssegers/date.git",
-                "reference": "ddc073e6cbba95bf298bf2bf43f2ebb2ec47b050"
+                "reference": "68262b137a4811cc351b3c26308e252505606b8f"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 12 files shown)


djaiss:
Warning: this change requires a migration.

martimarkov:
This is a breaking change for PostgreSQL. Please do not use RAW queries as they break migrations and have unforeseen consequences.

martimarkov:
Fixed in #247. Checked in PostgreSQL. No data loss.

Issue #97, opened 2017-06-07
Bug: htaccess file is missing in the public folder


erdmenchen:
Vanilla Laravel ships with a predefined htaccess file in the public folder.
It is however missing in this repository.

PR #121, opened 2017-06-07
Add .htaccess

erdmenchen:
See #97.
Just added the .htaccess file as it is provided with laravel.


enko:
I would vote against this PR, because this is specific to apache and apache is not so wide spread any more that it justifies to put this file here. I think it would be a better approach to put this into some installation instructions.

erdmenchen:
Why? Installation is already complicated enough and not well documented. 
Since this file is just shipped with Laravel, which this project is based on, I cannot understand why it was removed in the first place.
Also web.config is present, too.

See source: https://github.com/laravel/laravel/blob/master/public/.htaccess

peterbrinck:
I must agree with @erdmenchen 

It's a weird thing to leave out `.htaccess`.
Apache is STILL wide spread and used on many servers.

djaiss:
@erdmenchen honestly I don't know why it was removed. I don't remember - perhaps when it was hosted on Heroku? ¯\_(ツ)_/¯

adrum:
@enko What are the negative side effects of including this? I think it’s a no brainer and just adds to the number of servers this project can deploy on. I created another PR for this because I didn’t see this original one and thought it’s a pretty trivial thing to add back into the project.

djaiss:
Before I merge this (because I agree that Apache should be equally supported), as I run Nginx in production, do you think it's possible that it causes problem, and should I do something about this in production - or won't it have any impact in production whatsoever?

adrum:
I think the side effect would be other server types (nginx) would server the .htaccess file if someone requested it. We would add a change in the nginx config to deny access to hidden files, or just .htaccess if needed. Unless I'm completely missing something.

erdmenchen:
@djaiss  the Laravel framework has the same files in place. so there is no need to fear any problems...

djaiss:
Thanks for this PR!

Bibabidu:
Hey, I'm not sure this is the right place to discuss my question but I've encountered a problem related to the .htaccess-file. My redirect didn't work properly so I tinkered a bit with the rule. The following configuration works for me:

Original: `RewriteRule ^ index.php [L]`
New Rule: `RewriteRule ^(.*)$ /index.php?/$1 [L]`

Can someone, with a better understanding of .htaccess, tell me, if this is an unique issue only faced by me? Can my fix possibly cause any problems with using monica?

Thanks a lot already!

asbiin:
@Bibabidu this is the default .htaccess file for all Laravel sites. See https://github.com/laravel/laravel/blob/master/public/.htaccess

PR #134, opened 2017-06-08
Fix wrong Timezone settings (#132)

### LINKED ISSUE #132: Timezone: Can't select Berlin/Bern
The code concerning the issue can be found here: https://github.com/monicahq/monica/blob/2e6760e7dbcccd5375707ce1b7ce9c099f40f4b7/resources/views/settings/index.blade.php#L85

When updating the timezone to *(UTC+01:00) Berlin* on the settings page, it selects *(UTC+01:00) Bern* after saving. In the code, both Berlin and Bern have the value *Europe/Berlin*.

~~I'd create a PR for that, but I'm not sure that this line is everything that needs to be changed.~~

Also: Why is Paris so far up in the select list?

looperhacks:
Fixes #132 

Includes:
 - Berlin/Bern Problem (#132)
 - Removal of doubled entry for paris
 - Weird space für UTC

I'm not sure whether this PR includes everything that's needed, but I didn't find any other mentions of these timezones in the code.


djaiss:

PR #138, opened 2017-06-08
remove n+1 problem when getting a users timezone for multiple contacts

Lovatt:
Removes the n+1 problem when calling getLastUpdated() on a contact in the contacts collection in resources/views/people/index.blade.php


djaiss:
From 75 queries to 3. Awesome @Lovatt Thanks a lot!

Issue #137, opened 2017-06-08
Postgresql compatibility


beajeanm:
I'm running a monica instance on postgres and I run into a few issues during my set-up:

- [https://github.com/monicahq/monica/blob/c199c5e4f017404a18b3c9733d8f97ff2fb3e5cb/database/migrations/2017_05_04_193252_alter_activity_nullable.php](url) uses the MODIFY keyword. This is valid for MySql and Oracle, but not Postgresql.
I had to use `DB::statement('ALTER TABLE "activities" ALTER COLUMN "activity_type_id" TYPE INTEGER;');`
I'm not familliar with Laravel migration, so I'm not sure how to make the code pick the right SQL statement based on the configured database.

- A few columns in the users and contacts were created with NOT NULL constraints and monica violated those constraints when creating a new users or inserting a new contact.
**users**: gender, facebook_user_id
**contacts**: number_of_tasks_in_progress, number_of_tasks_completed, default_avatar_color
I'm not sure if this was due to a bug on the version of monica I've checked out or an imcompatibility with postgresql.

erdmenchen:
The NOT NULL constraints issue applies to SQLite, too.

djaiss:
The first item of this issue is fixed and in production.

jdambron:
I confirm having the same problem with NOT NULL constraints on SQLite

beajeanm:
Thanks for fixing the first point. Do we know why we are having inconsistency with nullable columns between MySql and Postgresql/SQLite?

rixx:
I received Postgres errors when setting up, aswell:

```
                                                                                                                         
  [Illuminate\Database\QueryException]                                                                                   
  SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "CHANGE"                                              
  LINE 1: ALTER TABLE contacts CHANGE COLUMN gender gender ENUM('male'...                                                
                               ^ (SQL: ALTER TABLE contacts CHANGE COLUMN gender gender ENUM('male', 'female', 'none'))  
                                                                                                                         

                                                                             
  [Doctrine\DBAL\Driver\PDOException]                                        
  SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "CHANGE"  
  LINE 1: ALTER TABLE contacts CHANGE COLUMN gender gender ENUM('male'...    
                               ^                                             
                                                                             

                                                                             
  [PDOException]                                                             
[...]

PR #140, opened 2017-06-08
Fix migrations for Postgres

alexhouse:
Postgres doesn't work with backticks; makes sense to use actual Laravel migration syntax for changing columns where possible to allow SQL grammar to take care of the rest.


beajeanm:
It would fix the first point of #137

alexhouse:
I was going to extend the migration to include nullable changes to the other two columns you mention in #137 but figured that might be better as a separate PR? Happy to include in this one if preferred, however.

djaiss:
Thanks a lot for this PR!

PR #158, opened 2017-06-09
Moved all instances of env out of app

degan6:
Moved all instances of env() out of the app folder.
Laravel doesn't like when env() is used in the app folder

Created two new varibles in the app config file
email_new_user_notification for email to send new registrations too
disable_signup for disabling signup


djaiss:
Thanks a lot for caring.

However I don't think this is the right move. Why is the current way of using env variables bad?

degan6:
If you call `php artisan config:cache` during setup to the `env()` function will return null or the wrong reset unless you use `config()`.

https://stackoverflow.com/questions/34420761/laravel-5-2-not-reading-env-file

Its a known problem in Laravel and can cause all sorts of headache if you run into it and don't know where to start troubleshooting. 

https://github.com/laravel/framework/issues/8191

djaiss:
Very interesting, I've learned something tonight :-D Thanks!

Issue #153, opened 2017-06-09
Global "From" Address is hardcoded


kstrauser:
Right now, all emails are hardcoded in `config/mail.php` to come from

```php
    'from' => ['address' => 'regis@monicahq.com', 'name' => 'Regis from Monica'],
```

This should come from an environment variable that that site owners can specify their own senders. The current value, besides being confusing to people getting reminders from a domain they don't recognize, is likely to be lost to spam filtering. Right now `monicahq.com` has an [SPF record](http://www.openspf.org) that says "all valid email from a sender @monicahq.com originates from these specific IP addresses":

```shell
$ dig +noall +answer -t txt monicahq.com
monicahq.com.		199	IN	TXT	"v=spf1 a mx include:spf.mtasv.net ~all"
$ dig +noall +answer -t txt spf.mtasv.net
spf.mtasv.net.		65757	IN	TXT	"v=spf1 ip4:50.31.156.96/27 ip4:104.245.209.192/26 ~all"
```

A lot of mailservers would see my server claiming to send an email from that domain, say "that mailserver is lying! It must be a spammer!", and cause all sorts of problems.

adrum:
It seems weird that there's a setting for `APP_EMAIL_NEW_USERS_NOTIFICATION` in .env as well. We should probably streamline this.

## Possible Solution

- Add `APP_EMAIL_FROM_NAME` and `APP_EMAIL_FROM_ADDRESS` to `.env` for use in `config/mail.php`
- Deprecate `APP_EMAIL_NEW_USERS_NOTIFICATION`
- Use `APP_EMAIL_NEW_USERS_NOTIFICATION` as a fallback for either of the new .env vars to guarantee backwards compatibility for existing installations. We might might be able to use `Monica` (or something) as the name fallback.
- Use `APP_EMAIL_FROM_ADDRESS` in `app/Jobs/SendNewUserAlert.php` and whatever email jobs could popup in the future

kstrauser:
In #133, @adrum wrote:

> #153 differs from the .env settings APP_EMAIL_NEW_USERS_NOTIFICATION? It seems like these should be one settings, and by possibly adding APP_EMAIL_FROM_NAME?

I think they're different things. The `_NEW_USERS_NOTIFICATION` is a recipient. The proposed `_FROM_NAME` (and/or `_FROM_ADDRESS`) would be a sender.

[...]

PR #160, opened 2017-06-09
Global mail config for address and name 

adrum:
This is a fix for #153


.env.example
```diff
@@ -26,6 +26,8 @@ MAIL_PORT=2525
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
+MAIL_FROM_ADDRESS=regis@monicahq.com
```
djaiss: Please replace this by `ValidEmailAddress` or something, so this email is never used by people ever.

.env.example
```diff
@@ -26,6 +26,8 @@ MAIL_PORT=2525
...
+MAIL_FROM_NAME="Regis from Monica"
```
djaiss: Please replace this by `YourName` or something.

.env.example
djaiss: The problem is that people will copy this example file and barely touch the default if they don't have to. By removing MY info entirely, we are sure this won't pop up in my email ever.

adrum: Yup, that's a good call.

PR #163, opened 2017-06-09
Translation/small changes

errogaht:
hello, im scan templates for some not translated and found this. also small changes to russian translations


djaiss:
Thanks a lot!

PR #165, opened 2017-06-09
added support for a unix socket

adrum:
When setting up the project for a MAMP installation on a Mac (LAMP), I ran into some issues with the database connection via `localhost` while running `php artisan migrate`. This pull request will automatically detect if `DB_HOST` starts with a `/` and will automatically set it up as a unix socket.

This might also help some people in the future when running `php artisan command` after downloading the forthcoming zip installation procedure.

https://stackoverflow.com/a/22241230/4127142

https://stackoverflow.com/a/20845309/4127142


djaiss:
Can you tell me if I should expect difficulties in production with this (I'm using nginx and the defaults provided by Homestead)? Thanks!

adrum:
You shouldn't run any issues with this in production, especially if you're using the default. It just basically detects if the db host is a path to a file and will inject that into the unix socket variable inside of Laravel's db config. 

Another approach would be creating a new .env var called like `DB_UNIX_SOCKET`, which would basically do the same, but just pull the value from a different environment variable. 

The approach currently in this PR just unsets the DB_HOST and DB_PORT since they're not needed due to the db connection happens over a unix socket.

config/database.php
```diff
@@ -130,3 +130,11 @@
     ],
 
 ];
+
```
djaiss: Can you add documentation in the code for this, as comments? I just don't want that someone reading this won't understand why this is necessary in some cases. Thank you!

djaiss:
`database.php` has significantly changed since then. Can you see what would be needed for this to work? Thanks so much and sorry about the very long time it took me to review this.

aejnsn:
Why alter all this configuration in the application, when you could configure the database server to bind to a network address? The configuration is only being altered so this will run in local environments like MAMP, XAMPP, WAMP, etc. I would honestly just bind MySQL to my IP or localhost and be done with it.

Typically you will not see production deployments using Unix sockets instead of TCP/IP.

djaiss:
Because I'm not expert enough to know if this change could break existing installations (on Docker, Heroku or else), I don't know if we should merge it.

What about adding a note in the documentation?

CONTRIBUTORS
```diff
@@ -7,7 +7,7 @@ Format: Firstname Lastname @github-handle <email>
 
 CONTRIBUTORS
 ------------
-Regis Freyd @djaiss
+Regis Freyd @djais
```
asbiin: Not sure why this is changed ?

config/database.php
```diff
@@ -144,6 +144,20 @@
...
 
+/*
+ * If the database host starts with a /, it is likely to be used with a
+ * unix socket. We can safely unset the host and the port since they
+ * likely won't be used in this configuration. This configurion likely
+ * won't be used in production, but rather when running with MAMP, XAMPP,
+ * WAMP, etc.
+ */
+
+if (substr($db['connections']['mysql']['host'], 0, 1) == '/') {
```
asbiin: IMO, it's safier just to add a new line in connections/mysql array, like this:
     'unix_socket' => env('UNIX_SOCKET', ''),
If the socket is set, laravel will use it first, so we don't need this config that depend on if the first character of "host" is a slash or not (which is dangerous)

CONTRIBUTORS
djaiss: It's a mistake for sure.

PR #169, opened 2017-06-09
converted columns for emoji support

### LINKED ISSUE #168: Emoji Support
Currently, the default database encoding for MySQL is `utf8`. It needs to be `utf8mb4` to add this support. This is the default starting in Laravel 5.4

adrum:
Fixes #168 

This was a little bit involved and touches every text based column, table and schema.

Links to help me along the way are below:
 

- [Generated Statements](https://dba.stackexchange.com/a/104866/125329)
- [I wanted to know why this wasn't there to begin with. Now I know: Laravel 5.4 add it](https://parvezmahbub.wordpress.com/2017/06/04/laravel-5-4-specified-key-was-too-long-error/)
- [DB Schema conversion issue](https://stackoverflow.com/a/34545211/4127142)
- [String/varchar length needs to be 191, due to 4-byte character encoding](http://aoverton.com/web-development/using-mysql-utf8mb4-with-laravel-5/)


app/Providers/AppServiceProvider.php
```diff
@@ -17,6 +18,9 @@ public function boot()
       View::composer(
         'partials.components.country-select','App\Http\ViewComposers\CountrySelectViewComposer'
       );
+
+      Schema::defaultStringLength(191);
```
djaiss: Why is that? There are some places in the code where we wanted to limit the number of characters in an input field to 255 characters. If we put this, does that mean that the default VARCHAR is limited to 191?

degan6: Its for backward compatibility for older versions of mysql. 

https://laravel-news.com/laravel-5-4-key-too-long-error

TobiasFroberg: Just looking through some PRs and noticed this one.

What would happen to the installs of Monica that have a varchar-field that have values of 192 characters or longer? Truncated data is not so fun.

If breaking changes is fine, just forcing users to enable `innodb_large_prefix` on the older MySQL-installs would be a much more elegant solution and wouldn't cause any data loss.

https://stackoverflow.com/questions/43153719/laravel-5-4-migrate-key-is-too-long-error/43162269#43162269

dereuromark:
This looks useful and necessary, I once [faced the same issue](https://www.dereuromark.de/2016/11/11/emoji-and-cakephp/).
Why didn't you guys proceed here yet? Anything below 4byte risks data loss when saving it.

asbiin:
OK this PR finally works for me.
I removed this (dangerous) `Schema::defaultStringLength(191);` line, because we don't want to risk on truncate datas.
mysql 5.7.7 or later don't need it. We can simply specify it needs this version of mysql or later.

PR #170, opened 2017-06-09
Default Config + APP_KEY seed

jeremydenoun:
- replace some constant by config value (env) especially for TimeZone & Locale
- change default APP_KEY to 32 char (avoid warning on specif openssl version) and update readme to remember 'php artisan key:generate' usage


djaiss:
I agree with the comments @erdmenchen made. Can you check @jeremydenoun ?

jeremydenoun:
It's fix @djaiss @erdmenchen

djaiss:
Thanks a lot but I can't merge this as is.

* You've set `APP_DEFAULT_LOCALE=en` as a new ENV variable in order to replace the `locale` setting defined in `config.php`. It's ok, but the problem is that currently this variable is not used anywhere in the code, so changing it has no effect at all. Locale is actually defined at the user level (it's the `locale` value in the User table). Can you remove this change?

Apart from that, I think we are ok.

jeremydenoun:
I think we should use it as default locale before user sign-in (especially for system mail) but this can be done in another PR

-- 
JD (iPhone)

jeremydenoun:
done

rabblac:
Could you add the command that generates the new key in the Docker configuration too?
You should just add this new line in https://github.com/monicahq/monica/blob/master/docker/test-server.sh
```
${ARTISAN} key:generate
```
before
```
${ARTISAN} migrate --force
```
Thanks!

jeremydenoun:
done

djaiss:
Sorry it took so long to review the PR. Thanks a lot for your work.

Issue #180, opened 2017-06-09
Docker version doesn't send reminder emails


kstrauser:
I've been running a site in Docker for the last day or so, but haven't received any reminder emails yet. What triggers the cron job that runs `monica:sendnotifications`?

enko:
@kstrauser a cron daemon ;-)

The most simple way to solve this problem would be to install a cron daemon into the docker container.

This brings another problem I've been thinking about. The current docker environment is missing direction, as it seems to be wants to be production and development environment at the same time. But there can be only one. My thoughts are, that in this repo should be a docker and docker-compose file for easy development of this software. This means the docker image should be lean and only contain all the stuff for running a development environment: php, composer and npm. In my dev environment I don't want a cron daemon which can cause unnecessary side effects.

The production docker image should be a separated repository (fe `monicahq/monicahq-docker`) and contain all the stuff necessary to run an instance of monica with a tuned webserver with php and a cron daemon for the notifications. It should also contain some docker-compose examples on how to run it with a database.

What do you think?

kstrauser:
It's already 90% of the way there. `docker/Makefile` is meant to be a set of entry points into the container, so that `docker-compose run monicahq test_server` runs the current instance as you see it, and `... shell` gets you a command line so you can poke at it.

In the future it would be very easy to add a target for `production_server` that launches both a server and a cron daemon, but you'd still be able to select which one you want to run.

djaiss:
`monica:sendnotifications` is a command ([documentation](https://laravel.com/docs/5.4/artisan#introduction)) that is triggered by a cron job that should run every minute. All commands are run through a single CLI instruction:

```
php /home/forge/default/artisan schedule:run
```

This job needs to run every minute, and Laravel has made something super convenient actually. All the jobs can be defined inside the application itself, and `schedule:run` will check them. For instance right now we have two commands defined in Kernel.php:

```
$schedule->command('monica:sendnotifications')->hourly();
$schedule->command('monica:calculatestatistics')->daily();
```

kstrauser:
Verified fixed in #183.

PR #183, opened 2017-06-10
Run cron jobs once a minute in Docker

kstrauser:
This addresses #180 by running Apache as a background process, then calling `artisan schedule:run` in a loop.

The next step from here would be to switch to a real init process like runit or supervisor, but if we can avoid that so easily I'd prefer to wait until we really have to.


djaiss:
Thanks for the PR!

PR #192, opened 2017-06-10
Fix registration for deleted users bug

funcoding:
Email exists bug in deleted user registration has been resolved.


djaiss:
Thanks a lot for this work. However the code has changed a lot since then and this was fixed in another PR. Again, thank you!

renedekat:
When I delete my account somewhere I expect it to be deleted, not marked as deleted. 

If you take privacy seriously, all data should be permanently removed.

djaiss:
@renedekat This is the case: https://github.com/monicahq/monica/blob/master/app/Http/Controllers/SettingsController.php#L79

Please read the code before making this kind of statement.

PR #201, opened 2017-06-11
Add import:vcard command

yamakadi:
This pull request adds a command to import vcards (similar to #60 ).

Usage: `php artisan import:vcard {user: User's email address} {path: Path to a .vcf file}`

The command checks whether an email already exists, so duplicates shouldn't be an issue. However, I think the unique constraint is too eager. If another user wants to add the same person (whether through import or otherwise), they will not be added. I think this is a bug and the constraint should be `['account_id', 'email']`. I can handle it in another pull request.

### DIFF
--- a/app/Console/Commands/ImportVCards.php
+++ b/app/Console/Commands/ImportVCards.php
@@ -1,0 +1,189 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Contact;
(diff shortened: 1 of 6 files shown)


djaiss:
Thanks so much for this PR. I've tested with a couple of different vCards and it works perfectly.

PR #203, opened 2017-06-11
Changes the unique constraint for contacts from 'email' to ['account_…

yamakadi:
…id', 'email']

I think the unique constraint is too eager for the contacts table. 

If a user wants to add a contact with an already existing email, they'll encounter an error. Since they aren't the one who added that email to their contacts, it will be confusing & frustrating. 

I think this is a bug and the constraint should be `['account_id', 'email']`.


BranTurner:
Should `user_id` not be used to associate the email address with the user, rather than the `account_id`?

djaiss:
It was indeed something I didn't think about when I wrote this. It's definitely a bug. I'll merge it in a few minutes.

PR #209, opened 2017-06-11
Fixing some relationship inconsistencies and reducing dashboard queries

yamakadi:
This normalizes all relationships and adds relations to models to take advantage of laravel's eagerloading and reduce queries to the database.

This might go against the 'add any kind of complexities whatsoever' rule in the contribution guide, but I think adding relations would eventually simplify the code base. The amount of queries just on the dashboard fell from ~70 to 26. There's also a lot of boilerplate code in the Contact model to retrieve relationships and to maintain accurate count of them (fields like `number_of_*`). They can be greatly simplified with the addition of relations.

One other thing in the Contact model was the definition of the Contact-User relation. This definition uses the `account_id` field in the `contacts` table and and relates it to `id` in `users`. I'm not sure if this is the intention, since the `id` from `accounts` table is used when assigning `account_id` for a contact. This seems to have worked so far, because both a user and its associated account are created at the same time during registration and their ids match, but if there was any difference, this relationship would relate wrong contacts and users. I changed the relationship name to `account`. This makes the path to a user from a contact is `$contact->account->user` right now.

I may have gone overboard with this though, so I'm ready for the request to be rejected.


app/Http/Controllers/DashboardController.php
```diff
@@ -23,100 +23,72 @@ class DashboardController extends Controller
...
-        $lastUpdatedContacts = Contact::where('account_id', Auth::user()->account_id)
-                                    ->orderBy('updated_at', 'desc')
-                                    ->limit(10)
-                                    ->get();
+        $account = Auth::user()->account;
 
-        // Latest statistics
-        $contacts = Contact::where('account_id', Auth::user()->account_id)
-                                    ->get();
+        $lastUpdatedContacts = $account->contacts->take(10);
```
julesjanssen: Suggestion: change this line to `$account->contacts()->limit(10)->get()` to reduce the data that has to be returned by the DB.

app/Http/Controllers/DashboardController.php
```diff
@@ -23,100 +23,72 @@ class DashboardController extends Controller
...
 
-        if (count($contacts) == 0) {
+        // Latest statistics
+        if (count($account->contacts) === 0) {
```
julesjanssen: `$account->contacts()->count()` would execute a `COUNT` query instead of retrieving all contacts from the DB.

erdmenchen:
I like it a lot! Good job!

BranTurner:
This seems much more efficient, and would be a great improvement! How thoroughly have you tested? I can't see any errors myself, but the tests present don't cover many features so it's always risky pushing large changes.

yamakadi:
@julesjanssen The reason I used collections instead of querying for counts in some cases was to further reduce the query count (all of the relationships are already load or will need to be loaded by the end of the request), but it would be more consistent to change them to the suggested versions. I'll do that in a few hours.

@BranTurner Tests wise, you are right. Not everything is covered. I checked manually for every change in the contact model and the dashboard/people controller, but it could still give an error elsewhere. Still, precisely due go the size of this commit and where it touches, I think it'd be better to merge it sooner and fix any bugs quickly instead of waiting. New pull requests that try to add missing features will likely rely on some of the methods changed here and might keep enlarging this commit.

@erdmenchen Thank you!

julesjanssen:
@yamakadi If you have a lot of contacts in the DB, selecting them all and limiting / counting them like it is done here is not only slow and heavy on the DB. It's also going to cause memory issues.

BranTurner:
@yamakadi I agree pushing this earlier rather than later would simplify further development. I actually think your changes would simplify testing. If I have time tomorrow I'm going to write some tests for the fundamental features which should help alleviate the problem.

yamakadi:
@julesjanssen made the changes

djaiss:
Thanks a lot for all this work. This is great and much better than the original code. I've played with it and everything seems to work, but definitely more tests in the future will help make this kind of refactors and not being scared of breaking down everything.

It's better to merge this sooner rather than later and check any problems that might occur.

Thanks again.

PR #211, opened 2017-06-11
Fix debt description on the dashboard not fetching the localized 'debts_for' resource

### LINKED ISSUE #210: Debt: Literal "dashboard.for" should not be presented to the user
For example adding a $999 debt owed to you (the user) by PersonA yields:

`PersonA: You are due $999 dashboard.for test debt description`

https://github.com/monicahq/monica/blob/master/resources/views/dashboard/index.blade.php#L160

I believe the fix is to change this:

```html
<span class="debt-description">{{ trans('dashboard.for') }}</span>
```

to:

```html
<span class="debt-description">{{ trans('dashboard.debts_for') }}</span>
```

Phrohdoh:
Closes #210

### DIFF
--- a/resources/views/dashboard/index.blade.php
+++ b/resources/views/dashboard/index.blade.php
@@ -157,7 +157,7 @@
                         ${{ $debt->amount }}
 
                         @if (! is_null($debt->reason))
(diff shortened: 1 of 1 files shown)


djaiss:
Good call!

PR #213, opened 2017-06-11
Change the email text to be a mailto link

### LINKED ISSUE #208: Email should be a mailto: link
Similar to #77

Phrohdoh:
I do not know php so there may be a way to only call `getEmail` once (so we don't have to fetch it over and over again). If so please let me know and I'll do that.

If there is a better way to implement this I am certainly all for that too.

If merged this will close #208.

### DIFF
--- a/resources/views/people/dashboard/people-information/index.blade.php
+++ b/resources/views/people/dashboard/people-information/index.blade.php
@@ -33,7 +33,9 @@
         @if (is_null($contact->getEmail()))
         {{ trans('people.information_no_email_defined') }}
         @else
-        {{ $contact->getEmail() }}
+        <a href="mailto:{{ $contact->getEmail() }}">
+          {{ $contact->getEmail() }}
+        </a>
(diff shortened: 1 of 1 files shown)


Phrohdoh:
I could use `<?php ... ?>` if desired.

http://laravel-recipes.com/recipes/256/assigning-a-variable-in-a-blade-template

djaiss:
@Phrohdoh there is no need to use `<?php`. Laravel uses Blade to output PHP data (using `{{...}}`), which is then compiled into `<php>`, but Blade lets us have cleaner templates.

Thanks for the PR!

Phrohdoh:
@djaiss I mentioned `<?php` so we wouldn't have to do `$contact->getEmail()` multiple times.

djaiss:
@Phrohdoh oh I see. Indeed we could probably improve that, but I wonder if Laravel makes an extra call to the DB or cache the result.

Issue #215, opened 2017-06-11
Deploy to Heroku button


djaiss:
We need to provide an official Deploy to Heroku button for people who want to host Monica on Heroku on a basic plan.

Documentation: https://devcenter.heroku.com/articles/heroku-button

The only problem I can see with this is how we can script the fact that we need a database, as Heroku only provides DB through addons. We need to investigate.

BranTurner:
You can specify required addons in the [app.json](https://devcenter.heroku.com/articles/app-json-schema#addons) file. You would need to request a Heroku postgres instance, and then should be good to go!

Issue #187, opened 2017-06-10
RuntimeException ciphers


pcegarra:
[RuntimeException]
  The only supported ciphers are AES-128-CBC and AES-256-CBC with the correct key lengths.

Any idea?

webdistortion:
Remove the predefined key in your .env file it's not long enough for Laravel 

change it to : 
APP_KEY=xlhF31NeOlibJcoOW9tvZg7TkHcAZI3a

Should get you up and running.

djaiss:
Indeed. I've update the README to indicate that.

Actually you can run `php artisan key:generate` to create the right random key for you.

Feel free to close the issue once it's solved.

kstrauser:
That doesn't work for me if `APP_KEY` is unset:

```shell
/var/www/monica # unset APP_KEY
/var/www/monica # php artisan key:generate


  [RuntimeException]
  The only supported ciphers are AES-128-CBC and AES-256-CBC with the correct key lengths.
```

erdmenchen:
Insert a random 32 character long string as the app key. THEN run the artisan command key:generate to generate a valid new one.

kstrauser:
Is there a reason to do that beyond just using the random 32 character long string you made?

BranTurner:
@kstrauser No, any 32 char string will do! If `php artisan key:generate` isn't working, you can always try `php artisan key:generate --show` which will give you a string you can manually enter into the .env file.

tetov:
Should this really be closed? I think there should be an update either to the README or .env.example should have 32 char string in APP_KEY.

BranTurner:
@tetov I agree a 32 char key should be included in `.env.example`, I think the error might confuse a lot of people who aren't familiar with Laravel.

djaiss:
This is now included as part of `.env.example`.

https://github.com/monicahq/monica/blob/master/.env.example

PR #219, opened 2017-06-12
Heroku Deployment Button

BranTurner:
This pull request is to solve issue #215. 

Summary:

1. Created app.json which (a) deploys a ClearDB instance, (b) makes required environment variables available
2. Created a console command which simplifies setup to one command: `php artisan setup:production` - this runs the commands detailed in the docs
3. Altered the database configuration file, so that when using Heroku the database information is parsed from the environment variable provided by the ClearDB instance

I attempted to use Heroku's Postgres instance as this is slightly more generous in its limits and speed, however Laravel has some issues with the use of the `enum` column type when using postgres so it wasn't possible.

Additionally, I have included `APP_KEY` as an environment variable in the app.json file for two reasons: 

1. It will allow the user to access this from within the Heroku interface should they need it
2. `php artisan key:generate` is still throwing errors due to the lack of `.env` file as detailed in #187 

In order to test the Heroku deployment before merging, you can use the URL: [https://www.heroku.com/deploy?template=https://github.com/BranTurner/monica/tree/deployment-heroku-button](https://www.heroku.com/deploy?template=https://github.com/BranTurner/monica/tree/deployment-heroku-button)

### DIFF
--- a/app.json
+++ b/app.json
@@ -9,14 +9,6 @@
     "postdeploy": "php artisan setup:heroku"
   },
   "env": {
-    "HEROKU": {
-      "description": "Informs Monica that the app is running on Heroku.",
-      "value": "true"
-    },
-    "DB_CONNECTION": {
-      "description": "Informs Monica to use the Heroku database.",
-      "value": "heroku"
-    },
     "APP_DISABLE_SIGNUP": {
       "description": "Disable user signup.",
       "value": "false"
--- a/app.json
+++ b/app.json
@@ -3,8 +3,18 @@
   "description": "Personal Relationship Manager - a new kind of CRM to manage your friends and family.",
   "repository": "https://github.com/monicahq/monica",
   "logo": "https://raw.githubusercontent.com/monicahq/monicahq.github.io/master/assets/img/monica-logo.svg",
-  "keywords": ["CRM", "family", "friends", "relationship", "PRM"],
-  "addons": [{"plan": "cleardb:ignite"}],
+  "keywords": [
+    "CRM",
+    "family",
+    "friends",
+    "relationship",
+    "PRM"
+  ],
+  "addons": [
+    {
+      "plan": "cleardb:ignite"
+    }
+  ],
   "scripts": {
     "postdeploy": "php artisan setup:heroku"
   },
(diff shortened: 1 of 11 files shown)


config/database.php
```diff
@@ -130,3 +130,23 @@
     ],
 
 ];
+
+// If the app is deployed on Heroku interpret the postgres URL
```
djaiss: We need to document this slightly better I think. Like exactly why this is bad and why we do this instead. The reason youve posted on the PR is good enough to me.

app.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "Monica",
+  "description": "Personal Relationship Manager - a new kind of CRM to manage your friends and family.",
+  "repository": "https://github.com/monicahq/monica",
+  "logo": "https://raw.githubusercontent.com/monicahq/monicahq.github.io/master/assets/img/monica-logo.svg",
+  "keywords": ["CRM", "family", "friends", "relationship"],
```
djaiss: Could you add `PRM` too?

app/Console/Commands/Setup.php
```diff
@@ -0,0 +1,45 @@
...
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
```
djaiss: As this is called after the deployment during the setup on Heroku, you can also call `php artisan key:generate` to generate automatically a random string. This would need to be tested on heroku though, but it would be one less step to do.

app.json
```diff
@@ -0,0 +1,45 @@
...
+      "value": "noreply@example.com"
+    },
+    "MAIL_FROM_ADDRESS": {
+      "description": "",
+      "value": "noreply@example.com"
+    },
+    "MAIL_FROM_NAME": {
+      "description": "",
+      "value": "Bob Smith"
+    }
```
djaiss: Have you tried this, without those env variables set? Does it work?

```
CACHE_DRIVER=database
SESSION_DRIVER=file
QUEUE_DRIVER=sync
```

djaiss:
Left a couple of notes. I'm impressed with what you've done. Thanks so much.

Issue #218, opened 2017-06-11
Journal bug


webchun:
Hi I tried to add a journal on monicahq.com and it returned some random characters
![image](https://user-images.githubusercontent.com/4277547/27015432-c9914d88-4f37-11e7-80ef-ad8e4227fd76.png)

BranTurner:
This is a problem with encryption in the JournalController - I've submitted a pull request to fix this at #220. Thanks for pointing this out!

djaiss:
![image](https://media.giphy.com/media/mHjBdBZuqBvJC/giphy.gif)

I'm sorry you encountered that bug.

webchun:
niiice ;)

PR #220, opened 2017-06-12
Fix journal bug (#218)

BranTurner:
I believe PR #199 missed the encrypt/decrypt functions in JournalController leading to the issues seen in #218. 

Summary:

- Removed encrypt function in JournalController@save

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,11 @@
+2017-06-11
+----------
+
+Improvements:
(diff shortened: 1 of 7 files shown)


djaiss:
Thanks a lot for taking care of this. Facepalm to me.

BranTurner:
Not at all - very easy to miss. Glad it's solved now!

PR #225, opened 2017-06-12
Deployment heroku button

BranTurner:
Updated with the comments as mentioned in #219. I accidentally deleted the branch, so had to submit a new PR.

> Have you tried this, without those env variables set? Does it work?
>CACHE_DRIVER=database
>SESSION_DRIVER=file
>QUEUE_DRIVER=sync

The deployment works without explicitly setting these variables, as Laravel defaults to them in the config file.

>As this is called after the deployment during the setup on Heroku, you can also call php artisan key:generate to generate automatically a random string. This would need to be tested on heroku though, but it would be one less step to do.

Unfortunately, Heroku does not pull environment variables from `.env` files, so `php artisan key:generate` does not work. Heroku has the option to generate a random key, however the length is 64 characters and not customisable.

[Feel free to test the deployment here before merging.](heroku.com/deploy?template=https://github.com/BranTurner/monica/tree/deployment-heroku-button)


djaiss:
Just tested it, and it works. Thanks a lot for this amazing job.

PR #227, opened 2017-06-12
Refactoring PeopleController for better performance and readability

yamakadi:
I refactored PeopleController in the same way as the DashboardController to improve performance and readability. I think this needs to be separated into files and logic for adding/removing associated models to contacts should be more standardized.

I also think there's need for a global account scope that'll take the account id from the logged in user and apply it as a constraint to all queries, so we can stop checking whether a model belongs to an account in every controller method. The only problem this might cause is not for the users but for the administrative sections where we need access to all accounts, not a specific one. In such cases, we can just use `withoutGlobalScopes()` and query that way. I can add it if you think it's necessary, too.


djaiss:
Thanks a lot! Looks much better indeed.

Can you update it to match master as I've changed some stuff on `Contact.php` moments ago to fix a bug on debts? Thanks

yamakadi:
@djaiss I'll get to it in the next few hours.

If possible, I'd like to separate these controllers in another pull request, do they can be more easily tested. Would you be open to this by any chance? Separating the controllers would also allow moving out some of the code that doesn't necessarily belong in the Cobtact model.

djaiss:
@yamakadi I agree on the splitting into multiple controllers. `PeopleController.php` is definitely way too big.

yamakadi:
I'll send another pull request after fixing this one then.

djaiss:
> I also think there's need for a global account scope that'll take the account id from the logged in user and apply it as a constraint to all queries, so we can stop checking whether a model belongs to an account in every controller method

YES. But perhaps in another PR? This is is already huge.

djaiss:
❤️ 

Thanks a lot.

yamakadi:
Sorry you had to fix the clashes. I couldn't get to a computer in time. Now I can work in the other PR.

djaiss:
No worries!

PR #228, opened 2017-06-12
Eager Loading People

phainv:
Load relationship kids before count total kid (N + 1 query). Check laravel debugbar see result


phainv:
same #227

PR #238, opened 2017-06-12
Remove US duplicate from countries seeder table

### LINKED ISSUE #233: Country Dup: United States listed Twice
country

When editing a contact and selecting country, "United States" is listed twice.

BranTurner:
Small change: remove one instance of US from countries seeder table as it was duplicated. Fixes #233.


djaiss:
Thanks for this PR. This is definitely a bug.

I still need to fix something related to this: 
* we now have two `US` in the database. I need to remove the second entry, but also, for every user that had the second `US` as their country, I need to change this value to the first `US` id instead. I'll submit a PR.

PR #239, opened 2017-06-12
Reordered countries seed by ISO code to improve readability

colourgarden:


djaiss:
Tests fail because we've changed the order of the seeder. The test is definitely not robust enough.

Phrohdoh:
How does a reorder improve readability?

colourgarden:
Maybe digestibility would be a better description. If this list needs to be edited again in the future, it's easier to navigate if the countries are ordered.

djaiss:
Thank you very much for this PR. The test is still failing, so I'll close the PR because of the lack of activity.

PR #247, opened 2017-06-12
Grammar agnostic enum change

martimarkov:


Phrohdoh:
Travis failed PHP 7 and 7.1 with:

```
The command "php artisan migrate --env=testing --no-interaction -vvv" exited with 1.
```

Is this something that can be fixed?

martimarkov:
As per https://stackoverflow.com/questions/33140860/laravel-5-1-unknown-database-type-enum-requested it's a known issue.

yamakadi:
It's not ideal, but if the issue is renaming the column, as per the stack overflow article, maybe you can cache the existing column data before dropping and reading it.

martimarkov:
You will always have rename it at some point. Strangely thou this works with Postgres. Any chance you can run the migration on MySQL test server to check if it's working?

yamakadi:
Well, if it's a known issue and doctrine can't handle any renaming,
there'll have to be workarounds as long as the enum fields are kept.

I think a better option would be to handle enum fields in the code and just
have a string field for gender, but that's another discussion.

As it is, the pr doesn't work on my end either.

djaiss:
I agree that we should avoid from now on having ENUM fields in the codebase. It's weird though that Laravel allows us to use ENUM and says Postgresql is supported, when it obviously does not.

djaiss:
This has been fixed since then. Thanks a lot for this PR though, highly appreciated.

martimarkov:
@djaiss where?

djaiss:
@martimarkov check https://github.com/monicahq/monica/blob/master/database/migrations/2017_06_07_173437_add_multiple_genders_choices.php

PR #243, opened 2017-06-12
Added basic tests for contacts

BranTurner:
Added basic tests for contacts:

1. Ensure user can see contacts
2. User can add contacts
3. User can add a note, activity, event, task, debt, and food preference to a contact
4. Test a contact can be deleted

There are many more contact tests I would like to add, and I feel the tests I'm pushing now can be refactored to make them more efficient. However, I don't have much more time to work on them today, and feel like some tests are better than no tests (especially when big changes are being merged!). 

I will hopefully improve this over the coming days, but I think this is a reasonable start.

PR #253, opened 2017-06-13
Separate people controller into smaller chunks [Completed]

yamakadi:
This is an incomplete pull request aimed at separating PeopleController into smaller chunks and move some of the logic to more appropriate places. I'm sending it in before completion, so we can discuss some of the decisions I've made and talk through them in case there are any disagreements.


resources/views/people/activities/edit.blade.php
```diff
@@ -32,57 +32,10 @@
...
-              </div>
-
-              <div class="form-group actions">
-                <button type="submit" class="btn btn-primary">{{ trans('people.activities_add_cta') }}</button>
-                <a href="/people/{{ $contact->id }}" class="btn btn-secondary">{{ trans('app.cancel') }}</a>
-              </div> <!-- .form-group -->
-            </form>
+            @include('people.activities.form', [
+              'method' => 'PUT',
+              'action' => route('people.activities.update', [$contact, $activity])
```
BranTurner: As a brief warning, if #243 is merged before this it will throw errors because the route URLs were not standardised. Some forms posted to `/save` while some posted to `/store`, so the tests I wrote reflected this. Once this is merged I will update the tests to reflect the new URLs.

djaiss:
Reading through the code.  Perhaps it's time to rename PeopleController and all `People` mentions to Contacts.

Historically, and stay with me because this is an interesting concept in my opinion: I wanted to have two kind of entities in the system
* People
* Families

A family would have been a reunion of at least 2 people. Instead of having one contact to add to the system, you would have a choice to add either a contact, or more into one single entity.

Think of it this way: we either have single people as friends, or couples. But I abandoned this idea because of the technical complexity and I wanted to ship. Therefore, I called the controller and methods People instead of Contacts.

app/Contact.php
```diff
@@ -228,7 +254,7 @@ public function getLastActivityDate($timezone)
         $lastActivity = $this->activities->sortByDesc('date_it_happened')->first();
 
         return DateHelper::getShortDate(
-            DateHelper::createDateFromFormat($lastActivity->date_it_happened, $timezone),
+            Carbon::parse($lastActivity->date_it_happened, $timezone),
             'en'
```
djaiss: We probably need to return the locale of the user that owns this contact instead here.

app/Event.php
```diff
@@ -25,6 +26,24 @@ public function contact()
...
 
+    /**
+     * Limits the results to a specific object
+     *
+     * @param Builder $query
+     * @param Model $object
+     * @param string|null $key
+     * @return Builder
+     */
+    public function scopeForObject(Builder $query, Model $object, string $key = null)
```
djaiss: Where would you use this method?

app/Gift.php
```diff
@@ -2,99 +2,134 @@
...
+     * @return Builder
+     */
     public function scopeIsIdea(Builder $query)
     {
         return $query->where('is_an_idea', 'true');
     }
 
-    public function getName()
+    /**
+     * Check whether the gist is meant for a particular member
```
djaiss: `gist` → `gift`

app/Providers/RouteServiceProvider.php
```diff
@@ -19,14 +23,31 @@ class RouteServiceProvider extends ServiceProvider
...
-     * @param  \Illuminate\Routing\Router  $router
      * @return void
      */
     public function boot()
     {
-        //
-
         parent::boot();
+
+        Route::bind('contact', function ($value) {
```
djaiss: What's the benefit of putting this in this file?

app/Contact.php
yamakadi: The locale part is what was already there, but I can handle the change.

app/Event.php
yamakadi: If you look at the destroy methods for the new controllers, I'm using this scope to find related objects more easily and erase them in one go. Like:

$contact->events()->forObject($activity)->get()

For standard event object names (kid, activity, etc) just lowercasing the model name works, so I don't have to explicitly pass the event object type this way and the method handles it.

app/Providers/RouteServiceProvider.php
yamakadi: If you noticed, I'm not using contactId as field names in the routes file. It's to allow using route-model binding. You can take a look at "Explicit Route-Model Binding" section of Laravel's documentation.

These handle all the necessary integrity checks (contact belongs to user, task belongs to contact, etc) so you don't have to do it manually in each controller method. Coupled with FormRequest classes, we can be sure that if the request reached the controller, it has valid and appropriate data. That's why the new controllers I'm adding are much more skinny and easy to read through.

resources/views/people/activities/edit.blade.php
yamakadi: I can fix the tests if this one is merged after #243.

yamakadi:
I like this idea. If the related code and tables are simplified and
isolated further, something similar might be more easily achieved.

I think the misstep wasn't naming the controller PeopleController, but the
model "Contact". It would be much better to have a Person model that
centralized all contact related things. This way, a kid or a significant
other, an aunt, an uncle, etc. would just be another person in the
database, reducing the burden of maintaining kids and significant others.
The only required field for a contact is their name right now, so even if
all you know of someone's kid is their name, it would still work.

djaiss:
Once we finalize this PR, we can still rename the Contact object to something more solid. No problem with that.

Perhaps it's best to keep this PR the smallest possible (it's already super big) - I'm already worried about testing this before pushing to production 😀

yamakadi:
I'm testing everything manually before committing and trying to keep each
commit about one specific model, so you can checkout different commits to
try out the changes.

Let's hope nothing breaks.

Nebucatnetzer:
Would it maybe make sense to have a "dev" branch where new features get merged into?
This would maybe help with testing and wouldn't impact the stability of the master branch.

djaiss:
What's the difference between having a dev branch and letting you pull this PR and test it on your machine?

Nebucatnetzer:
Ah yes you're right. Wouldn't work here. I'm more used to a release based model :)

erdmenchen:
@djaiss this would let you integrate a set of changes into production (master) e.g. once a day (or once a week). No need to integrate and test every pull request excessivly, fearing that production could break. But we should discuss this in a seperate ticket, if needed.

yamakadi:
@djaiss PeopleController is already down to ~500 lines from ~1700. 

The only parts left to move are jobs, and food preferences, but I don't want to touch food preferences until it can be renamed to food preferences (not preferencies) across the board and I believe jobs might eventually become its own model, so moving it then would be better.

I think the PR can be tested and merged now. I've rigorously tested everything before committing them, but another set of eyes would be good.

djaiss:
Here comes my entire morning. Going to review and test in an hour.

app/Http/Controllers/People/ActivitiesController.php
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Show the form for creating a new resource.
+     *
+     * @param Contact $contact
+     * @return \Illuminate\Http\Response
+     */
+    public function create(Contact $contact)
+    {
+        return view('people.activities.add')
+            ->withContact($contact)
```
djaiss: I didn't know this kind of methods. Is this a Laravel thing or is this something you've implemented (I haven't read everything yet).

resources/views/people/gifts/index.blade.php
```diff
@@ -82,10 +82,10 @@
                 <a href="{{ $gift->getUrl() }}">{{ trans('people.gifts_link') }}</a>
               </span>
             @endif
-            @if(!is_null($gift->getWhoIsItFor()))
+            @if($gift->hasParticularRecipient()))
```
djaiss: Contains on extra `)` that should be removed.

app/Http/Controllers/People/ActivitiesController.php
BranTurner: I'm pretty sure it's a laravel thing, but I've never seen it covered in the documentation.

app/Http/routes.php
```diff
@@ -33,63 +35,61 @@
...
-        Route::get('/people/{people}/debt/add', ['as' => '.debt.add', 'uses' => 'PeopleController@addDebt']);
-        Route::get('/people/{people}/debt/{debtId}/edit', ['as' => '.debt.edit', 'uses' => 'PeopleController@editDebt']);
-        Route::post('/people/{people}/debt/{debtId}/update', ['as' => '.debt.update', 'uses' => 'PeopleController@updateDebt']);
-        Route::post('/people/{people}/debt/store', 'PeopleController@storeDebt');
-        Route::get('/people/{people}/debt/{debtId}/delete', 'PeopleController@deleteDebt');
+        Route::get('/people/{contact}/debt/add', 'People\\DebtController@create')->name('.debt.add');
+        Route::post('/people/{contact}/debt/store', 'People\\DebtController@store')->name('.debt.store');
+        Route::get('/people/{contact}/debt/{debt}/edit', 'People\\DebtController@edit')->name('.debt.edit');
+        Route::put('/people/{contact}/debt/{debt}', 'People\\DebtController@update')->name('.debt.update');
+        Route::get('/people/{contact}/debt/{debt}/delete', 'People\\DebtController@destroy')->name('.debt.delete');
```
djaiss: When trying to delete a debt

image

app/Http/Controllers/People/NotesController.php
```diff
@@ -0,0 +1,124 @@
+<?php
+
+namespace App\Http\Controllers\People;
+
+use App\Note;
+use App\Contact;
+use App\Http\Controllers\Controller;
+use App\Http\Requests\People\NotesRequest;
+
+class NotesController extends Controller
```
djaiss: `PeopleController.php` still has all the methods to manage a note. Is that an omission or on purpose?

app/Http/Controllers/People/ActivitiesController.php
yamakadi: It's a laravel thing. You can add variables to views with '->with{Variable}($variable)' method. It allows you to forgo using compact to turn variables into arrays or create an intermediary array to hold variables.

yamakadi: Also, the reason I'm passing an empty model to add views is to allow sharing forms for edit and add views.

app/Http/Requests/People/ActivitiesRequest.php
```diff
@@ -0,0 +1,34 @@
...
+use Illuminate\Validation\Rule;
+
+class ActivitiesRequest extends FormRequest
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
```
djaiss: Can you explain to me where we now check if we are authorized to perform a method?

app/Note.php
```diff
@@ -4,35 +4,60 @@
...
-        return $this->belongsTo('App\Contact');
+        return $this->belongsTo(Contact::class);
+    }
+
+    /**
+     * Return the markdown parsed body
+     *
+     * @return string
+     */
+    public function getParsedBodyAttribute()
```
djaiss: Did you silently add support for Markdown now?

app/Providers/RouteServiceProvider.php
```diff
@@ -2,6 +2,13 @@
 
 namespace App\Providers;
 
+use App\Activity;
+use App\Gift;
+use App\Reminder;
+use App\SignificantOther;
+use App\Task;
+use Route;
+use App\Contact;
```
djaiss: `App\Debt` is missing?

app/Note.php
```diff
@@ -4,35 +4,60 @@
 
 use App\Helpers\DateHelper;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Parsedown;
```
djaiss: Does this ship with Laravel? I don't see `composer.json` modified.

resources/views/people/gifts/index.blade.php
yamakadi: Done

app/Http/routes.php
yamakadi: Done.

This was oversight on my part, sorry. Adding App\Debt to RouteServiceProvider fixes it.

app/Http/Controllers/People/NotesController.php
yamakadi: Done.

This was another oversight.

app/Http/Requests/People/ActivitiesRequest.php
yamakadi: Authorization is done in the RouteServiceProvider while binding the model to route. You can see that it checks for `contact_id` and `account_id`.

app/Note.php
yamakadi: This ships with Laravel's mail component, but I also added it to composer.json now.

app/Note.php
yamakadi: I thought it might be a good addition, but only parsing is possible right now and it's not used anywhere in this PR.

app/Providers/RouteServiceProvider.php
yamakadi: Added.

yamakadi:
@djaiss All requested changes were done

djaiss:
Looks good to me. Thank you very much for the time and energy you've put into making this happen. A bit stressed to merged this to production but it should be ok.

djaiss:
There a some bugs in production that this PR introduced.

Next time we shouldn't introduce PRs that large.

We can't delete a kid anymore for instance. Trying to fix the other bugs.

yamakadi:
Certainly. Manual checking can only get so far, but at least there
shouldn't be a need for such big PRs anytime soon.

Thanks for fixing the kid issue. I'll help fix the others as they appear.

erdmenchen:
@djaiss: Thats why we asked for a develop branch beside the master branch.

That way, we can merge pull request to the dev branch and do more tests and further fixing before we finaly merge to master and production.

Maybe we should do this now. 
For more information on this: http://nvie.com/posts/a-successful-git-branching-model/ 

However: Great work @yamakadi ! This makes the code much easier to extend and reduces complexity.

PR #254, opened 2017-06-13
Fix Debian installation link in README

sid3windr:
The installation link for Debian was pointing to the Docker instructions.


djaiss:
Facepalm.

Thanks for the fix.

PR #262, opened 2017-06-13
Added php7-xmlreader to Docker as a new requirement

kstrauser:
@djaiss This adds the php7-xmlreader module that the new sabre dependencies require.


djaiss:
👍 

Thanks a lot

Issue #139, opened 2017-06-08
Want To Install In Sub Directory 📁 


GaneshKandu:
### I Want to install Monica in SubDirectory 

Hi,
I Want to install Monica in SubDirectory ie. https://localhost/monica/ instead of https://localhost/
i tried but getting 404 error of assets

i am trying to install public folder in sub directory of web root

- private
  - CHANGELOG
  - CONTRIBUTING.md
  - LICENSE
  - Procfile
  - app
  - artisan
  - bootstrap
  - bower.json
  - composer.json
  - composer.lock
  - config
  - database 
  - .........
- htdocs
  - subdir
    - build
    - css
    - favicon.ico
    - fonts
    - img
    - index.php
    - js
    - robots.txt
    - web.config
    - storage

```please mention version also it helps to keep up to date```

**Comment by virtadpt** [comment]
I am trying to do the same thing.  I can't log in, either - trying to log into https://server/monica/server.php tries to redirect me to /login instead of /monica/login, which gets me a 404.

**Comment by BranTurner** [comment]
This is because none of the URLs change when the `APP_URL` is changed. This would require an update of all links. If I have time today I might submit a pull request with updated links!

**Comment by somas95** [comment]
I suggest as workaround to use virtual hosts

**Comment by peperunas** [comment]
Yep @BranTurner that would be great.

**Comment by BranTurner** [comment]
@peperunas I'll try to get around to it at some point. I have a few appointments today, but may have some time when I get back!

**Comment by peperunas** [comment]
@BranTurner, that's great to hear. Kudos to you!

**Comment by adriangibbons** [comment]
Hi - same issue for me...

I can't use https://domain.com/monica/server.php though perhaps I am doing something wrong? I'm assuming I should go for server.php as otherwise I just get a directory listing...

Have set APP_URL=https://domain.com/monica/ in .env

**Comment by BranTurner** [comment]
@adriangibbons You should serve the `index.php` file in the `public` directory as the access point to Laravel applications. I don't believe that Monica will support subdirectories at present, so this is something that needs to be fixed. I aim to get around to it at some point!

**Comment by GaneshKandu** [comment]
need to change all ```*.blade.php``` all file of ```resources/views``` 
```html
<img class="logo" src="/img/small-logo.png" alt="">
``` 
to 
```html
<img class="logo" src="{{ url(/img/small-logo.png) }}" alt="">
```
and
```html
<form class="" action="/login" method="post">
```
to
```html
[...]

PR #267, opened 2017-06-13
Change URLs to use named routes

BranTurner:
I have changed URLs to use named routes, and changed hardcoded assets to use Laravel's `helper` function. This should solve #139 and make links more robust. 

All major functions have been tested and all seem to work.


adriangibbons:
Hi - same issue for me...

I can't use https://domain.com/monica/server.php though perhaps I am doing something wrong? I'm assuming I should go for server.php as otherwise I just get a directory listing...

Have set APP_URL=https://domain.com/monica/ in .env

peperunas:
Why is it closed?

BranTurner:
@peperunas Because I ran into some problems I only realised after submitting the PR and didn't have time to fix. I will fix and resubmit at some point.

PR #278, opened 2017-06-14
Fix LinkedIn profile field not updating

tudor-gala:


app/Http/Controllers/PeopleController.php
```diff
@@ -205,6 +205,12 @@ public function update(Request $request, $contactId)
             $contact->twitter_profile_url = null;
         }
 
+        if ($request->input('linkedin') != '') {
```
Phrohdoh: Should probably `trim(..)` this before checking against an empty string.

tudor-gala: For the sake of consistency I just followed the exact same pattern that the other fields had, as I didn't want to change more lines than needed, should I just trim() all the other field inputs as well?

Phrohdoh: Will defer to @djaiss then. I hadn't taken the time to look at the other fields, sorry about that.

My opinion is yes trim all in another commit.

djaiss:
Thanks a lot. This is trickier than this actually.

I've just forgotten to remove the field from the Edit personal information view.

LinkedIn is now associated to the Work information section. To update it, you need to edit the Work information - not the Personal information.

I've just pushed a fix in production to change this - LinkedIn is now completely out of the Personal information section.

In my opinion, the Personal information shouldn't list LinkedIn.

That being said, thanks a lot for caring and submitting this PR. I truly appreciate it.

PR #280, opened 2017-06-14
Add ability to edit debts

tudor-gala:


Phrohdoh:
How is update different from edit (semantically)? Why have two methods?

app/Http/routes.php
```diff
@@ -84,6 +84,8 @@
 
         // Debt
         Route::get('/people/{people}/debt/add', ['as' => '.debt.add', 'uses' => 'PeopleController@addDebt']);
+        Route::get('/people/{people}/debt/{debtId}/edit', ['as' => '.debt.edit', 'uses' => 'PeopleController@editDebt']);
+        Route::post('/people/{people}/debt/{debtIt}/update', ['as' => '.debt.update', 'uses' => 'PeopleController@updateDebt']);
```
Phrohdoh: `debtIt`?

tudor-gala:
Conventionally, Laravel uses edit when referring to the action of showing the user the edit form for a certain resource, and update when referring to the POST action that actually updates that resource with new data.

app/Http/routes.php
tudor-gala: Good catch! That was a typo, fixed it now

app/Http/Controllers/PeopleController.php
```diff
@@ -1540,6 +1556,38 @@ public function storeDebt(Request $request, $contactId)
...
+            return redirect()->route('people.index');
+        }
+        
+        $debt->in_debt = $request->input('in-debt');
+        $debt->amount = $request->input('amount');
+        $debt->reason = $request->input('reason');
+
+        $debt->save();
+
+        $contact->logEvent('debt', $debt->id, 'edit');
```
Phrohdoh: In https://github.com/tudor-gala/monica/blob/e1e08d349210df7be2937785a53b99a8c3dcba16/app/Http/Controllers/PeopleController.php#L150 this is logged as `update`, not `edit`.

IMO this should be `update` given the above.

Phrohdoh:
LGTM with my nit addressed 👍

app/Http/Controllers/PeopleController.php
djaiss: Indeed, this should be `update`. If it's `edit` it's gonna raise an error on the Dashboard.

tudor-gala: Fixed this. Also changed the docblock for logEvent to correctly indicate that 'add', 'update', or 'delete' should be used :-)

resources/lang/en/people.php
```diff
@@ -276,5 +276,7 @@
     'debt_add_amount' => 'the sum of',
     'debt_add_reason' => 'for the following reason (optional)',
     'debt_add_add_cta' => 'Add debt',
+    'debt_edit_update_cta' => 'Update debt',
```
djaiss: I'm going to be super annoying, but you've introduced new keys, which is not a problem - but they new keys should also be put on the other translation files as well (you don't need to translate them, just paste them as is).

tudor-gala: No problem, you're not annoying at all, I actually appreciate you taking the time to explain these things so I'll know in the future!
I've added the missing translations in a new commit.

djaiss:
This is amazing. Thanks for your work.

PR #284, opened 2017-06-14
Implement edit note functionality

tudor-gala:


djaiss:
❤️ 
Thanks a lot, it's awesome.

Phrohdoh:
So edit is `note/1234/edit` while delete is `notes/1234/delete` (singular vs plural), is this desired or conventional?

djaiss:
@Phrohdoh good catch. Should be `notes`.

tudor-gala:
I saw a mixture of 'note' and 'notes' being used, so I just went with note.
Also while some routes use singular ( /significantother/ /kid/ /debt/ ) others use plural ( /gifts/ /activities/ /reminders/ ).
My personal preference is singular (I think it sounds more natural 'note add', 'note edit', 'note delete' - you're adding/editing/deleting a single note at a time, not multiple, hence 'notes delete', 'notes add', 'notes edit' doesn't make as much sense to me 😄 )
But in the end it's up to you @djaiss, just let me know which you prefer and if you want me to update the notes route URLs or all the other routes URLs.

djaiss:
@tudor-gala I agree. We should stick to one thing, whether plural or singular. Perhaps singular reads better.

Issue #263, opened 2017-06-13
error importing a vcard file


Nebucatnetzer:
I tried to import a vcard file exported from Thunderbird however I got this error:

`The provided vcard file was not found or is not valid!`

Is there some way to validate the file?

yamakadi:
The file needs to have .vcf as it's extension and must be located somewhere relative to the base directory.

For instance, if it's in storage, you need to put 'storage/contacts.vcf'. Home directory also works, like '~/Desktop/contacts.vcf'.

Nebucatnetzer:
I have a test.vcf and used this command:
sudo artisan import:vcard myemail@something.com ~/ansible/test.vcf

as well as:
sudo artisan import:vcard myemail@something.com /home/ansible/test.vcf

both failed.

yamakadi:
Could you provide the file by any chance so I can check the issue? And can
you also try to put it in the base directory of your Monica installation
and just try with 'test.vcf'?

Nebucatnetzer:
I could create a new version with a test contact and send you that.

I give that a try.

Nebucatnetzer:
It worked when I moved the file to the installation root.
I'll create a PR to update the documentation for this.

Nebucatnetzer:
@djaiss does the vcard import work for you if the file is not in the
root of the Monica installation?

yamakadi:
For me, it only works if I use ~ as home folder. @djaiss also used that
successfully I think. It's not meant to work that way though. It appends
'./' to the address and uses the root directory as base directory.

Nebucatnetzer:
Okay, I was asking because I wasn't sure if it really makes sense if I update the readme
or if it's just a quirk with my installation.

However it's probably safer when we write that the file should be put into the
installation root?

yamakadi:
Yup!

djaiss:
This has been pushed to production (see #302 ).

PR #302, opened 2017-06-15
update the section about the vcf import

Nebucatnetzer:
On some systems it's not possible to import a .vcf file if it isn't in the root
of the installation. To reduce issues about imports that don't work I updated
the readme with a way that should work on every system.
#263


djaiss:
❤️

Issue #98, opened 2017-06-07
Bug: Wrong validation when you edit a child with less than 1 year old based on birthday


jaindsonvs:
Steps to reproduce:

1. "Add a child" to one contact and use option "I know the exact birthdate of this child, which is", so that the child is less than 1 year old (ex: 05/01/2017).
2. Fill other informations and save.
3. Try to edit this child and hit save (you don't even need to change anything).
4. App will provide an error validation about "This child is probably", even though it is not used: "The value must be greater than 1"

Expected behaviour: No error validation about "This child is probably" field.

Nebucatnetzer:
I can confirm this bug.
It doesn't let you save the form unless you increase the age to something >1.
Even when you select one of the other options.

PR #304, opened 2017-06-15
allow children to be younger than 1

Nebucatnetzer:
Children born this year have an age of 0. Currently it's not possible
to enter an aproximet age of 0.
This commit allows to do this.

#98


djaiss:
The codebase has just changed dramatically in the latest commits. Can you solve the conflicts? Thanks a lot!

Nebucatnetzer:
I have to create a new PR because I deleted the repository this is based on.
I'm not that used to working with git :)

PR #308, opened 2017-06-15
Refactor SettingsController for readability

yamakadi:
I believe this is more readable and extendable. Plus, if the SQL export function is outside the controller as a job, it can be reused elsewhere if necessary.


djaiss:
Thanks for this!

Deleting an account leads to an error now.

screen shot 2017-06-15 at 3 25 42 pm

yamakadi:
I have no idea why it would try to run an insert request. I'll try to fix
it later today.

yamakadi:
The user wasn't getting deleted completely and logout caused some issues, but it should all be fixed now.

PR #320, opened 2017-06-16
Small refactor of CSV importing

yamakadi:
You can use user emails now.
The validity of path and user will also be checked.

### DIFF
--- a/app/Console/Commands/ImportCSV.php
+++ b/app/Console/Commands/ImportCSV.php
@@ -13,7 +13,7 @@ class ImportCSV extends Command
      *
      * @var string
      */
-    protected $signature = 'import:csv {user} {file}';
+    protected $signature = 'import:csv {user : email or id of a user} {file : path to the CSV file} {--format=google}';
 
     /**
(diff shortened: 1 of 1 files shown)


djaiss:
Thanks that's great.

Could you update the README as well which mentions this feature?

yamakadi:
I don't have any more time today, but I'm planning on expanding this with different drivers (Google, Outlook, etc.) later. It might be better to merge this now and add it to the readme then.

djaiss:
Awesome. Thank you.

cthurber:
Hi! Looking at working on a derivative of this function to handle importing contacts exported to CSV from LinkedIn (In undergrad, and have a lot of student connections on there I'd like to import). 

Curious if this is stable to add the button to the page similar to the existing VCard import feature?

cthurber:
Just finished up a slight derivative of this to allow LinkedIn contact exports (CSV) to be imported as well. Import claims to be successful, and I see the rows in the database with my user id but they aren't populating for my user...

PR #342, opened 2017-06-17
Allow Postgres database connection when using the docker image

supergillis:
Added two dependencies in the Dockerfile, i.e. `php7-pgsql` and `php7-pdo_pgsql`.

Use `${DB_PORT}` environment variable in `docker/waitfordb.sh`.


Phrohdoh:
Is there any way to not install those packages if we aren't going to use pg?

docker/waitfordb.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/sh
 
-echo "Connecting to ${DB_HOST}"
+echo "Connecting to ${DB_HOST}:${DB_PORT}"
 
 while true; do
-    nc -z ${DB_HOST} 3306 && break
-    echo "Unable to connect to MySQL; sleeping"
+    nc -z ${DB_HOST} ${DB_PORT} && break
+    echo "Unable to connect to database; sleeping"
```
Phrohdoh: Might as well make these `database`s `${DB_HOST}`.

supergillis:
Does not seem worth the trouble of detecting the database driver and installing then at runtime as the packages are relatively small. 

https://pkgs.alpinelinux.org/package/edge/community/x86_64/php7-pgsql
https://pkgs.alpinelinux.org/package/edge/community/x86_64/php7-pdo_pgsql

djaiss:
Thanks a lot for this PR!

PR #369, opened 2017-06-22
Hide gift value, link and comment when not set

alexcrooks:
This commit updates the gift UI to hide the `value`, `link` and
`comment` blocks when the values are not set. The usage of `is_null`
does not work because when the values are empty strings when not set.


djaiss:

alexcrooks:
💯

PR #376, opened 2017-06-22
fix content table seeder

Lovatt:
Fixes error when trying to seed fake notes, also removed logEvent as method doesn't exist on this class.


djaiss:
Ah ah damn. I actually fixed it myself a **mere minute ago**

Sorry about this waste effort, but thank you very very much for your help.

PR #388, opened 2017-06-24
issues/345 - /people slow with many contacts

zach2825:
This pull request currently includes only the basic pagination for the /people page

Added pagination to the /people page. Should help with slow page load with a lot of results. Tested this with 10,000 contacts on 1 account.

* new migration to create the new display_count column in the users table defaults to 50
* added setting to allow the user to select which will allow the user to choose how many results to display, defaults to 50
* new config/view default display_count value
* in the people controller changed ->get to ->paginate and added the laravel pagination output in the view
* in the users model i added an accessor to assure we don't get an empty display count value. which could cause an issue if empty

**We might want to consider adding another index in the kids table on the column child_of_contact_id. What do you guys think about that?**


djaiss:
Thanks a lot for the PR.

I like having a pagination - I dislike having a settings for it, and having a new env variable for it. I don't think we should go through that route. I'll have to document that actually, that two things should be avoided when we can: having a new ENV variable, and having a setting. This just complicates the product and if we do it for every option, we'll have a bloated product.

I want to paginate the results, but I think we should aim for the upmost simplicity. As a first improvement I think we should NOT have a setting for this, and put a limit like 50 items. That's it.

Then, we'll see over time if this deserves to have its own setting.

zach2825:
Do you just want to hard code the 50 for pagination? I can remove the other stuff for the pull request. but, I want to see how much to remove. I do have a biased against hard coded counts but I can remove it for this PR.

I'm planning to remove all display_count preferences stuff and hard code 50 for the display count, which will still be too much for people on small devices or slow internet

erdmenchen:
There are two ways to improve performance on the people page:
* Remove duplicate database queries (e.g. debug shows that the query "select * from `accounts` where `accounts`.`id` in ('1')" is executed multiple times)
* Reduce the displayed data. Pagination is the right direction but maybe 50 is still too much. Reducing it to e.g. 25, could be better. From usability perspective you do not need that amount of people displayed at once, because you would want to search for a particular one.

Just keep in mind, that this only depends on server performance, not client. Rendering hundreds of rows is not the bottleneck, but the database queries are.

Another big usability improvement would be to load the data async via javascript and have the page itself rendered first. But this is a mucch advanced step, since we would need an AI for this approach.

zach2825:
I have updated this pull request with a simple pagination change

djaiss:
Thanks for this.

Can you please remove the pagination at the top?

image

zach2825:
I removed the top pagination

djaiss:
I can finally go back to this. Sorry for the very long delay.

Issue #392, opened 2017-06-25
Improve the journal's entries style: respect the white-space


harokb:
When writing a journal entry in Monica, the white-space is completely ignored, so that everything is in one continuous block of text (no newlines, for example). Here's an example:

![Editing](https://user-images.githubusercontent.com/4773657/27515506-8472a4a8-599c-11e7-8f79-f3071b8452f0.png)

turns into

![Result](https://user-images.githubusercontent.com/4773657/27515510-9d6585f2-599c-11e7-9110-1e707a4b1351.png)

As you can see, the newlines aren't respected, which hurts readability. I'll write a fix very soon and send a pull request.

PR #393, opened 2017-06-25
Journal entries respect whitespace in content (#392)

harokb:
This should fix issue #392. I tested the solution by changing the HTML directly in Chrome, and it worked, but I haven't tested through Homestead, but it should work since the problem is really simple.


djaiss:
Thanks, but your fix breaks the layout.

image

harokb:
My bad, I added a second opening tag instead of a closing tag. It's now fixed. 

I tested it locally, and the layout seems correct:

image

djaiss:
Thanks a lot!

Issue #332, opened 2017-06-16
Currency still appears on $ on contact page even if it's set to another currency


djaiss:
Quick note on currency - while I can change the currency in settings and have it show up on the Debts page as £, on the contact profile page it still shows up as $

w00p:
The same goes with dashboard page.

PR #394, opened 2017-06-25
Format monetary values with currency symbol consistently (#332)

rpalladino:
This fixes issue #332:
 
- Add a MoneyHelper class, with a unit test
- Alias MoneyHelper to be available in global namespace in templates
- Remove hardcoded '$' from translation files
- Modify templates to use MoneyHelper to format debt and gift amounts
- Fix Tests\TestCase constructor to accept arguments of original
implementation (unrelated to #332)


djaiss:
Big love for this. Thanks a lot!

PR #396, opened 2017-06-26
Please Close  - 270 storage files in aws s3

degan6:
Please close, refer to #547 

Support for storing avarts in AWS S3

Added database column to keep track of avatars per contact. Defaults to public (same as now) so all instances will continue to work without issue.

To change to s3:

1. Create a bucket, [make it public and webhost](http://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html)
2. Update your .env from the new .env.example
3. Add your AWS keys, bucket name, and region to .env
4. Change the `DEFAULT_FILESYSTEM` from `public` to `s3`

All files that were uploaded before the switch will continue to be loaded from the local disk while all new files will be uploaded and served from s3.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -75,8 +75,8 @@ CACHE_DRIVER=database
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
 
-#Default fileststem to update contact images too
-#S3 Requires AWS Keys configured and a public bucket created
+# Default fileststem to update contact images too
+# S3 Requires AWS Keys configured and a public bucket created
 # Two choices: public|s3.
 DEFAULT_FILESYSTEM=public
 


djaiss:
Thanks for this. Can you check the failed build? 

image

.env.example
```diff
@@ -74,3 +74,13 @@ PAID_PLAN_PRICE=
 CACHE_DRIVER=database
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
+
+#Default fileststem to update contact images too
```
djaiss: It's a small detail but can you add an ` ` between `#Default` so it becomes `# Default`? Same for line 79.

degan6: No problem

djaiss:
I finally have had some time to review the PR. I'm going to add some comments to the PR. Sorry about the time it took to get back at you, that wasn't my intention.

.env.example
```diff
@@ -74,3 +74,13 @@ PAID_PLAN_PRICE=
 CACHE_DRIVER=database
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
+
+# Default fileststem to update contact images too
```
djaiss: Can you change this to

```
# Default filesystem to store uploaded files.
# Possible values: public|s3
```

.env.example
```diff
@@ -74,3 +74,13 @@ PAID_PLAN_PRICE=
...
+# S3 Requires AWS Keys configured and a public bucket created
+# Two choices: public|s3.
+DEFAULT_FILESYSTEM=public
+
+AWS_KEY=
```
djaiss: Above this, indicate 

```
# AWS keys for S3 when using this storage method
```

database/migrations/2017_06_25_200318_add_contact_avatar_location.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('contacts', function (Blueprint $table) {
+            $table->enum('avatar_location', ['s3', 'local'])->after('avatar_file_name')->default('local');
```
djaiss: Enum will break postgresql migrations. We need to avoid them. Can you rewrite the migration without them, and perhaps simple add `$table->string` type?

nightexcessive:
I'd be interested in this, but with Google Cloud Storage. My use case is that I'd like to run Monica on App Engine's flexible environment – it'd be cheaper than running it on a compute instance because it would only start when requests were made.

It seems that the author of this PR has abandoned it. Would you be interested in a PR that adds support for Google App Engine, probably through the use of [shpasser/GaeSupportL5](https://github.com/shpasser/GaeSupportL5)?

database/migrations/2017_06_25_200318_add_contact_avatar_location.php
degan6: Sounds good. I'll update this tonight.

Issue #240, opened 2017-06-12
Encoding error in reminder emails


ajdavis:
I added a reminder on the MonicaHQ website, `Ask how the summer's going`, and received an email, `Ask how the summer&#039;s going`.

ajdavis:
Thanks!

PR #397, opened 2017-06-26
Display reminder title and description unescaped (#240)

rpalladino:
This fixes issue #240. 

Displaying the title and description unescaped prevents characters such as single quotes from being displayed as html entities. Since this is a plain text email, XSS should not be a concern.


djaiss:
Good one. Thank you.

PR #404, opened 2017-06-27
fix bug in latest actions

Lovatt:
Fixes bug in latest actions that was causing contact_complete_name to return null.


w00p:
i think you have to change it to: 
`'contact_complete_name' => $event->contact->getCompleteName(auth()->user()->name_order)`
Or else the settings where you chose first name first, og last name first wont work.

Lovatt:
@w00p again great point 😃  i've added in the user preference check, should sort that issue.

djaiss:

PR #406, opened 2017-06-27
show time diff from when action was created to now

Lovatt:
Changed the way the 'Last Actions' tab shows event timestamps, at first glance it's hard to reason how long ago an action took place without counting backwards from the current date, updated to use Carbons diffForHumans.

**Before**
screen shot 2017-06-27 at 20 36 56

**After**
screen shot 2017-06-27 at 20 36 38

### DIFF
--- a/app/Http/Middleware/CheckLocale.php
+++ b/app/Http/Middleware/CheckLocale.php
@@ -3,6 +3,7 @@
 namespace App\Http\Middleware;
 
 use Auth;
+use Carbon\Carbon;
 use Closure;
 
 class CheckLocale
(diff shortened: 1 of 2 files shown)


w00p:
This does not take into consideration which language/locale you have chosen. It will always show "hours ago", and not "timer siden" in Norwegian. I think the best way would be to make a new function in the DateHelper that checks what local a user use.

Lovatt:
@w00p good point! i've added in a call to Carbon::setLocale during the middleware check, this should sort that problem.

w00p:
But, you are referencing to app.config, it shout be: 
`Carbon::setLocale(Auth::user()->locale);`
Its the user that sets the local.

Lovatt:
If you look at app/Http/Middleware/CheckLocale.php, before i call Carbon::setLocale the app.locale is set using App::setLocale(Auth::user()->locale) so they are now the same thing

djaiss:
Thanks for this PR ❤️ , it makes a lot of sense

PR #412, opened 2017-06-28
Use @php instead of php in composer.json

rndmh3ro:
The composer post-install scripts use a hardcoded `php` to run scripts.
I have two different versions of php installed (5 & 7):
```
root@localhost:/var/www/monica# php -v
PHP 5.6.30-0+deb8u1 (cli) (built: Feb  8 2017 10:31:58) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
```
```
root@localhost:/var/www/monica# /usr/bin/php7.0 -v
PHP 7.0.20-1~bpo8+1 (cli) (built: Jun 10 2017 20:21:39) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.20-1~bpo8+1, Copyright (c) 1999-2017, by Zend Technologies
```

Now if I run `/usr/bin/php7.0 composer.phar install`, composer will be run with php7.0 but the scripts will run with php5 since that is what gets called when I run `php` instead of `/usr/bin/php7.0`.
This PR fixes that in the way that now the php process being currently used will execute the scripts, not a new process.

See the docs of composer:

> To execute PHP scripts, you can use @php which will automatically resolve to whatever php process is currently being used:
- https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts


djaiss:
Thanks for this. I didn't know about it. I'll merge this.

PR #413, opened 2017-06-29
Fix display of reminders on people detail view (#244)

### LINKED ISSUE #244: Adding a note to a reminder breaks the table
Steps:

* Add a reminder and type a note
* People detail view breaks

schermafbeelding 2017-06-12 om 20 31 58

rpalladino:
This PR removes comments from displaying in the reminders table on the person detail page, to fix #244. We may want to display reminders comments in the table, but its not clear how that should be done. Other entities that display here (e.g. activities, tasks) do not display the comment/description. 

### Checklist


djaiss:
Thanks a lot for this PR. But somehow I'm torn with it. Yes, currently it breaks, but at least the information is there. What you've done is removing the problem, not really fixing it. A true fix would be to keep the information and display it in a way that wouldn't break the layout.

rpalladino:
How about this, then?

screen shot 2017-06-30 at 10 44 27 am

I've modified the reminders to display consistently with the other things on the people detail view (activities, gits, etc), by replacing the table with an unordered list in the template markup. I also cleaned up the now obsolete table header strings, removing them from the translation files.

djaiss:
Why did you add the blue stuff around the frequency? I don't think it adds value and the padding is also off here.

image

rpalladino:
@djaiss That was unintentional and I've fixed it.

djaiss:
Nice thank you very much for your efforts.

PR #415, opened 2017-06-30
Speed up contact list display

djaiss:
This PR drastically speeds up the display of the contact list.

What took a lot of time to display the list when you had a lot of contacts, was the fact that for every contact, we were looking if there was a gravatar for the contact. That means one request to an external service per contact. The nightmare.

To solve this, on page load, we no longer check Gravatar. Gravatar is only requested once now: when you add the email address for a contact. The benefit: we no longer request Gravatar on every page load. The drawback: we'll look for a gravatar only if you add an email address to a contact.

Also, because now the list loads really fast, there is no need for a pagination anymore, which we introduced one hour ago thanks a pull request from the community.


app/Http/Controllers/PeopleController.php
```diff
@@ -29,7 +29,7 @@ public function index(Request $request)
             $user->updateContactViewPreference($sort);
         }
 
-        $contacts = $user->account->contacts()->sortedBy($sort)->paginate(25);
```
zach2825: Is there a reason to remove pagination from here?

PR #464, opened 2017-07-12
Trim contact names

a3020:
I've added mutators to the Contact model to prevent having to put the same logic in multiple places.

Notes:
- `null` values are still respected.
- `ucfirst` behavior for last_name remains the same (only ucfirst on store, not on update).


itsthejoker:
@a3020 your PR is out of date; might want to update it.

bw-adri:
Aight, done

PR #476, opened 2017-07-18
BRL added to currencies

nerijunior:
BRL added to currencies

### DIFF
--- a/app/Console/Commands/SendNotifications.php
+++ b/app/Console/Commands/SendNotifications.php
@@ -50,13 +50,9 @@ class SendNotifications extends Command
             $contact = Contact::findOrFail($reminder->contact_id);
             $account = Account::findOrFail($contact->account_id);
             $user = User::where('account_id', $account->id)->first();
-            $reminderDate = $reminder->next_expected_date->hour(0)->minute(0)->second(0)->toDateString();
+            $date = $reminder->next_expected_date;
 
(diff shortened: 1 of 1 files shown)


database/migrations/2017_06_10_155349_create_currencies_data.php
```diff
@@ -26,6 +26,7 @@ public function up()
       DB::table('currencies')->insert([ 'iso' => 'GBP', 'name' => 'British Pound', 'symbol'=>'£' ]);
       DB::table('currencies')->insert([ 'iso' => 'EUR', 'name' => 'Euro', 'symbol'=>'€' ]);
       DB::table('currencies')->insert([ 'iso' => 'RUB', 'name' => 'Russian Ruble', 'symbol'=>'₽' ]);
+      DB::table('currencies')->insert([ 'iso' => 'BRL', 'name' => 'Brazilian Real', 'symbol'=>'R$' ]);
```
djaiss: Instead of modifying this file, you should use a new migration, like this one (https://github.com/monicahq/monica/blob/master/database/migrations/2017_06_14_132911_add_zar_currency_to_currencies_table.php). If we do like you've done, no existing installations will contain this new currency, because this migration has already been executed.

nerijunior: You are right I'll change to a new migration

Issue #51, opened 2017-06-06
Feature Request: Add search bar in navigation


tylerbmcsilva:
It would be nice to have the "search" function that is able to pull from all data on Monica

djaiss:
I agree. I'll do that in the future for sure.

HHErebus:
I'd like to have a stab at this if possible. @djaiss are you working on this already?

I think there's a few things open for debate:

1.  **Placement in the UI:** my personal experience with using Monica would put this in the top navigation as @tylerbmcsilva suggested, visible on _all_ pages. I think it's important enough as a quick way to retrieve the information stored in the db to warrant that spot.

2. **Style:** I think the only sensible way to integrate it in the top nav, if we end up deciding that, is to do so like [Laravel's documentation](https://laravel.com/docs/5.4), e.g. an "invisible" input.

3. **Where to search:** I feel like searching on all data by default would be overkill, but I do see how searching tables other than contacts would be useful. I don't have a strong opinion either way so I'm open to guidance on the matter.

This seems like a fairly straightforward feature, so I chose it for my first contribution to Monica. Differing opinions are welcome.

djaiss:
Thanks @HHErebus for your contribution.

I think it'd be a great feature. In terms of design I don't think it should be an invisible field. Check the design idea below.

![artboard 3](https://user-images.githubusercontent.com/61099/28192910-2f0bf842-6807-11e7-8297-3bcc5f80a31a.png)

![artboard 3](https://user-images.githubusercontent.com/61099/28192917-3a4e94b2-6807-11e7-9c30-1edee3b992e0.png)

About searching data themselves, I'd say let's search for contacts first. It'd be like a "quick jump" feature. I think that's the simplest first iteration we can do.

HHErebus:
That's great, many thanks for the quick turnaround :D On mobile, I think I can just put it below the list of nav actions (i.e. it would be the last item in the nav), wdyt?

djaiss:
This is done and in production.

PR #481, opened 2017-07-18
51 feature search contacts

HHErebus:
PR for [issue 51](https://github.com/monicahq/monica/issues/51)

Everything should be fine, testing on local env and in my self-hosted version was successful. Hooray for first PR (and hopefully first accepted PR  too :D )!

### Changes
- **Added search input in nav**, activates after 200ms of the last keystroke or on submit

- **_Searchable_ trait** should make adding this functionality to other models fairly straightforward

- **CSRF token added to skeleton.blade.php** to allow for ajax forms

- **Icons' zIndex has been reset**, as no issues with the default value were found (using which, in turn, prevents them from appearing above the search results and forcing adding a custom zIndex value to them)

- **Basic Unit tests** for search successful and empty. I encourage any reviewer to look at these especially since they're very basic - I'm learning about unit testing, so they might not be at the level you want them to be.

- **Translation entries added** to all 4 current translations, but are in English across the board as I don't know any of the other languages.

### How the feature works

1. Type in search bar. 
2. After 200ms (or when submitting the form) a POST will be sent to the `people/search` route
3. Through `PeopleController`, the DB is queried.
3a. The query is made looking for any match of the "needle" into the columns defined in the related model, in the `$searchable_columns`.
3b. The query then `SELECT`s only the columns defined in the `$return_from_search` variable, also defined in the related model.
3c. If no model is found, the "no result found" message from the correct translation file is sent back.
4. The js that sent the request parses and shows the results.

### Notes
- I didn't use typeahead.js (even though it appears in the project's dependencies) because it doesn't seem to be actually used in Monica.

- Not sure if this is the smallest _possible_ change to the code (hence the missing mark), but I think it's self-contained enough to qualify.

- On sm and xs display sizes I put the bar at the top - having it as the last item in the list (more or less at the center of the page) seemed like a mis-logout waiting to happen.

### Checklist

- [ ] Make sure that the change you propose is the smallest possible.

### Screenshots
Desktop
image
---
Desktop, no results
image
---
Destkop, some results
image
---
Mobile
image
---
Mobile, some results
image


djaiss:
Thanks so much for your effort and for caring about the project. I'm going to review all this in a few hours.

djaiss:
> I didn't use typeahead.js (even though it appears in the project's dependencies) because it doesn't seem to be actually used in Monica.

Indeed. It was used in the past but not anymore. [This](https://github.com/monicahq/monica/issues/482) will take care of that.

resources/assets/js/search.js
```diff
@@ -0,0 +1,147 @@
...
+        // We want typeahead-like behaviour, but not a query for every keystroke.
+        // We submit a query 200ms after a key has been pressed.
+        search.input.on('keyup', function () {
+            if (search.timeoutId !== undefined) {
+                window.clearTimeout(search.timeoutId);
+            }
+            search.timeoutId = window.setTimeout(searchInContacts, 200);
+        });
+
+        search.input.on('submit', function (e) {
```
djaiss: Can you add a check here? At the moment, if we enter something and press Enter, the Search will be triggered and we'll receive an error page.

HHErebus:
this last commit should do it 👍

djaiss:
Nice. This looks awesome. I'm so happy about this. Thanks a lot! I'll merge this back from work tonight.

HHErebus:
Thank _you_ for Monica, it's shaping up to be a great tool :D I'll get started on other useful additions and fixes soon, hopefully :)

djaiss:
@HHErebus it works locally, it doesn't work in production. That's the same code. I don't get it. Trying to debug.

djaiss:
I've found the issue. For some reasons `vanilla.js` file was not refreshed, so I added versioning for the file and it's working.

That being said I've noticed, thanks to this investigation, a HUGE bug in this PR, that I've fixed (not in the [most elegant way](https://github.com/monicahq/monica/pull/489) though): we didn't make a check to restrict the search to your account only. So we could search for ALL the contacts in Monica. We could only see the name though - not load the Contact view itself as it's protected with permissions. Still, a big bug :-D

HHErebus:
@djaiss 😱 using Monica for myself I didn't realize that! Thankfully you spotted that before it went live :S I'll be more thorough with cross-user testing.

As for vanilla.js - I usually expect them to be generated "in place" as a deploy step, should the ones created by elixir be updated in pull requests? Or only some of them? When we clarify this I would add the clarification to the contributing guide

djaiss:
@HHErebus at the moment, all compiled assets are stored in Git. It sucks, I know. One of my task in the future is to compile assets on the server on each deploy BUT there is one huge drawback to that. Remember that Monica is installed on many servers out there, some of them on Heroku. And it's not easy to install Gulp/NPM etc... on Heroku to compile. It is easier to commit the compiled assets on Git, but this creates a lot of conflicts on PR etc... I don't know which solution I should do instead actually :-/

PR #487, opened 2017-07-19
BRL added to currencies and MoneyHelper

nerijunior:
I've added the BRL currency to currencies table and made a change at MoneyHelper to correct format for BRL.


djaiss:
Thanks a lot for this PR. I'll merge this right away!

PR #495, opened 2017-07-23
Add version checking

djaiss:
This PR adds version checking.

If you install Monica on a server that you own, chances are that you will want to be aware when a new version is available. This is what we introduce here.

In short, your Monica instance will send a daily ping to https://version.monicahq.com. The ping contains three information:
* a UUID: the unique, anonymous identifier of your instance.
* the current version of your instance
* the number of contacts in your instance.

The last information is used to know the size of the instances of the people who download Monica. I need to know this information (and only this information) so I can move forward with better knowledge of how people use Monica. Also, there is no way to know who you are: the UUID is random, unique and anonymous.

The code of https://version.monicahq.com is open-source, stored [on GitHub](https://github.com/monicahq/version) like everything in Monica.

In return of this call, you will get a JSON with:
* if a new version is available as a boolean
* what is the latest version available
* all the release notes for each release that has been made since the version of the ping
* the number of versions available since the version of the ping.

This information will be used in the UI. In the footer, a new link will appear if a new version is available. If you click on this link, a popup will appear, with the release notes of the new version.

image

This feature is enabled by default on all instances. You can disable this ping by adding a new ENV variable in your `.env` file: `CHECK_VERSION=false`, and no ping will ever leave your instance.

I hope you will like the feature.

### Checklist

- [X] Screenshots are included if the PR changes the UI.
- [x] Make sure the call is not made when calling from a dev/staging/test environment


djaiss:
Tests will be added later today.

app/Console/Commands/PingVersionServer.php
```diff
@@ -48,6 +54,42 @@ public function handle()
...
 
-        dd($json);
+        $data["uuid"] = $instance->uuid;
+        $data["version"] = $instance->current_version;
+        $data["contacts"] = Contact::all()->count();
+
+        // Send the JSON
+        $jsonData =json_encode($data);
+        $json_url = config('monica.weekly_ping_server_url');
+        $ch = curl_init( $json_url);
```
Phrohdoh: This should probably be async and use Guzzle.

app/Console/Commands/PingVersionServer.php
```diff
@@ -48,6 +54,42 @@ public function handle()
             'contacts' => Contact::all()->count()
         ];
 
-        dd($json);
+        $data["uuid"] = $instance->uuid;
+        $data["version"] = $instance->current_version;
+        $data["contacts"] = Contact::all()->count();
+
+        // Send the JSON
+        $jsonData =json_encode($data);
```
Phrohdoh: Style nit: Missing a space after `=`.

app/Console/Commands/PingVersionServer.php
```diff
@@ -48,6 +54,42 @@ public function handle()
             'contacts' => Contact::all()->count()
         ];
 
-        dd($json);
+        $data["uuid"] = $instance->uuid;
+        $data["version"] = $instance->current_version;
+        $data["contacts"] = Contact::all()->count();
```
Phrohdoh: Can this be `Contact::count();`?

app/Console/Commands/PingVersionServer.php
```diff
@@ -48,6 +54,42 @@ public function handle()
...
+
+        curl_setopt_array($ch, $options);
+        $result = curl_exec($ch);
+        curl_close($ch);
+
+        if ($result == false) {
+            return;
+        }
+
+        if (is_string($result) == false) {
```
Phrohdoh: Why use `== false` instead of `!foo`?

app/Console/Commands/PingVersionServer.php
```diff
@@ -48,6 +54,42 @@ public function handle()
...
+            return;
+        }
+
+        // Receive the JSON
+        $json = json_decode($result, true);
+
+        if ($json['latest_version'] != $instance->current_version) {
```
Phrohdoh: Will this (entire block) throw and cause issues if response JSON is bogus?

djaiss: Indeed. I'll fix this

djaiss:
I went to sleep yesterday instead of finishing this. I'll do it tonight :-D

app/Console/Commands/PingVersionServer.php
djaiss: Nice catch.

PR #518, opened 2017-08-06
Fixed type on settings pages

bradddd:
Changed "than" to "that" across multiple settings.php files


djaiss:
Thanks a lot for your help!

PR #519, opened 2017-08-07
Fix sign up test.

aejnsn:
By default the `.env.example` file's `APP_DISABLE_SIGNUP=true` causes this test to fail. I've added a `putenv()` statement similar to the test below it.


djaiss:
Thanks a lot for this PR!

PR #520, opened 2017-08-07
Add exclusion for routes file to allow code coverage.

aejnsn:
A routes file in Laravel will not contain imports (`use` statements), so when PHPUnit includes this file it will encounter an error like so:
```
Fatal error: Uncaught Error: Class 'App' not found in /home/vagrant/Code/monica/app/Http/routes.php on line 3

Error: Class 'App' not found in /home/vagrant/Code/monica/app/Http/routes.php on line 3
```

However I should note that I only encountered the above error when generating code coverage via:
```
$ ./vendor/bin/phpunit --coverage-html coverage


aejnsn:
@djaiss Pulled this back up to par with master.

djaiss:
@aejnsn Thanks a lot. However I'm not sure I fully understand the impact of this change. I wonder if this would have an impact later down the road?

aejnsn:
This is setup in a default Laravel 5.2 installation. I am guessing this is not present in Monica because it started as 5.2 and but the phpunit.xml was updated to 5.3, in which the routes file location changed to top-level routes directory. Thus there would be no need to exclude, because the phpunit.xml wholly includes the app directory.

https://github.com/laravel/laravel/blob/5.2/phpunit.xml

djaiss:
Thanks a lot for the explanation, they make a lot of sense.

PR #547, opened 2017-08-16
270 storage files in aws s3

degan6:
Support for storing avarts in AWS S3

Added database column to keep track of avatars per contact. Defaults to public (same as now) so all instances will continue to work without issue.

To change to s3:

1. Create a bucket, [make it public and webhost](http://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html)
2. Update your .env from the new .env.example
3. Add your AWS keys, bucket name, and region to .env
4. Change the `DEFAULT_FILESYSTEM` from `public` to `s3`

All files that were uploaded before the switch will continue to be loaded from the local disk while all new files will be uploaded and served from s3. 

### Checklist


degan6:
New Pull Request created because I was dump and deleted my fork that I created the old pull request with. Won't be doing that again...

djaiss:
Thanks a lot, I'll check this at the end of my day (Eastern time).

PR #558, opened 2017-08-20
[wip] 39 login with Facebook and Google

degan6:
Support Login from OAuth Providers!

### Changes to Note
- Create UserSocial model, table, and controller. Tracks connected account per user. Meaning that users can connected many OAuth accounts to their one Monica account. Without making another column in the users table for every service.

This will allows us keep the user token meaning we can send messages via Slack, Facebook and others.

- Set password column in users table to nullable
- Created middleware to check if requested service `monica.app/login/google` for example is enabled and supported. Returns to login screen if not.
- Moved code to create a new account to NewAccount trait used by SocialController and RegistrationController
 

### TODO

- [ ] Add login buttons on login screen
- [ ] Ensure new accounts can't be created when signup is disabled
- [ ] Add error flash messages to responses on login screen when needed (service disabled, no new users)
- [ ] Translate messages mentioned above 
- [ ] list connected accounts in user profile (maybe add this later) Thoughts?
- [ ] Test what happens if a user changes their email inside Monica 
- [ ] Pass all tests

### Checklist


djaiss:
I guess I'll need to create a Facebook application for this.

Do you think it'll work for hosted instances too?

> list connected accounts in user profile

Hum, not sure about this task.

evaryont:
One thing I'd really appreciate is the addition of a Socialite provider that would let me use my own identity server. I did a quick look through Socialite's documentation but couldn't find any obvious example of doing that from environment variables or any other not hardcoded source. In my use-case, I have a [Keycloak](http://www.keycloak.org/) instance running at `https://identity.example.com`, with exposes OAuth 2, OpenID Connect and other protocols.

djaiss:
@degan6 question: why don't you use Socialite? https://github.com/laravel/socialite

degan6:
@djaiss I am using Socialite. If your wondering why I haven't edited composer.json, its because its already socialite is already included.

djaiss:
It's been a few months for this PR and I wonder if you want to continue it or if we should close it for now. Let me know!

degan6:
@djaiss I am bit embarrassed by this, but school just got out for the winter so I will have time compete it by end of the month. 
I'll keep you undated.

djaiss:
We are going to close this pull request. Moreover, I've thought a lot about it, and considering what we've learned about Facebook recently (that we knew already but well...), I've decided:
- we won't support sign up/sign in with external third parties that we don't fully trust (does that even exist?)
- I've already removed most of the tracking that we had on the hosted version.

I know I'm creating a barrier as login with Google or Facebook is really popular and convenient, but I'm really against their practices, so I can't support this.

Anyway, thanks so much for the work already on this. I really hate closing a PR, but I think it's better philosophically.

petrk94:
Well, if you close so many users out, provide at least the manual adding of third OAuth parties to the admin interface, so every Humhub instance can decide by manual what will be used.

For modern solution, in europe is currently a standard in development called [VERIMI](https://verimi.de/en/), where people have more power about their data.
Alternative in development is [Nuggets](https://nuggets.life/) which has Blockchain identity solution.

One Problem Im facing the company Im working in, the extern webservices, even selfhosted once, have to be able to provide SSO/OAuth, so no one always has to sign up to each service with their credentials.

PR #561, opened 2017-08-22
New account limit for free accounts

djaiss:
This PR adds account limitation for free accounts, as well as some nice graphics, and a brand new Welcome screen when you first enter in the product.

For the new accounts, here are the new limitations:
* No reminders by email
* No import.

Also, reducing all the images of the app thanks to ImageOptim.

Exemptions: 
* Existing accounts are not limited - they will have free access forever for everything. 
* People who have downloaded Monica won't have any limits as usual.

- [x] Make sure the check works in every use case
- [x] Tests
- [x] Lower prices drastically
- [x] Remove number of users limitation


degan6:
Wait, In the readme it says that free accounts will have no contact limit?

ziming:
Dissapointed with this move

djaiss:
This is a wip. I need time to think about this, and I decided to publish this PR because well, everything I do is public. No matter what I do and choose, existing accounts are not impacted, and custom installations still have access to everything.

degan6:
Whats frustrating about this is not that you are creating a limit for free accounts but you are creating bad limits for free accounts.

Companies use limits for free accounts all the time. But most of those limits are set in a way where if you run up against the limit you want to upgrade. Google, OneDrive and Dropbox set limits on the amount of data you can store for free. If you hit the limit highly it's likely that the product is important to you and are willing pay for it.

This new limit doesn't let users see the usefulness of the project. Instead it stops them with a paywall right as the are just thinking about adding close friends and family. Right now the average user of MonicaHQ on the hosted version has about 11 contacts. I think we should focus on getting that number higher and create features that make that data moreful to people. 

Then you can set limits on those new and advanced features. 

Some features that you could include for paid accounts include:
-Reminder Notications
-Notifications to SMS, FB, and WhatsApp
-API Access (I have mixed feelings about this one)
-Larger Image Sizes - Maybe reduce the free size now
-Image Galleries - I want to upload more photos as it is
-Auto Import From Google, Twitter, Gmail, Office365

I don't think it's a good idea to limit the core function of the product, especially such a low limit that will turn people away from trying the product.

djaiss:
@degan6 Thanks a lot for your input. You make an astute comment. My goal is to be able to work all day on Monica - this will actually benefit everyone. I can't at the moment because for some reasons I need another source of income to actually live 😀 That means I try to work all evenings and weekends on Monica, even though I have a family. So the sooner I can increase the incomes, the better for **everyone**. That being said, I shouldn't limit the core of the product by limiting contacts. Let's do that at the moment - I'm saying at the moment, because I don't know how it will evolve, and I need to figure out how I can find a solution that works for both users, the project as a whole and me.

PR #571, opened 2017-08-30
Use str_random instead of RandomHelper

ngyikp:
The current string randomiser function is not strong/secure and attackers may be able to guess values, see https://paragonie.com/blog/2015/07/how-safely-generate-random-strings-and-integers-in-php

Laravel has a built-in string randomiser function that uses the secure [`random_bytes`](http://php.net/manual/en/function.random-bytes.php) function, no need to hand-roll a custom one: https://laravel.com/docs/5.4/helpers#method-str-random


djaiss:
I'll review this at the end of my day. Thanks so much for your contribution.

Issue #576, opened 2017-09-05
Installation guide for debian stretch conflicts with php7.1 requirements


Phlogi:
Seems like monica requires php7.1 however debian stretch supports only php7.0. This should be mentioned in the [guide](https://github.com/monicahq/monica/blob/master/docs/installation/debian.md). 

```
composer install --no-interaction --prefer-dist --optimize-autoloader
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for doctrine/annotations v1.5.0 -> satisfiable by doctrine/annotations[v1.5.0].
    - doctrine/annotations v1.5.0 requires php ^7.1 -> your PHP version (7.0.20) does not satisfy that requirement.
  Problem 2
```

Phlogi:
Solution is to set php version to
`"php": "7.0.*"`
in composer.json.

timjrobinson:
This didn't work for me with the latest master. After the change I got:

```
 Problem 1
    - Installation request for doctrine/annotations v1.5.0 -> satisfiable by doctrine/annotations[v1.5.0].
    - doctrine/annotations v1.5.0 requires php ^7.1 -> your PHP version (7.0.19) does not satisfy that requirement.
  Problem 2
    - Installation request for doctrine/cache v1.7.0 -> satisfiable by doctrine/cache[v1.7.0].
    - doctrine/cache v1.7.0 requires php ~7.1 -> your PHP version (7.0.19) does not satisfy that requirement.
  Problem 3
    - Installation request for doctrine/collections v1.5.0 -> satisfiable by doctrine/collections[v1.5.0].
    - doctrine/collections v1.5.0 requires php ^7.1 -> your PHP version (7.0.19) does not satisfy that requirement.
  Problem 4
    - Installation request for doctrine/common v2.8.0 -> satisfiable by doctrine/common[v2.8.0].
    - doctrine/common v2.8.0 requires php ~7.1 -> your PHP version (7.0.19) does not satisfy that requirement.
  Problem 5
    - Installation request for doctrine/dbal v2.6.1 -> satisfiable by doctrine/dbal[v2.6.1].
    - doctrine/dbal v2.6.1 requires php ^7.1 -> your PHP version (7.0.19) does not satisfy that requirement.
  Problem 6
    - Installation request for dompdf/dompdf v0.8.0 -> satisfiable by dompdf/dompdf[v0.8.0].
[...]

PR #577, opened 2017-09-05
Update php version to work on debian stretch and php7.0.X versions

Phlogi:
Avoid dependency errors. Fixes https://github.com/monicahq/monica/issues/576


djaiss:

rndmh3ro:
This only works if the composer.lock is updated, too. Right now, there are still requirements for 7.1 in there: https://github.com/monicahq/monica/blob/master/composer.lock#L154

I have a working (for me) composer.lock here: https://gist.github.com/rndmh3ro/45e17e1fe7f9197375be4918b4ba88c5

However there are many other changes to requirements in there that I can't say work.
Should I create a PR, nonetheless?

Phlogi:
Just run: `composer update` and it should work.

PR #585, opened 2017-09-12
Fix install composer steps on debian

timjrobinson:
The verification hash was incorrect because composer had been updated. It would be better in the long term to link to their documentation.


djaiss:
Thanks a lot for this PR ;-D Gonna merge it now.

PR #595, opened 2017-09-23
Fix migration

LongDuongBIT:
### Checklist

- [ ] Read the [CONTRIBUTING document](https://github.com/monicahq/monica/blob/master/CONTRIBUTING.md) before submitting your PR.
- [ ] If the PR is related to an issue or fix one, don't forget to indicate it.
- [ ] Make sure that the change you propose is the smallest possible.
- [ ] Indicate `[wip]` in the title of the PR it is is not final yet. Remove `[wip]` when ready. Otherwise the PR will be considered complete and rejected if it's not working.


djaiss:

a3020:
Why is modifying an existing migration considered to be a good thing? Wouldn't it make more sense to create a new migration for things like these?

djaiss:
It's not a good thing if it's a completely new behaviour or if it changes the database.
Here that's not the case. This is why I accepted it, but you are right, we should have done a new migration for it.

a3020:
> or if it changes the database

It does change the database. Now for some people the field is `nullable` and for some not. Meaning the database structure can differ per installation.

PR #597, opened 2017-09-27
Upgrade frontend build system to mix

mauricew:
### Overview
Laravel now uses a slimmer build system that makes the process smaller and simpler. Gulp and bower are both gone, and npm handles all dependencies while webpack handles the build process. Absolutely no actual code should change or upgrades to dependencies made. Some unused files were also removed.

If there are any additional changes that should be made for technical reasons, let me know.

### Checklist

- [ ] Make sure that the change you propose is the smallest possible. _(It isn't.)_
- [X] Screenshots are included if the PR changes the UI. (It should not)
- [X] Tests added for this feature/bug. (no actual site changes)
- [X] Indicate `[wip]` in the title of the PR it is is not final yet. Remove `[wip]` when ready. Otherwise the PR will be considered complete and rejected if it's not working.


djaiss:
@mauricew I totally disregarded this PR. So sorry about that.

Let me read more about mix to see how it will affect the codebase.

mauricew:
Two quickest things I can say about it are
* Use of webpack for all of the css & js - You can get rid of gulp if you want, but more importantly bower being removed is a good thing.
* When the js files are regenerated, the file name doesn't  change - it gets appended to the `rev-manifest.json` file. This is pretty good for source control.

mauricew:
Oops I accidentally closed it...

djaiss:
@mauricew 

> When the js files are regenerated, the file name doesn't change - it gets appended to the rev-manifest.json file. This is pretty good for source control.

Hum. Name won't change? That might be a problem for cache invalidation.

mauricew:
It gets appended when being served through http, just not directly in the filesystem.

Before: /build/css/app-1234567.css
After: /css/app.css?id=1234567 (autogenerated whenever it changes)

package.json
```diff
@@ -1,18 +1,28 @@
...
+    "list.js": "^1.5.0",
+    "jquery": "^3.2.1",
+    "jquery-tags-input": "git://github.com/xoxco/jQuery-Tags-Input#v1.3.6",
+    "laravel-mix": "^1.0",
     "lodash": "^4.16.2",
-    "vue": "^2.0.1",
-    "vue-resource": "^1.0.3"
+    "octicons": "^4.3.0",
+    "tachyons": "^4.8.1",
+    "typeahead.js": "^0.11.1"
```
djaiss: We are missing `"vue": "^2.1.10"` in this file.

resources/assets/js/app.js
```diff
@@ -6,26 +6,11 @@
...
-      activities_description_show: false,
-      reminders_frequency: 'once',
-      accept_invite_user: false,
-    },
-    methods: {
-    },
-});
+require('./tags');
+require('./search');
+require('./contacts');
```
djaiss: We need to add 
```
window.Vue = require('vue');

const app = new Vue({
    el: '#app',

    data: {
        activities_description_show: false,
        reminders_frequency: 'once',
        accept_invite_user: false,
    },
        methods: {
    },
});
```

here, otherwise we'll break some screens.

package.json
```diff
@@ -1,18 +1,28 @@
...
+    "font-awesome": "^4.7.0",
     "gulp": "^3.9.1",
-    "jquery": "^3.1.0",
-    "laravel-elixir": "^6.0.0-14",
-    "laravel-elixir-vue-2": "^0.2.0",
-    "laravel-elixir-webpack-official": "^1.0.2",
+    "hint.css": "^2.3.2",
+    "list.js": "^1.5.0",
+    "jquery": "^3.2.1",
+    "jquery-tags-input": "git://github.com/xoxco/jQuery-Tags-Input#v1.3.6",
```
djaiss: I had to change this by `jQuery-Tags-Input` to make it work. Is that the case for you as well?

mauricew: Yeah, probably because my homestead instance was on Windows.

resources/assets/js/app.js
mauricew: Ok I wasn't aware that Vue was actually being used (since it's included in the default template) but I'll go ahead and make the changes

mauricew: I see they were added with the API

djaiss: Yeah it's especially used for Laravel Passport, and on some screens here and there.

mauricew:
A few things:

* Documentation has a whole section on bower so I don't know how it should be changed
* Dockerfile has been changed
* It seems to be suggested that package/yarn.lock files are check into the repo, I wasn't sure before if you were using npm@5

djaiss:
@mauricew 
* You can omit the documentation if you want - or if you know what we should remove about Bower, do it. I'll update it in case you don't.
* Dockerfile will have to be changed indeed. If you know how, you can do it, otherwise I'll take care of it also.
* I actually don't know what to do with Yarn. Are you still using it in this PR? I don't think you do. Can we safely remove it? I really am not familiar with this whole JS universe, unfortunately, but I'm eager to learn.

mauricew:
I'm using npm 5 but I'm not gonna commit a lockfile. Also I changed the Dockerfile (just removed the two lines having to do with bower)

djaiss:
Thanks so much @mauricew for your help!

PR #608, opened 2017-10-11
Use the config helper instead of putenv

TobiasFroberg:
We've used putenv() to set environment variables in tests forcing us
to reload the environment. We don't have to if we use the config
helper in Laravel instead.


djaiss:
Oh wow. If it works (going to check that soon), that's super awesome because I've found that what I've done was not elegant at all.

djaiss:
It's working fine, you are right. Thanks so much for your help.

PR #615, opened 2017-10-21
Removed usused package martinjoiner/ordinal-number from composer.json

martinjoiner:
Really small change! 

I was working on an unrelated issue and as I ran `composer install` I noticed my own name fly past in the package list. I was excited to see that my Ordinal Number package was required by `composer.json` but upon exploring I found it is never actually used in the code base....  so I removed it to keep things tidy. :-) 

Cheers!


djaiss:
I think I used this package in an earlier version of Monica, last year, and for whatever reason I forgot, I didn't use it anymore. Thanks anyway so much @martinjoiner

Issue #346, opened 2017-06-18
Add new activity categories


djaiss:
Feedback:

```
* Went for drinks
* Went for coffee 
* Played a game together
```

Nebucatnetzer:
Would it be possible to let the user edit this list?

BranTurner:
I think it would be a good idea to make these user editable. Unless the list is incredibly long, there will always be something someone wants to add that does not apply to 99% of the userbase.

djaiss:
I agree - a good idea for v2  😀

Phrohdoh:
Let the user put text into a free-form text field and auto-suggest categories they have previously entered.

pqhf5kd:
This has been completed correct?

djaiss:
@pqhf5kd indeed 😀

PR #616, opened 2017-10-21
Add/Delete activity types

mheap:
This PR allows users to create their own activity groups + activity types and is related to #346.

Instead of creating a whole UI for something that is likely to be infrequently used, I've implemented the functionality as `artisan` commands. There are instructions in `readme.md` how to use it under `Advanced Usage`, and I've recreated here to show how the commands can be used:

### Activities

You can change the activities shown in your Monica instance by creating your own
set of activity groups and activity types. To do this:

First, empty the existing actitivity list (*this will remove all activities from
your contacts. BE CAREFUL*):

```bash
php artisan monica:clear-activities
```

Next, register a new activity group:

```bash
php artisan monica:add-activity-group "My New Group"
```

Finally, add an activity type within that group:

```bash
php artisan monica:add-activity "Did Something" "location_type" "icon" "My New Group"
```

There are limited options available for `location_type` and `icon` due to existing design constraints:

`location_type`:
* `his_place`
* `my_place`
* `outside`

`icon`:
* `ate_his_place`
* `ate_home`
* `bar`
* `concert`
* `hang_out`
* `movie_home`
* `museum`
* `picknicked`
* `play`
* `restaurant`
* `sport`
* `talk_home`
* `theater`

The last thing to do is add translations for your new activities. The way
to do this is create a `resources/lang/<lang>/people.php` file, and add
some new keys for your new activities. They will be prefixed with either
`people.activity_type_group_` or `people.activity_type_`. *These will be deleted
each time you update Monica*

### Checklist

### DIFF
--- a/app/Console/Commands/ActivitiesClear.php
+++ b/app/Console/Commands/ActivitiesClear.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Support\Facades\DB;
+
+class ActivitiesClear extends MonicaCommand
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'monica:clear-activities {--confirm}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Remove all existing activities and activity groups from your installation';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // If they added the confirm flag, we assume they know what they're doing
+        if (! $this->option('confirm')) {
+            // Do we really want to do this?
+            if (! $this->confirm('This will delete all contact activities and activity types. Are you sure you want to do this?')) {
+                return;
+            }
+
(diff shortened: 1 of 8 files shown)


djaiss:
Thanks @mheap 

I won't merge this immediately because I'm finishing the API and I don't want to add another layer in complexity for shipping the API. I'll be here right after though! Thanks!

djaiss:
I'm sorry for all the time it takes me to review this and add it to the repository. First of all thanks so much for all this hard work.

I actually wonder if we should do this in the command line, or if I should add it inside the UI entirely. To be honest I'm considering the second route: letting users change the activity types (or add them) at their leisure. I'll actually do this in a future version, and when I'll do it, this code will be invalid.
What do you think about this?

mheap:
No problem!

I agree that it'd be best with a real UI but creating commands was the quickest solution for me on my local install. Feel free to merge and remove later, or close and implement the UI instead :)

djaiss:
@mheap I can finally close this PR 😀We now have the feature backed in Monica itself, let me know if you find it useful.

And thanks again for all your work there.

PR #617, opened 2017-10-21
Add missing significant other translations for german

dbanck:
Just a small PR for some missing translations I've noticed. I might do another one if I find more 😉


djaiss:

PR #632, opened 2017-11-02
Update minimum PHP version required

anatolinicolae:
Most of the packages already require PHP 7.1, there's no reason to keep main app at 7.0 since it won't install.


djaiss:

PR #657, opened 2017-11-25
Add support for multiple contact fields and addresses

### LINKED ISSUE #128: Allow multiple contact fields for an individual
Currently the software only allows for a single entry for the contact info fields: phone, email, address, etc.  It would be great if you could enter multiple.  This would also help support #125

djaiss:
This PR closes #128 

* It allows multiple contact fields and addresses for a contact.
* A contact field has now a mandatory contact field type (like `email`, `phone number`, `telegram`, ...). These contact field types can be configured per account on the Settings page.
* We leverage Vue.js much more in this release, that allows inline editing for the contact fields and addresses. This will let us have a better user experience.
* The PR requires to run a migration `php artisan migrate` as we add many new tables and move data a bit to populate all the data.


degan6:
This is great work and an awesome feature.

djaiss:
@degan6 thanks :-D

PR #660, opened 2017-11-26
Update i18n contactfields

djaiss:


zach2825:
Instead of copying the untranslated words to the other lang files just utilize the Fallback Language setting
[Laravel localization](https://laravel.com/docs/5.0/localization) 

Sorry if I misunderstood the objective of this PR

PR #688, opened 2017-12-12
Fix contact creation

### LINKED ISSUE #384: Cannot register new user when using Postgres
I tried to set up an instance of Monica with Postgres today and unfortunately ran into the following error any time I try to register a new user (including myself as the first user).

SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "account_id" violates not-null constraint

I am able to register a user with a MySQL (MariaDB) database without issue.

simhnna:
Fix #384


djaiss:
Thanks a lot for this fix. You've actually put the file in the right order. Funny how mySQL didn't behave like Postgre does in this case.

simhnna:
I have a suspicion, that in MySQL such enforcements aren't checked which could theoretically result in a database that is not valid in Monica's view. 

My observations that led to that are this merge request, which fixes a problem which would be fine if the account_id is allowed to be null. The other thing I observed, is that a migration that drops an index is considered invalid by PostgreSQL (it's a migration that was added in the second half of this year, I'm currently mobile so I can't tell you which one exactly it is...)

Someone that knows more about the internals should write a test or two to confirm or deny this...

PR #689, opened 2017-12-12
Run tests using postgres as well

simhnna:
I'm not sure if this will actually build for postgres, but I guess it's a start. If it doesn't and someone knows how to get it running please take over. I'm not a laravel dev, so I really have no idea...

I would really like it if monica wouldn't only be usable with mysql but also with postgres and also sqlite. I don't see a reason why not, CI shoud be able to take care of incompatibilites and alert about them...


djaiss:
Thanks a lot for your contribution! It's definitely a first good step to have this check in Travis.

I can't review this at this time but I will in the coming days. Thanks for your patience!

djaiss:
@simhnna the build fails on Travis for this. I have no clue about Postgre unfortunately.

simhnna:
Now the infrastructure for postgres is well configured. The current errors are from migrations that are only valid for mysql. These need to be changed so they are also valid for postgresql.

The current errors are issues that are already created in the bugtracker:
* [Postgresql doesn't have the keyword `MODIFY`, instead it uses `ALTER`](https://stackoverflow.com/questions/36262357/error-syntax-error-at-or-near-modify-in-postgres)

I'm not sure why raw sql statements are used in the first place, but then again I don't know enough about the internals of monica.

The mysql build fails because of something in the `travis_sonar.sh` but it provides absolutley no output, so there is no way for me to know what the issue is.

djaiss:
@asbiin do you know why `travis_sonar.sh` raises an error here?

.env.postgres.travis
```diff
@@ -0,0 +1,14 @@
+APP_ENV=testing
+APP_KEY=base64:NTrXToqFZJlv48dgPc+kNpc3SBt333TfDnF1mDShsBg=
+
+DB_CONNECTION=pgsql
```
asbiin: You should propose a new "pgsqltesting" connection configuration instead, using DB_TEST_* variables, like the "testing" configuration in config/database.php

simhnna: @asbiin you are welcome to take over. I'm not using monica anymore (partly because of the database issues)

PR #696, opened 2017-12-14
New feature : Password change in settings screen

asbiin:
Laravel has password **reset** functionality. This works, but we want a user to have the ability of **change** his password directly in the UI, without having to reset it.

This PR changes the vue of settings screen by adding a from with 3 inputs : current password (for security), and 2 inputs for a new password.
image

The main code is in PasswordChangeController class, wich is fully inspired by the ResetsPasswords trait.
Same rules for password check apply as Laravel, with the Password::broker()->validateNewPassword($credentials) code (ie. 6 characters length).


djaiss:
@asbiin 
Thanks a lot for this pull request!

Apparently style CI is complaining about your code. Could you check it out?

Also, could you add the English keys to the other translation files, so people will be able to translate them?

Thanks a lot!

djaiss:
@asbiin I keep having this error

image

asbiin:
@djaiss  This is odd : the "password current field is required" error has been fixed by 1dae0f9 commit (Fix input name), which is not marqued as "checked"

djaiss:
Yes it works :-D Awesome.

djaiss:
@asbiin want me to merge it or are there other stuff to do in the PR?

asbiin:
A good thing to add would be to check if the new password and the confirmation one matches, client side, with some js. But I don't know the best way to do this in Monica.
Apart from that, you can merge it, thanks !

Issue #30, opened 2017-06-06
Nicely done, but please ease off the setup procedure


bensbox:
Hi!

This is a really nice project - i was totally excited and eager to test it out, until ....i read the setup procedures...
I am a professional php developer for years. I am able to create complex network and server setups inside and outside of the "cloud" and i am using most of your mentioned tools on a daily basis for quite some time - i am one of those people, who are able to setup this project without a problem.

I just don't see why small and promising projects like this one need to choke themselves on compiling css, having nodejs installed, setting up virtual machines or (soon(tm) ..pls not) providing docker stuff.

How about just simple php and css? Lamp stack anyone? Please make it fancy and please go ahead and use composer - maybe with a tarball, where all the composer files are already provided.
Just think about all the people out there with a limited php hosting package or a toyserver/raspberry at home with a spare-time-understanding of setting up stuff like wordpress. They will be eager to use it - that's why wordpress is so successful.

Thanks alot and keep up the good work!

ocdtrekkie:
I can definitely agree here that this is something to work on. The setup instructions assume you're using Homestead. But what if I'm not? What does the setup require from Homestead to install without it?

bensbox:
@ocdtrekkie : You don't need Homestead - it will work with a standard LAMP stack, but you need to tinker a bit.

What i did:
1. As requirements, you need the following php packages: php-mbstring, php-dom, php-intl, php-mysql
2. Follow the instructions until the first "php artisan" command. The command will fail, because the APP_KEY in the .env file is too short. Just replace it with some 32 char long key. After that execute "php artisan key:generate" and it will replace your key with a random base64 encoded string. After that you can continue with the setup.
3. After the setup, you need to add a .htacess file into the public folder and enable the "rewrite" module of apache2 (i just suggest you use the lamp stack :) ). The content is provided in the "web.config" file in the public folder - i just use the following:
`RewriteEngine on

RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d

RewriteRule ^(.*)$ index.php [L,QSA]`

Thats it :)

djaiss:
You don't need Homestead if you don't want to. It's just simpler.

Laravel requires Nodes by default to compile SCSS for instance. I think Laravel is one of the simplest framework out there. Composer is massively used actually here.

I do understand now though the importance of making this simpler to install. Sorry about that.

underyx:
[...]

PR #711, opened 2017-12-18
Added vagrant installation instructions as per issue #30

dp87:
I have added a single documentation file to explain how to create a VM running Monica using Vagrant, to have an even quicker installation method, using the Vagrant file and provisioning script I have created.


djaiss:
@dp87 Thanks a lot for your help!

PR #712, opened 2017-12-18
fix volume permissions on startup #710

### LINKED ISSUE #710: Exporting data in local install does not work
I've created a personal instance of Monica running on a Debian server with Docker. Before I get too committed to using it, I want to be certain that I can export and backup my data. I found the Export data section under settings, but when I try it I get an error saying "Whoops, looks like something went wrong."

If memory serves from the last time I tried Monica, I think this is being caused because the function is intended for getting data out of the hosted version of Monica for import to a personal version. I think export need to work no matter where Monica is installed. 

Ultimately, I am concerned about backing up data at this point. If there is a CLI way I can do this outside of Monica, I am interested.

mgolisch:
makes sure the public storage directory is writeable by the webserver user.
Fixes #710 

Not sure if people might want to mount other volumes into the docker container.
Maybe applying the permissions on /var/www/monica/storage would be a better idea?


mgolisch:
@kstrauser  whats your opinion on that. you seem to be the one that did the initial dockerisation.

kstrauser:
I'm not sure. The problem seems to happen randomly after a few days of uptime - like maybe a cron job is running `chown -R ...` after a log rotation - but I haven't seen it since upgrading to 1.4.1. Perhaps that's no longer an issue?

djaiss:
@mgolisch thanks for this PR! I've been busy with another feature - hence my silence.
I'm not familiar enough with Docker to know if this change breaks something.
I'm going to trust you both on that topic and merge it 😀

PR #714, opened 2017-12-20
Display external avatars when searching

### LINKED ISSUE #708: Search doesn't display external avatars
When searching for a contact, if the contact has an external avatar, we display the letters instead.

See below

image

BranTurner:
Shows external avatars in the search window - fixes #708


djaiss:
@BranTurner yet another excellent job. Thanks a lot!

PR #720, opened 2017-12-26
fix #664 allow setting of db prefix via env #664

### LINKED ISSUE #664: Allow setting of db_prefix via env
I want to save limited mysql databases on my hosting account and use monica together with other apps in on db.
I could use the db-prefix setting, but looking at database.php, it is not possible to configure the default '' string.

turtles2:
This allows db_preflix to be set by .env.


djaiss:

PR #728, opened 2017-12-28
Changed git clone URL to be publicly accessible

hexnet:
Change to git clone URL as the git@ will not work without a granted private key.


djaiss:

Issue #679, opened 2017-12-08
Adding a note doesn't work.


Nebucatnetzer:
When I try to add a note to a contact with the new form I get this error from the JavaScript console.
```
Blocked loading mixed active content "http://monica.2li.ch/people/71/notes"
```
I'm running Monica behind an Apache Reverse Proxy.

djaiss:
@Nebucatnetzer looks like an error from http to https or vice-versa.
Have you configured the Url properly in your `env` file?

Nebucatnetzer:
I think so:
```
APP_URL=https://monica.2li.ch
```

Nebucatnetzer:
@djaiss I'm not really familiar with Laravel but are we maybe missing something like this?
https://laravel.com/docs/5.5/requests#configuring-trusted-proxies

Or is this something that can get added individually?

djaiss:
@Nebucatnetzer unfortunately I don't know the impact this would have if I added this to the codebase. It'd be awesome if the community could help here, because I don't have the answer to your question 😢

Nebucatnetzer:
Just for reference :)
I searched around a bit today but wasn't really able to find much related information.
Adding the TrustedProxies file on its own didn't change anything. I reckon one needs to reference/include it somewhere.

As for information related to Vue.js behind a reverse proxy I basically found nothing at all.

tristankerner:
As far as I know, the "blocked loading mixed active content" is a client/browser based error caused by loading/accessing an HTTP resource on an HTTPS page. I'm not familiar with Apache Reverse Proxy, but I believe the theory is that all client traffic should be routed through the reverse proxy, and be HTTPS from the client perspective. On the servers side, traffic may be routed around in other ways, but that should be invisible to the client and it shouldn't cause an error like this. 

Any form using Vue.js/Axios/AJAX is sending a request from the client's browser, to the server, and that request should be taking place via an HTTPS endpoint. If this request were successfully done via the attempted "http://monica.2li.ch/people/71/notes" URL, the content of that form submission would be unencrypted).

The goal should probably be to identify why a plain HTTP request is being attempted from the client side, and change it to HTTPS. My initial thought is that there may be a configuration error, either in the reverse proxy, or the server software that is causing the URL to be redirected to plain HTTP for the client, instead of going through HTTPS like it should.

Just to confirm, you received this error when viewing the page in HTTPS (i.e. https://monica.2li.ch/people/71/ )?

djaiss:
[...]

PR #730, opened 2017-12-30
Fix the URL::forceSchema to forceScheme

mwalling:
Based on the [Laravel documentation][1], it looks like this is a typo. Even in production mode with `APP_URL` set correctly, `route` was returning `http://` URLs in the templates.

[1]: https://laravel.com/api/5.5/Illuminate/Routing/UrlGenerator.html#method_forceScheme

Fixes https://github.com/monicahq/monica/issues/679


djaiss:
Thanks so much for this!

PR #731, opened 2017-12-30
Fix an issue where February shows up as March in the date picker

mwalling:
When the current day of the month is greater than the number of days in
the month, the date class was rolling forward to March.

Before the fix:
screen shot 2017-12-29 at 8 17 38 pm

And after:
screen shot 2017-12-29 at 8 18 13 pm


djaiss:
Woot 🎆

PR #735, opened 2018-01-02
Fixes #729 Allow searching by phone number (of contact fields in general)

### LINKED ISSUE #729: Allow searching by phone number (of contact fields in general)
Scenario: somebody called, but I didn't answer. Now want to find who it was, but... (s)he called on new phone without contact list or my sibling's phone or I just don't have all people's numbers saved on the phone.

Ideally, there should be option to specify type of contact field similarly to how it's possible in GMail or Github - eg. `mail:somebody@somewhere.com`, `phone:555123456`, but with ability to define custom contact field types it might be hard. It will probably require to specify alphanumeric ID when creating field.

This feature could be hidden behind keyword for advanced users (maybe some generic `contact:xxx` or `contact-field:xxx`?) to not pollute standard search results that use first/middle/last name. Or maybe show results at the end of the list - first the results found using existing mechanism (matching first/middle/last name) and then the results found using this feature.

turtles2:
This will allow a user to search there contacts by there contact information. It is done by entering the field they are searching by followed by a : then the search term. For example to find a contact with the email example@example.com you would enter email:example@example.com. 

Fixes #729


djaiss:
This is awesome.

Issue #659, opened 2017-11-25
Feat Request: Ability to filter by more than one tag


chrisfosterelli:
It'd be great to be able to filter the contact list by more than one tag at a time. Say a user is going through a nearby city and wants to find a friend to catch up with while there, they could achieve that by being able to filter by both the `<nearby town>` tag and by the `social` tag.

chrisfosterelli:
Thanks @djaiss and @turtles2! 💯

PR #738, opened 2018-01-02
Add #659 Ability to filter by more than one tag

turtles2:
This will add #659 Ability to filter by more than one tag. It works the same as the current feature only it allows unlimited use of tags. It filters exclusively so the displayed contacts need to have all the selected tags.


djaiss:
It works perfectly! Thanks a lot for your work!

PR #741, opened 2018-01-02
Fixes #276

### LINKED ISSUE #276: Editing a contact which doesn't exists throws an exception
```
Sorry, the page you are looking for could not be found.
(2/2) NotFoundHttpException

No query results for model [App\Contact] 24
```

Steps to reproduce:

1. Open a contact twice
2. Delete it in the first tab
3. Move to the second tab and click on "Edit contact information"
4. Monica throws an exception

This should probably return a message like:

> The contact you are trying to edit doesn't exist anymore.

Otherwise it might be a bit confusing for the users.

Edit:
The same happens by the way if you want to open a contact which got deleted.
Maybe a more generic message like

> The contact you're trying to interact with, doesn't exist. Would you like to create it?

Could be used for both cases and then offer a button to create a new contact?

turtles2:
This fixes #276 by sending the user to a page that says No Person Found Add New One Now and then allowing them to make a new contact


turtles2:
This passes travis it just needs to be rebuilt as for some reason travis timed out on composer install/update

turtles2:
Never Minded I added so do nothing commits to force Travis to rebuild

djaiss:
👍 

Super job!

PR #744, opened 2018-01-03
Revert "Fix the URL::forceSchema to forceScheme"

mwalling:
Reverts monicahq/monica#730

What I committed doesn't match the documentation or the title of the PR. I'm not sure how I screwed that up.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,7 +11,7 @@
 */
 
 if (App::environment('production')) {
-    URL::forceSchema('https');
+    URL::forceScheme('https');
 }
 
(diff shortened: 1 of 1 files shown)


mwalling:
This makes more sense now. I was running based on the Docker image pushed to https://hub.docker.com/r/monicahq/monicahq/ and didn't realize that the `.env.example` was packaged. The default for the App environment was production, but the `.env` file overrode that to `local`, which is why I never hit it in testing.

(The last time I wrote PHP was about a decade ago. I'm a bit rusty with these new things like DotEnv)

djaiss:
My bad too :-D Thanks for this.

Issue #609, opened 2017-10-12
Manage Tags in Contacts View


jwynn6:
Add the ability to manage tags inline, in the contacts view.  At the very least show the tags that are associated with each contact. It is all well and good to be able to sort by tags but it isnt helpful when several users are added via "significant other" or "child" dialog boxes. Those contacts don't get tagged until their contact card is deliberately opened. When doing a  large volume of entries ,t it is easy to forget to tag a contact, and impossible to determine who is not tagged.

gediminaska:
Please check out my pull request https://github.com/monicahq/monica/pull/751

PR #751, opened 2018-01-05
#609 tags in contacts view ready for review

gediminaska:
tags1
tags2
tags3
tags4
tags5


djaiss:
Thanks for your contribution @gediminaska 

Unfortunately, your proposed implementation has two problems:
* the UI does not have the level of quality we expect in Monica
* the UX is not right.

I think we better have a system where we can do bulk actions, like checkboxes on the left column and a dropdown at the top - exactly how GitHub does it with the issues ( for instance https://github.com/monicahq/monica/issues).

PR #772, opened 2018-01-07
Remove trailing slash from notes endpoint

henk23:
With the trailing slash the server sends a "301 Moved Permanently" and redirects from `/notes/` to `/notes` but also changes the request method from POST to GET, resulting in the note not being stored.
The trailing slash handling is probably configurable somewhere? But this fixes the issue.


djaiss:
Thanks for your help!

PR #781, opened 2018-01-09
Use strict mode for mysql

simhnna:
Use strict mode in mysql. I suspect a couple of errors that are reported when using postgresql result from the fact that testing doesn't enforce strict rules in the first place.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -62,7 +62,7 @@ $db = [
             'charset' => 'utf8',
             'collation' => 'utf8_unicode_ci',
             'prefix' => env('DB_PREFIX', ''),
-            'strict' => false,
+            'strict' => true,
             'engine' => null,
         ],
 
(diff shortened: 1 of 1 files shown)


simhnna:
The tests confirm my suspicions... 
It's broken for mysql as well, it's just that nobody notices anything...

While this branch makes the tests fail it should still be merged because it shows real errors!
Feel free to adopt the branch and fixing the issues...

simhnna:
The current error is because of a bogus migration in 
`database/migrations/2016_12_28_150831_change_title_column.php`
```
16:            $table->string('title', 60000)->change();
```

divinity76:
should be developing with sql_mode=TRADITIONAL (and that's pretty much how postgres work by default) - when the code works in TRADITIONAL mode, it should work anywhere, and you'll catch bugs much sooner than running with default mysql settings. (where default mysql will truncate your data, or write `0`, or otherwise __SILENTLY__ corrupt your data to make it fit with the schema, TRADITIONAL will throw errors... it's hard to think of scenarios where default mysql mode is actually a good idea)

i3anaan:
Very much agree with the above, hope this gets picked up. Might actually look more into this myself...

PR #801, opened 2018-01-13
Use tag1 instead of tags when filtering contacts.

tomzx:
Fix a regression in contact/people filtering based on tags.
Introduced by b1708801e425ffe98d0e65e14601bedcadad4d45.
Relates to #659.


turtles2:
@tomzx Good Catch. I did not think of the impact said change had on that page.

djaiss:
Thanks @tomzx for this PR.

Issue #725, opened 2017-12-27
Add statistics about the moods


djaiss:
Since a few days we allow the user to indicate how the day went.

We could provide statistics in the Journal to indicate, per month and year, the number of times the day was great, not great or bad.

PR #823, opened 2018-01-17
Add initial journal mood trend graph

stevenmaguire:
Included in this changeset is an initial pass at a solution to #725. It includes:

- new journal controller method that returns aggregated day ratings all days in previous 12 months
- a new, optional day rating database seeder for testing
- a new vuejs component to manage the graph, including it's consumption of two new libraries

I am sure the linter will yell at me for a few of the changes. After I clean those up, I am open to code style or feedback as the group sees fit. 

screen shot 2018-01-20 at 12 26 43 pm


degan6:
This looks awesome but it looks like highcharts.js is a commercial product.
https://shop.highsoft.com/highcharts

djaiss:
@degan6 damn. You are right. We can't use it :-(

stevenmaguire:
Oh! Right, sorry. I will find a new vendor. Any preferences?

stevenmaguire:
Updated to use ChartJS. Sizing and styling is a bit trickier with this one.

djaiss:
@stevenmaguire don't hesitate to ping me when you make update you want me to check because otherwise I don't think of checking PR regularly.

stevenmaguire:
@djaiss I appreciate that. This is one is definitely ready for some feedback :)

app/Http/Controllers/JournalController.php
```diff
@@ -188,4 +192,30 @@ public function deleteEntry(Request $request, $entryId)
...
+            'range' => 12,
+            'data' => [],
+            'available_units' => ['days', 'weeks', 'months', 'years'],
+        ];
+        $start = Carbon::now()->subMonths($results['range']);
+        $dateIntervalValue = 'P1D';
+        $period = new DatePeriod($start, new DateInterval($dateIntervalValue), Carbon::now());
+        foreach ($period as $date) {
+            $dateString = $date->format('Y-m-d');
+            $rating = DB::table('days')->whereDate('date', $dateString)->pluck('rate')->first();
```
djaiss: 1. Why use `DB::table('days')` and not Eloquent?
2. You don't filter by `account_id`, which means you will take all the days for all the accounts here, which is bad 😀

database/seeds/FakeDayRatingTableSeeder.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+use App\Day;
+use Carbon\Carbon;
+use App\JournalEntry;
+use Illuminate\Database\Seeder;
+
+class FakeDayRatingTableSeeder extends Seeder
```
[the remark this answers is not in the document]
stevenmaguire: Ah! I didn't see that seeder; maybe I didn't look hard enough. Will remove this one. Thanks!

app/Http/Controllers/JournalController.php
stevenmaguire: Good catch. Will do.

Issue #328, opened 2017-06-16
Gifts received


alpedhuez:
Enjoying playing around with this website. 

It would be good to have a way to list gifts received from a contact as well as offered

(Linked to #269 ?)

djaiss:
The ability to mark a gift as offered is now live.

Issue #269, opened 2017-06-14
Mark a gift as given/offered


sloria:
There is currently no way to mark a gift as "given."

dpinezich:
I would love to implement this, but there are some questions regarding the current datastructure:

If we want to transform a giftidea to a gift already offered it is straightforward (in my understanding) add a button to action, get a date when it was offered, change the according fields in the gifts table. 

But if I look at the current "gift already offered" datastructure we have one problem to solve. We should also offer the possibility to insert a date when the gift was offered (by default it is just the creation_date to avoid null entries)?

How are your intentions into this?

djaiss:
The ability to mark a gift as offered is now live.

Issue #155, opened 2017-06-09
Edit Gift


adrum:
It would be nice to be able to edit a particular gift entry. I envision this being nice for brainstorming for a gift you know will need to purchase, like a nephew's or niece's birthday. I could then create this placeholder gift idea, and then come back to this placeholder gift and change the name or whatever.

djaiss:
This is now in production.

Issue #122, opened 2017-06-07
Allow tracking Gifts received from a Contact


andruby:
It's possible to add gift ideas for and gifts offered to a Contact, but no way to track the gifts you received from a contact. It would be great to add that option.

djaiss:
This is now in production.

PR #831, opened 2018-01-19
Add ability to mark a gift idea as being offered

djaiss:


erdmenchen:
So this adresses #269, #328, #155, #122

djaiss:
@erdmenchen excellent catch! Thanks for keeping track!

PR #871, opened 2018-01-31
Fix Gitter link and read badge for sonar

asbiin:
image


djaiss:
@asbiin you can merge!

Issue #800, opened 2018-01-13
Remove compiled assets from repository


tomzx:
Hello,

Is there a reason that compiled assets (`public/css/app.css`, `public/js/app.js`) as well as some fetched assets (`public/fonts/vendor/font-awesome/fontawesome-webfont.svg`, 
`public/fonts/vendor/octicons/build/octicons.svg`) part of the repository? As they are either compiled or fetched when you install/update monica, there is no reason those should be part of the repository and they should actually be included in the `.gitignore`.

Thoughts?

djaiss:
@tomzx technically they are not necessarily compiled if you install Monica, both locally or on a server. They just are if you run `npm run dev` or `npm run prod`. If we remove the compiled assets, we would force people to install node etc... while at the moment, they don't have to if they don't want to.

I think it's simpler that way. Moreover, it's also simpler for me so I don't need to compile them on the server 😀 But I'm open to the debate here.

tomzx:
The issue is that the git repository now contains those files and they are large, ~3MB for `app.css` and ~5MB for `app.js`. Now, I'm not complaining about the size in itself, as text is highly compressible, but it's generally a best practice not to commit anything that is generated.

If we want to make it easy for people to run it without requiring `npm`/`yarn`, then a better alternative would be to prepare a ready-to-use archive.

Finally, it begs the question "when should these files be commited?" Should they be only updated by the person preparing a release, or by anyone that modifies a file that will end up in `app.css`/`app.js`, or are they just updated "when someone feels like it"?

djaiss:
I agree that the current state of the repository is not clean. I hate having generated files inside git.

I think we should move forward with removing them from the repository and I would have to compile them on the server after each pull. I just need to figure out how 😀

tomzx:
Given that you use laravel, you could consider using [envoy](https://laravel.com/docs/5.5/envoy) to script your deployment. Or you could write a php script or a node.js script and call that after `git pull`.

djaiss:
@tomzx oh that's nice! I thought it was only available to the Envoyer subscribers!
Thanks for the reco, I'll check it out.

asbiin:
This is a duplicate of #136

PR #884, opened 2018-02-05
Build assets on travis and deploy them on bintray

asbiin:
Linked to issue #800.
Assets will be build and then deploy on bintray, only for tags versions.


docs/contribute/contribute.md
```diff
@@ -76,17 +76,19 @@ We've installed the development version with [Valet](https://laravel.com/docs/va
 
 Once the above softwares are installed (or if you've finished the installation of Homestead/Valet):
 
-1. `composer install` in the folder the repository has been cloned.
+1. `make build-dev` in the folder the repository has been cloned. This will run :
+    - `composer install` to install all packages.
+    - `npm install` to install all the front-end dependencies and tools needed to compile assets.
+    - `npm run dev` to compile js and css assets.
 1. `cp .env.example .env` to create your own version of all the environment variables needed for the project to work.
```
djaiss: It would be good to do `cp .env.example.env` and `php artisan key:generate` in the makefile as well.

asbiin: There is a `make install` command for that ;) Except for the `cp .env.example .env` command. I could add it if you insist :)

PR #898, opened 2018-02-14
Fix genders setup production

asbiin:
From a fresh installation, genders table are not set for the first register account/user.


asbiin:
[test]
capture
[id:AzCqkY2T46VLvoh]

asbiin:
@djaiss  thanks ! Except tests are KO 🤣

Issue #817, opened 2018-01-15
version.monica.com stil reporting 1.6.0?


tlwt:
I am running version 1.6.1 (in docker).

It displays that there is a newer version and suggests 1.6.0 (see screen shot below). I assume that version.monica.com is not updated yet to 1.6.1. 

Another fix may be to change the following line so it does not just check for NOT EQUAL but for LARGER.

https://github.com/monicahq/monica/blob/647e8fa4f8ef2fb9242f4ca0834618c26c1da975/resources/views/partials/check.blade.php#L5 

![monica](https://user-images.githubusercontent.com/8267086/34944749-1058a922-fa01-11e7-9b55-76297bff4dc4.png)

djaiss:
As a matter of fact, version.monicahq.com is updated to 1.6.1.
Can you run `php artisan monica:ping` and see if it fixes the problem?

tlwt:
yes, it does fix the problem.

djaiss:
Do you know if the cron that should do it every day is setup?

tlwt:
Would this be run as root? I did not find anything there.

```
~ # whoami
root
~ # crontab -l
# do daily/weekly/monthly maintenance
# min   hour    day     month   weekday command
*/15    *       *       *       *       run-parts /etc/periodic/15min
0       *       *       *       *       run-parts /etc/periodic/hourly
0       2       *       *       *       run-parts /etc/periodic/daily
0       3       *       *       6       run-parts /etc/periodic/weekly
0       5       1       *       *       run-parts /etc/periodic/monthly

~ # ls -Rt /etc/periodic/
/etc/periodic/:
15min    daily    hourly   monthly  weekly

/etc/periodic/15min:

/etc/periodic/daily:

/etc/periodic/hourly:

/etc/periodic/monthly:

/etc/periodic/weekly:
~ # 
```

djaiss:
I think you have to set it manually.

tlwt:
Even in the dockerized setup?

djaiss:
Do you still encounter the problem?

tlwt:
No - looks good to me

djaiss:
Thank you!

PR #906, opened 2018-02-17
Docker run real cron

### LINKED ISSUE #875: Docker image: Run real cron tasks
In docker/test-server.php:
```
while true; do
    sleep 60
    ${ARTISAN} schedule:run
done
```
We should try to run cron tasks otherwise.

asbiin:
Fix #875

Running cron to run `php artisan schedule:run`.
Fix some other docker conf.


asbiin:
This should close #817 as well.

PR #911, opened 2018-02-18
Tests with sqlite [wip]

asbiin:
Run tests with sqlite.
Actually tests are failling ...


asbiin:
This is not working. Sqlite support must be review from the beginning.

PR #912, opened 2018-02-18
Fix gender add button

asbiin:
In these GET/POST request, the trailing slash makes the http server to respond with a 301 redirect.
This is because of this line in .htaccess: `RewriteRule ^(.*)/$ /$1 [L,R=301]`

docker image is not impacted because it has it own Redirect rules (for now).
app.monicahq.com is protected behind cloudflare, so maybe it's why it's working ?
heroku uses nginx, it might not be impacted as well ?
Other apps should be impacted ....


asbiin:
Should we fix this too ?
[routes.web.txt](https://github.com/monicahq/monica/files/1735850/routes.web.txt)

djaiss:
@asbiin excellent catch. We should probably fix the routes/web.php file as well indeed, but this could be done in a separate PR.

PR #950, opened 2018-02-28
Sentry config file: test if .git directory is present

### LINKED ISSUE #872: Ugly config
in config/sentry.php:
```
...
    // capture release as git sha
    'release' => trim(exec('git log --pretty="%h" -n1 HEAD')),
...
```
This is ugly. We don't want to publish .git directory in production, but this line is executed each time "php artisan" is called.
Is this still needed ?

asbiin:
This fixes #872.
If you run `php artisan config:cache` the command will be run once.

### DIFF
--- a/config/sentry.php
+++ b/config/sentry.php
@@ -4,7 +4,7 @@ return [
     'dsn' => env('SENTRY_DSN'),
 
     // capture release as git sha
-    'release' => trim(exec('git log --pretty="%h" -n1 HEAD')),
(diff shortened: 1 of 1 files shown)


asbiin:
@djaiss do you agree with this ?

asbiin:
[test]
capture
tests/Browser/screenshots/failure-test_it_displays_reminder_rules-0.png
[id:MRtGHJB6FcGJ7Ye]

djaiss:
Sentry is really specific to our hosted version, so it won't affect other instances.

That means we can assume that `php artisan config:cache` is indeed actually run, and therefore it will indeed be processed only once (actually each time we deploy a new version).

I think it's ok to do this.

PR #952, opened 2018-03-01
Optimizations for composer

asbiin:
Optimization for composer:
- Add apcu optimization
- Remove prefer-dist and optimize-autoloader from composer commands and documentation as it's already included in the composer.json file
- update composer.lock


asbiin:
[test]
capture
[id:AKBUhdvJvJ75WIp]

asbiin:
@djaiss I had to add symfony/translation dependency: was "v3.4.4", now "v3.4.5", because v4.0 is not compatbile with php 7.0 - The component is needed by jenssegers/date and nesbot/carbon.
Do we still want to support php 7.0 ?

djaiss:
@asbiin yes, we need to support exactly what Laravel supports. And they support 7.0.

PR #983, opened 2018-03-05
Require debugbar on dev only

barryvdh:
Remove debugbar from production, suggested to only install it on dev environments. Especially for personal information like this. (Eg. people who accidentally forget to disable DEBUG).

Also bumped to 3.x. ServiceProvider is no longer required because of the auto detecting of packages.


asbiin:
That's a good idea ! Thank you @barryvdh

PR #1041, opened 2018-03-23
Fixed usage of wrong success message when adding existing contact as …

DarkGigaByte:
…kid of someone


asbiin:
@DarkGigaByte thank you !

PR #1043, opened 2018-03-24
Fix called_at field in the Call object returned by the API

djaiss:


djaiss:
This will address https://github.com/monicahq/chandler/issues/129

PR #1064, opened 2018-03-27
New Crowdin translations

djaiss:
Enjoy!

PR #1065, opened 2018-03-27
German and French localization updates

jkbecker:
Found some gaps and errors in the German localization.

Sorry; I only realized the existence of Crowdin after the fact! Is this still mergeable?


jkbecker:
Should be merged after #1064

djaiss:
Thanks so much for your help! But... we can't merge it at all, since we use Crowdin now. If we do merge it, we'll have a lot of conflicts and we'll be out of sync :-(

Really sorry about this.

PR #1066, opened 2018-03-27
adding reference to Crowdin

jkbecker:
Hi @djaiss , following up on #1065 -- could we add this to the CONTRIBUTING page? I only knew about Crowdin after the fact, which is unfortunate. I'm sure this would make it clearer for future localization contributors.
What do you think?


djaiss:
Thanks so much for your help!

PR #1068, opened 2018-03-28
Use lang list instead of auto for lang-detector

asbiin:


asbiin:
Linked to https://github.com/vluzrmos/laravel-language-detector/issues/27

PR #1069, opened 2018-03-29
Fix i18n support

asbiin:
- Fix i18n quotation
- Add hreflang on links
- Fix lang list on settings which does not appear ! Bug introduced with #1068


asbiin:
[test]
capture
[id:mMjaaFvxXbEzfBl]

PR #1073, opened 2018-03-30
Order lastname and middlename parameters correctly for contact->setName call

HHErebus:
Hello hello,

I bumped into an issue on master when editing a contact, the order of parameters in ContactsController:243 (firstName, middleName, lastName) is different than in the function being called at Contact:729 (firstName, lastName, middleName = null), which results in an exception.

I have not added a test since the whole PR is literally swapping two parameters; I can add it if required, of course.

#### Before submitting the PR
- [ ] If you change the UI, make sure the user experience is consistent with the current interface.

#### Code-related tasks
- [ ] Impact on the seeders.
- [ ] Impact on the API.

- [ ] Impact on account export.
- [ ] Impact on importing data with `vCard` and `.csv` files.
- [ ] Impact on account reset and deletion.

#### Other tasks
- [ ] If it's relevant, add the documentation about your feature in the README file.
- [ ] Indicate `[wip]` in the title of the PR it is is not final yet. Remove `[wip]` when ready. Otherwise the PR will be considered complete and rejected if it's not working.

### DIFF
--- a/app/Http/Controllers/ContactsController.php
+++ b/app/Http/Controllers/ContactsController.php
@@ -240,7 +240,7 @@ class ContactsController extends Controller
                 ->withErrors($validator);
         }
 
-        if (! $contact->setName($request->input('firstname'), $request->input('lastname'), null)) {
+        if (! $contact->setName($request->input('firstname'), $request->input('lastname'))) {
             return back()
                 ->withInput()
                 ->withErrors('There has been a problem with saving the name.');
--- a/app/Http/Controllers/ContactsController.php
+++ b/app/Http/Controllers/ContactsController.php
@@ -240,7 +240,7 @@ class ContactsController extends Controller
                 ->withErrors($validator);
         }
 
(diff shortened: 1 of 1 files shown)


HHErebus:
Not quite sure what would be causing the failure of 2205.2, to be honest. I'll look into it during the weekend

app/Http/Controllers/ContactsController.php
```diff
@@ -240,7 +240,7 @@ public function update(Request $request, Contact $contact)
                 ->withErrors($validator);
         }
 
-        if (! $contact->setName($request->input('firstname'), null, $request->input('lastname'))) {
+        if (! $contact->setName($request->input('firstname'), $request->input('lastname'), null)) {
```
asbiin: We should remove the `null` value here, because the third parameter of `setName` is optionnal.

asbiin:
@HHErebus thank you !

PR #1092, opened 2018-04-02
Upgrade to laravel 5.6 [wip]

djaiss:
Warning:
* Laravel now requires `DB_TEST_HOST` in your .env file otherwise you won't be able to launch your tests.
* Monica now only supports PHP 7.1.3+.


djaiss:
@asbiin that's unfortunate.

djaiss:
@asbiin do we need this library?

asbiin:
@djaiss yes we do now, it is used to get code coverage when running browser tests. See https://github.com/monicahq/monica/blob/master/scripts/tests/server-cc.php
I'm doing some research so see what to do with it.

asbiin:
@djaiss I've forked the project and created a PR to update the dependency of matthiasnoback/live-code-coverage
Unfortunately, Laravel 5.6 is not compatible with php 7.0 anymore so we will need to remove travis run with php 7.0

djaiss:
@asbiin indeed, Laravel 5.6 needs PHP 7.1.3. We just follow what they require anyway, as we need to follow the framework.

asbiin:
[test]
capture
tests/Browser/screenshots/failure-Tests_Browser_Settings_MultiFAControllerTest_testBarCodeContent-0.png
[id:5BGnjQLX8ITZ4ul]

config/logging.php
```diff
@@ -0,0 +1,60 @@
...
+    |--------------------------------------------------------------------------
+    | Default Log Channel
+    |--------------------------------------------------------------------------
+    |
+    | This option defines the default log channel that gets used when writing
+    | messages to the logs. The name specified in this option should match
+    | one of the channels defined in the "channels" configuration array.
+    |
+    */
+    'default' => env('LOG_CHANNEL', 'stack'),
```
asbiin: Should not we use `env('LOG_CHANNEL', env('APP_LOG', 'stack'))` here for backward compatibility ?

PR #1103, opened 2018-04-04
Change some messages formulation

### LINKED ISSUE #1102: Activities list say "ate at his place", regardless of the gender of the person involved
Steps to reproduce:

1. Pick a person from https://app.monicahq.com/people. Open their profile.
2. Under "Activities", click "Add an Activity".
3. Look at the "Would you like to categorize this activity?" dropdown.

If I look at this for a contact whose gender is "Woman" (picked from the default list), I see the option _ate at his place_, which is wrong.

screen shot 2018-04-04 at 13 01 50

The person in question uses “she/her” pronouns, so this dropdown should change to match.

Even better would be to use a gender-neutral pronoun here (where the language allows) – consider the alternative _ate at their place_.

---

The code in question seems to have his/her variants, so it seems like a bug that the right one isn’t being selected here:

https://github.com/monicahq/monica/blob/18f63df9e59eb0e4238da1bee1853138c7f91e9f/resources/lang/en/people.php#L186-L187

And seems to be the only gendered phrase on a quick skim.

asbiin:
- Remove every quotes usages `'`, replaced by real apostrophe `’`
- Use en dash when it is usefull
- Use gender-neutral for 1 message – close #1102


asbiin:
[test]
capture
tests/Browser/screenshots/failure-Tests_BrowserFeature_UploadVCardTest_test_upload_vcard_is_accessible_from_add_contact_view-0.png
[id:BYfALrXSxZywhBQ]

asbiin:
[test]
capture
tests/Browser/screenshots/failure-Tests_BrowserFeature_UploadVCardTest_test_upload_vcard_is_accessible_from_add_contact_view-0.png
[id:NfBK6jZ57LQG21M]

Issue #743, opened 2018-01-02
Feature request: Let the user select a date for a journal entry.


Nebucatnetzer:
When a user adds a journal entry Monica uses the current date automatically without an option to select a custom date.
It would be nice to have an option to select a custom date because I often don't have time to write down an entry on the same day the event occurred.

djaiss:
@Nebucatnetzer completely agree on this one.

PR #1116, opened 2018-04-07
Add journal entry date

Dagolin:
Fixes : #743

- Add date entry in journal, save it to `journal_entries.date`
- Move `journalEntries` relationship from `Model App\Day` to `Trait Journalable`
- Do nothing to API and UnitTest since both of them are a little out-of-date and might require more extra works to update which not belongs to this issue.

It seems that there are three `Polymorphic Relations` entries, which are
 - `activities`   (date saved to date_it_happened)
 - `entries(journel use)` (date saved to journal_entries.date in this PR, )
 - `days` (date saved to date)

However,  when the entries are loaded in page like `Journal.index` the `created_at` is used instead of `date_it_happened or any other`, which might be very confusing if user input an activity with different date, I suggest that the date column should be in `pivot` table since it's likely an important field like title, but It's not part of this issue so I decide to leave it untouched.

image


djaiss:
> However, when the entries are loaded in page like Journal.index the created_at is used instead of date_it_happened or any other, which might be very confusing if user input an activity with different date, I suggest that the date column should be in pivot table since it's likely an important field like title, but It's not part of this issue so I decide to leave it untouched.

That's an excellent comment. This definitely should be improved.

Thanks for your work. I'm going to review it now.

djaiss:
@Dagolin awesome work!

Dagolin:
@djaiss Thanks for the merge!

PR #1160, opened 2018-04-13
Improve cache handling

asbiin:
Improve cache handling in .htaccess file


asbiin:
Not sure we want to include this in the htaccess

PR #1167, opened 2018-04-13
Added docker for development support

Mokto:
Allows to have a quick development environment.

`docker-compose -f docker-compose.dev.yml up --build`


.dockerignore
```diff
@@ -41,6 +41,5 @@ storage/framework/cache/*
 storage/framework/sessions/*
 storage/framework/views/*
 storage/logs/*
-tests/
```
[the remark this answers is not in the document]
Mokto: Got it ! But it is needed for composer vendors installation.

The main issue in my opinion is this line on the prod Dockerfile.
ADD . .

We should specify which files will be added to the project in the dockerfile, not in the dockerignore.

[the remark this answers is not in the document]
Mokto: I just added phpmyadmin + volumes + removed tests directory from prod dockerfile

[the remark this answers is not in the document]
asbiin: I agree with this. I've listed a specific files/dir list to make the tar.gz in here: https://github.com/monicahq/monica/blob/master/Makefile#L66

Dockerfile
```diff
@@ -46,6 +46,7 @@ WORKDIR /var/www/monica
 # copy system files into their proper homes, and set file ownership
 # correctly
 ADD . .
+RUN rm -rf tests
```
asbiin: Could you move this line in the next `RUN` block? to limit the instruction number.

Dockerfile.dev
```diff
@@ -0,0 +1,69 @@
...
+ADD .env.dev .env
+RUN chown -R monica:monica . && \
+    chgrp -R apache bootstrap/cache storage && \
+    chmod -R g+w bootstrap/cache storage && \
+    # Apache2 conf
+    cp scripts/docker/000-default.conf /etc/apache2/conf.d/; \
+    # Composer installation
+    scripts/docker/install-composer.sh; \
+    # Set crontab for schedules
+    echo '* * * * * /usr/bin/php /var/www/monica/artisan schedule:run' | crontab -u monica -; \
```
asbiin: I think crontab is not needed for the env docker image (but it's not a problem)

djaiss:
@Mokto have you seen @asbiin 's comments?

Dockerfile.dev
Mokto: Won't we ever use this ?

Dockerfile
Mokto: I'm not sure to understand why. 

Don't we use https://github.com/jwilder/docker-squash ?

.dockerignore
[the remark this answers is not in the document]
Mokto: What do you prefer ? I'm not really confident modifying the prod docker image TBH.

Mokto:
@djaiss done answering.

Dockerfile
Mokto: Or another similar tool

asbiin: @Mokto nope! I was not aware of this tool. Anyway, for this dev image, it doesn't really matter

Dockerfile.dev
asbiin: @Mokto I think in dev, the easiest way is to call `php artisan monica:schedule` with command line, by demand.

theomathieubhvr: Ok I'll remove it

Mokto: Ok I'll remove the crons

Mokto:
Done

PR #1168, opened 2018-04-14
Indicate age at death for deceased people #1099 [Ready for Review]

### LINKED ISSUE #1099: Indicate age at death for deceased people
Feedback from user:

For contacts who are deceased, wouldn’t it be better to indicate their age at death versus the age they would be if they were still alive?  As cool as it is to see what they would be, IMHO I think age at death is probably more useful.

turtles2:
This closes #1099 By changing the age to the age at death See Screenshot. 

screenshot from 2018-04-13 19-38-59


app/SpecialDate.php
```diff
@@ -217,4 +217,23 @@ public function setToContact(Contact $contact)
         $this->contact_id = $contact->id;
         $this->save();
     }
+
+    public function getDeathAge()
```
djaiss: Could you add a DocBlock for new methods?

app/SpecialDate.php
```diff
@@ -217,4 +217,23 @@ public function setToContact(Contact $contact)
...
+    {
+        if (is_null($this->date)) {
+            return;
+        }
+
+        if ($this->is_year_unknown) {
+            return;
+        }
+
+        $contact = Contact::findorfail($this->contact_id);
```
djaiss: `findOrFail`

djaiss: Dismiss last comment: you don't have to find the contact. You can access the contact by using `$this->contact` as they are linked with a `belongsTo` relationship.

app/SpecialDate.php
djaiss: Also, a better name would be `getAgeAtDeath()`

resources/views/people/_header.blade.php
```diff
@@ -27,10 +27,14 @@
...
           <h2 class="{{ \App\Helpers\LocaleHelper::getDirection() }}">
             {{ $contact->getCompleteName(auth()->user()->name_order) }}
-            @if ($contact->birthday_special_date_id)
+            @if ($contact->birthday_special_date_id && !($contact->is_dead))
               @if ($contact->birthdate->getAge())
                 <span class="ml3 light-silver f4">(<i class="fa fa-birthday-cake mr1"></i> {{ $contact->birthdate->getAge() }})</span>
               @endif
+            @elseif ($contact->is_dead)
+                @if (! is_null($contact->deceasedDate))
+                  <span class="ml3 light-silver f4">(Age at Death {{ $contact->deceasedDate->getDeathAge() }})</span>
```
djaiss: Multiple comments there:
- don't forget to add new translation keys instead of writing plain English in Blade files (by adding your string to `en/people.php` for instance).
- it's not obvious that you need to call the deceased date on the contact object to find out what the age at death was. I think it would be better to call a method on the contact itself, like `$contact->getAgeAtDeath()`. Then, this method can call a new method in the SpecialDate class to do the trick. That way, the Contact object doesn't care how it should get this age. It just knows that it has a method to do it for him.
- However, if you still want to use something like `$contact->deceasedDate->`, then you need another method, more generic, instead of the one you wrote. Carbon provides a `diffInYears` helper that you can probably use, like `$contact->deceasedDate->diffInYears($contact->birthdate)` or something.

The last two options work and I have no real preference for any of them.

djaiss:
Thanks @turtles2 for your patience!

I'm going to merge this after my work day today.

PR #1178, opened 2018-04-15
add language dropdown in registration page

### LINKED ISSUE #1003: Add langage listbox in registration page
We should add a langage listbox in the registration page, so the user can create a new account with its good locale directly.

Registration page could use the `Accept-Language` header to guess the langage of the user ...

Dagolin:
Closes #1003 

Yes, thank you @asbiin,  it's a good idea to let people choose the language at the beginning, like me, I am not a Simplified Chinese speak so when Monica forced me to use it, it was confusing.

image

### DIFF
--- a/app/Account.php
+++ b/app/Account.php
@@ -652,9 +652,10 @@ class Account extends Model
      * @param string $last_name
      * @param string $email
      * @param string $password
+     * @param string $locale
      * @return this
      */
-    public static function createDefault($first_name, $last_name, $email, $password)
+    public static function createDefault($first_name, $last_name, $email, $password, $locale = null)
     {
         // create new account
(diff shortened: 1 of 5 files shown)


asbiin:
@Dagolin nice work! You just have to fix the issue with StyleCI and it could be merged.

asbiin:
I think the language dropdown should not be a part of the registration form, but it should permit to reload the page for another language - which will be used as user language.
This allow a user to load the registration form for his language, and understand all fields.

This language dropdown could be on upper left/right of the page.
Problem with this: the current lang-detector.driver cannot be changed at runtime
https://github.com/monicahq/monica/blob/master/config/lang-detector.php#L14

Issue #1213, opened 2018-04-25
SQL export assumes a "monica" database


tomzx:
https://github.com/monicahq/monica/blob/a6ac8a8f5fa526f4f903d80a9c3896b670c6f0bd/app/Jobs/ExportAccountAsSQL.php#L86

Seems it would be more appropriate to look for the current connection database (or the environment `DB_DATABASE`).

djaiss:
Ohhh that is a nice catch. Good one.

tomzx:
Apparently there's many more:

https://github.com/monicahq/monica/blob/da73a6e15d9d06702116fdc89d207d417ebef6ff/app/Http/Controllers/SettingsController.php#L107

https://github.com/monicahq/monica/blob/da73a6e15d9d06702116fdc89d207d417ebef6ff/app/Http/Controllers/SettingsController.php#L145

https://github.com/monicahq/monica/blob/da73a6e15d9d06702116fdc89d207d417ebef6ff/app/Http/Controllers/Api/ApiContactController.php#L346

https://github.com/monicahq/monica/blob/da73a6e15d9d06702116fdc89d207d417ebef6ff/app/Contact.php#L1241

djaiss:
Oops.

PR #1214, opened 2018-04-25
Add foreign keys

tomzx:
This change adds foreign keys to all tables.

The test have been run and they pass, but by playing around I was still able to find things that were broken, meaning that it is likely that things may be broken and have no tests that detect them.

- [ ] Impact on account export.
See #1213.
Seems to work. Can someone that knows better confirm?
- [ ] Impact on account reset and deletion.
Seems to still work (reports Your account has been reset successfully). However, it does not appear to delete contacts. Thus it is broken, but the UI is reporting a success. This needs to be fixed.

# Notes of changes occurring in the meantime
* [x] `addresses.country_id` column is dropped


tomzx:
@djaiss @asbiin Other than fixing the build, anything else needed to get this merged?

tests/Unit/AccountTest.php
```diff
@@ -31,20 +32,23 @@ public function test_it_has_many_genders()
 
     public function test_it_has_many_notifications()
     {
-        $account = factory('App\Account')->create([]);
+        $contact = factory(Contact::class)->create();
```
djaiss: I don't understand why you've changed this kind of things. What was wrong with `App\Account`? At least you don't need to import all the classes at the top of the test files that way.

tests/Unit/AccountTest.php
```diff
@@ -540,18 +548,20 @@ public function test_it_skips_default_relationship_types_table_for_types_already
...
-        $contact = factory('App\Call', 2)->create([
-            'account_id' => $account->id,
+        $calls = factory('App\Call', 2)->create([
+            'account_id' => $contact->account_id,
+            'contact_id' => $contact->id,
             'called_at' => '1992-03-02',
         ]);
 
-        $statistics = $account->getYearlyCallStatistics();
+        $statistics = $contact->account->getYearlyCallStatistics();
```
djaiss: So here you chose to keep `$contact->account` 😀 Why?

tests/Unit/ContactTest.php
```diff
@@ -586,20 +596,46 @@ public function testTotalOutstandingDebtAmountIsCorrect()
...
+        ]));
 
         $this->assertEquals(200, $contact->totalOutstandingDebtAmount());
 
-        $contact->debts()->save(new Debt(['in_debt' => 'yes', 'amount' => 100]));
+        $contact->debts()->save(new Debt([
+            'in_debt' => 'yes',
+            'amount' => 100,
+            'account_id' => $contact->account_id,
+            'contact_id' => $contact->id,
```
djaiss: You don't need that. `$contact->debts...` will already set the contact_id for you

tests/Unit/ContactTest.php
```diff
@@ -661,16 +697,14 @@ public function test_it_returns_an_unknown_birthday_state()
...
-        ]);
+        $contact = factory(Contact::class)->create();
         $specialDate = factory(SpecialDate::class)->create([
             'is_age_based' => 1,
         ]);
         $contact->birthday_special_date_id = $specialDate->id;
         $contact->save();
 
-        $specialDate->contact_id = $specialDate->id;
+        $specialDate->contact_id = $contact->id;
```
djaiss: Ahah good catch. How come this wasn't triggering an error?

tests/Unit/Jobs/SendRemindersTest.php
```diff
@@ -38,33 +38,6 @@ public function test_it_schedules_a_reminder_email_job_and_a_set_next_expected_d
         Bus::assertDispatched(SetNextReminderDate::class);
     }
 
-    public function test_it_deletes_the_reminder_if_contact_doesnt_exist()
```
djaiss: This method should not be deleted. I also notice that the method name is wrong. But the method tests that the reminder is not sent if the contact does not exist (and not even dispatched) - which is a valuable test in my opinion.

tests/Unit/TagTest.php
```diff
@@ -26,11 +26,11 @@ public function test_it_belongs_to_many_contacts()
         $account = factory('App\Account')->create([]);
         $contact = factory('App\Contact')->create(['account_id' => $account->id]);
         $tag = factory('App\Tag')->create(['account_id' => $account->id]);
-        $contact->tags()->sync($tag->id);
+        $contact->tags()->sync([$tag->id => ['account_id' => $account->id]]);
```
djaiss: No you don't need that to test what the tests do.

app/Contact.php
```diff
@@ -1078,12 +1079,12 @@ public function getIntroducer()
...
     public function setSpecialDate($occasion, int $year, int $month, int $day)
     {
-        if (is_null($occasion)) {
+        if (null === $occasion) {
             return;
         }
 
         $specialDate = new SpecialDate;
-        $specialDate->createFromDate($year, $month, $day)->setToContact($this);
+        $specialDate->setToContact($this)->createFromDate($year, $month, $day);
```
djaiss: Why do you change the order?
With your solution the specialDate is saved before it's created. What was wrong with how it was done before?

database/migrations/2018_04_21_175153_normalize_foreign_key_columns_to_unsigned_integer.php
```diff
@@ -0,0 +1,334 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('activities', function (Blueprint $table) {
+            $table->unsignedInteger('account_id')->change();
```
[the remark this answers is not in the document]
tomzx: Because you cannot create foreign keys from unsigned (primary keys) to signed (foreign keys).

tests/Unit/AccountTest.php
tomzx: Because it makes it easier to rename the class, if that ever happens. The issue with using the string approach is that you MUST know that they are used, while an IDE such as PHPStorm will make it easy to rename a class and all its references.

tests/Unit/AccountTest.php
tomzx: I'm not sure I understand what you mean. In this case we generate a contact instead of an account, which has an account generated at the same time. This is necessary since we need to assign both the contact_id and account_id, which wasn't done previously.

tests/Unit/ContactTest.php
tomzx: Yes, but it will not set the account_id. I can remove the `contact_id` line if you want, but the `account_id` line still needs to be there.

tests/Unit/ContactTest.php
tomzx: It's an `id`, so it's technically valid :laughing:.

tests/Unit/Jobs/SendRemindersTest.php
tomzx: I removed it because it's not going to be technically possible anymore.

Once you delete a contact, due to the cascade constraints, the reminder will get deleted as well. So it's impossible for the contact of a reminder not to exist.

However this is because I set the contact_id to be mandatory (a constraint) for a reminder, maybe that isn't right?

tests/Unit/TagTest.php
tomzx: Yes it's necessary, for the same reason as listed above (the debt example).

app/Contact.php
tomzx: Because you cannot create a special date without specifying the associated contact/account :stuck_out_tongue_winking_eye:.

However I'd prefer to get the calls to `save` out of the model so that the order would not matter. And also have `createFromDate` not do `save`.

This kind of logic should go into a service (a class dedicate to a specific purpose).

tomzx:
I agree it's pretty big.

I think we should try to get all the fixes to the tests in the main branch and merge into this branch as we go so we can keep the tests working correctly once we merge this in. I was able to find various issues just by doing that, so I'm not surprised there's a lot still "hidden under the hood".

app/Contact.php
```diff
@@ -1078,12 +1079,12 @@ public function getIntroducer()
      */
     public function setSpecialDate($occasion, int $year, int $month, int $day)
     {
-        if (is_null($occasion)) {
+        if (null === $occasion) {
```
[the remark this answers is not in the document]
djaiss: 👍

database/migrations/2018_04_21_175153_normalize_foreign_key_columns_to_unsigned_integer.php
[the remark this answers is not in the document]
djaiss: 👍

tests/Unit/AccountTest.php
djaiss: Interesting, thanks for the explanation

tests/Unit/AccountTest.php
djaiss: I mean that above in this file, you've assigned `$account = $contact->account;` and then use `$account` in the methods. 

In this case you've kelp `$contact->account`

tests/Unit/ContactTest.php
djaiss: Awesome.

tests/Unit/ContactTest.php
djaiss: One more good one.

tests/Unit/Jobs/SendRemindersTest.php
djaiss: It is right, for now at least. So no need to deal with that for now on.

tests/Unit/TagTest.php
djaiss: Thanks!

app/Contact.php
djaiss: Interesting. Thanks a lot for your patience in explaining all this to me. As you can see, I don't have your knowledge nor skills 😀

djaiss:
> I think we should try to get all the fixes to the tests in the main branch and merge into this branch as we go so we can keep the tests working correctly once we merge this in. I was able to find various issues just by doing that, so I'm not surprised there's a lot still "hidden under the hood".

In practice how do you propose that we move forward with this?

tests/Unit/AccountTest.php
tomzx: Oh right. Good observation. I think I just tried to modify as little code as possible. As you can see from the class, there's a lot that could be changed to make the style more uniform (use classes everywhere instead of strings, use `$account` instead of `$contact->account`, etc.).

djaiss:
Alright let's try to do it! That's going to be a hell of a challenge.

tomzx:
@djaiss I see you added a new table `CreateTermsTable` which should use foreigns keys. Is there any reason we don't use foreign keys on tables that are being created?

Not doing so means updating this PR with the necessary changes, which might make things more complicated as these changes will have be tracked and reviewed.

djaiss:
@tomzx because I'm stupid and I forgot :-/ It's still time for me to change it since I haven't updated the production yet.

djaiss:
@tomzx what do you think we should as the next steps?

tomzx:
@djaiss #1213 needs to be merged, then we need to fix the account reset. After that a manual test pass over most common features should be done. Once all of that is done, I think we could merge this into master and fix issues as they come.

djaiss:
@tomzx alright, I'm going to review all the important flows to make sure we haven't broke something. Then I'm also going to check the Account reset process.

tomzx:
@djaiss It might also be worth reviewing the list of "on update/delete" actions on foreign keys (the `AddMissingForeignKeys ` migration class) to make sure it is the behavior you guys expect in all cases. I've likely made a few errors given I don't know all of the use cases.

djaiss:
I keep this PR for this weekend as it will take me a lot of time to test everything. Thanks for your patience @tomzx  😀

djaiss:
@tomzx to prepare this, I've started to implement end 2 end testing (https://github.com/monicahq/monica/pull/1403). This will let me automate testing all the major flows of the app. We'll merge it then we'll address this PR here - that way we'll be more sure that this doesn't break anything important.

asbiin:
@djaiss should we close this one as you created #2280 ?

PR #1218, opened 2018-04-26
Fix StayInTouch i18n support

asbiin:
Some messages are not i18n compliant.


djaiss:
Nice! If it works, it's even better 😀

djaiss:
Your changes have led to some JS error in the console.

```js
[Vue warn]: Invalid prop: type check failed for prop "value". Expected String, got Number.

found in

---> <FormInput> at resources/assets/js/components/partials/form/Input.vue
       <SweetModal> at node_modules/sweet-modal-vue/src/components/SweetModal.vue
         <StayInTouch> at resources/assets/js/components/people/StayInTouch.vue
           <Root>
vue.common.js:593
```

This is what I had before and this is why I had to change that to strings.

asbiin:
@djaiss indeed, I've fixed it, thank you !

Issue #1059, opened 2018-03-26
[Feature Request] Add a "Today" option to the date picker for "How I met X" page


heliostatic:
I usually add contacts the day I meet them, so it would be great to have "Today" as a shortcut on the "How I met X" screen:
 
<img width="449" alt="screenshot 2018-03-26 08 57 00" src="https://user-images.githubusercontent.com/70898/37907700-d4ebe5de-30d3-11e8-8409-1e218851ad05.png">

asbiin:
Closed by #1256

PR #1256, opened 2018-05-01
Make select-date component works with SpecialDate object as an Carbon instance

### LINKED ISSUE #1058: Modify a few pages so the app is fast even with a huge number of contacts
We have a few users with thousands of contacts (one with 8000+, a few above 5000+, and more around 1000-3000). While Monica wasn't designed for having that many contacts, we can be better at managing this.

For instance, when adding a new relationship, a good iteration would be to load either only a X number of contacts, or the X number of most recent contacts, then also have a search box at the top so we can actually search for more if needed.

Dagolin:
Fixes #1058

The original  `select-date` component only accept default date as a `SpecialDate` model object, I update it to accept `Carbon` as well, and give a default value to `how I meet X ` date picker

image


djaiss:
Awesome, thanks for this!

However I think you are wrong in the issue you mention in your comment - it's not related (#1058).

Dagolin:
@djaiss My bad, thanks for the correction! Its related to #1059 not #1058.  should stop drinking :/

Issue #1211, opened 2018-04-24
Docker Update Fail


fmcpherson:
Previously running Monica 1.8.2 via the images hosted on Docker Hub. I did a docker-compose pull today to update my images but after docker-compose up I get MYSQL errors resulting on it being shut down and therefore Monica not successfully coming up. See the attached screenshot showing the error. I am going to revert back to a snapshot of the prior version. 

![startuperror](https://user-images.githubusercontent.com/7051925/39215270-3ec7b190-47e5-11e8-8ee0-9dc51bfb5519.png)

djaiss:
Damn. Sorry about that. We'll fix it.

asbiin:
@fmcpherson See comment in #1204 :
https://github.com/monicahq/monica/issues/1204#issuecomment-385285043
It seems you are using mysql 8.0, and revert it to mysql 5.7 fix the problem.

fmcpherson:
I am using the official images in Docker Hub, not creating my own. The whole point of Docker is to eliminate these type of dependencies, so it shouldn't matter which version of mysql is on my server.

asbiin:
Reverting to mysql 5.7 solve the problem

PR #1268, opened 2018-05-03
Fix for #1211 - Set fixed version for MySQL in docker-compose

kidk:
Fix for #1211

I hard coded MySQL to version 5.7, as 8.0 does not work at the moment.


djaiss:
Thanks a lot @kidk for your help! It's really appreciated.

@asbiin could you review this one? Thanks a lot to you too!

And thanks to me for thanking both of you! 😀

PR #1277, opened 2018-05-05
Add changelog entry about disabling birthday reminders

djaiss:


shaananc:
vCard added contacts still suffer from this problem.
There should also probably be a global switch for birthday reminders.

PR #1282, opened 2018-05-05
Build absolute path to stubs files in UploadVCardTest and UploadVCardsTest

tomzx:
This is necessary if the working directory is not set (or not the root of the project) when executing PHPUnit.


djaiss:
Woot 👍 
Thanks a lot for this!

PR #1289, opened 2018-05-08
Fix tests to prepare for foreign keys

tomzx:
This is the changes applied to #1214 in order to fix the tests in order to prepare for foreign keys to be implemented.

This is done separately so we can get those fixes in now until we get the foreign keys PR merged.


djaiss:
Excellent job (also, insane job).
We are good to merge as soon as the conflicts are resolved!

PR #1292, opened 2018-05-08
Add api tests for addresses

asbiin:


tests/Api/ApiAdressesTest.php
```diff
@@ -131,4 +130,115 @@ public function test_address_get_addressid()
...
+            'contact_id' => $contact->id,
+            'name' => 'address name',
+            'street' => 'street',
+            'postal_code' => '12345',
+            'country' => 'FR'
+        ]);
+
+        $response->assertJsonFragment([
+            'object' => 'address',
+            'id' => 1,
```
djaiss: IDs shouldn't be hardcoded. In my locale environment, each time I run the tests, the ID increments, making this test fail everytime.

djaiss: @asbiin ping.

djaiss: Here you can't really test the ID number. You can test the presence of an ID though, or you could test that the table `addresses` contains no row, then one row after your update is complete.

asbiin: Good idea

asbiin: Fixed by #1333

PR #1295, opened 2018-05-08
Fix API locale

asbiin:
Locale was not set on API calls


asbiin:
Some responses depend on the locale of the user, like the list of countries.

PR #1296, opened 2018-05-09
Add GDPR compliance

djaiss:
This PR adds support for Monica to be compliant with policies required by law. We need this for the upcoming GDPR and probably new laws after that.

Basically, when we change the terms of use and the privacy policy, we need to ask users if they want to accept these new terms before they can continue using the software. We also need to record when they've accepted them, and from which IP addresses.

I've added new API methods to cover all those new use cases. In the process, I've added a new User object in the API, as well as a new Currency object.

#### Code-related tasks
- [x] Impact on the seeders.
- [x] Impact on the API.

- [x] Impact on account export.
- [x] Impact on account reset and deletion.

#### Other tasks
- [x] If it's relevant and worth mentioning, create a changelog entry for this change. The changelog entry will appear inside the UI for all users to see. To know if your change is worth the creation of a changelog entry, [read the documentation](https://github.com/monicahq/monica/wiki/Advanced-content-tips-for-developers#when-is-it-relevant-to-create-a-changelog-entry).
- [x] Don't forget to [ask for a free account](mailto:regis@monicahq.com) on https://monicahq.com as anyone who contributes can request a free account.


app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -70,6 +70,7 @@ protected function validator(array $data)
             'first_name' => 'required|max:255',
             'email' => 'required|email|max:255|unique:users',
             'password' => 'required|min:6|confirmed',
+            'gdpr' => 'required',
```
asbiin: You should not call every variables like "gdpr". OK, the aim of ths PR is to solve the GDPR question, but this functionnnality is more than that. What if an other government create a new law ? You want to have an other variable with the second law's name ?

resources/views/people/index.blade.php
```diff
@@ -163,6 +163,9 @@
                     <span class="number-contacts-per-tag {{ \App\Helpers\LocaleHelper::getDirection() }}">{{ trans_choice('people.people_list_contacts_per_tags', $dbtag->contacts()->count(), ['count' => $dbtag->contacts()->count()]) }}</span>
                 </li>
                 @endif
+                <li class="f7 mt3">
```
asbiin: Thay should not be here

resources/views/auth/register.blade.php
```diff
@@ -55,12 +55,16 @@
                 <input type="password" class="form-control" id="password_confirmation" name="password_confirmation" required>
               </div>
 
-              <div class="form-group actions">
-                <button type="submit" class="btn btn-primary">{{ trans('auth.register_action') }}</button>
+              <!-- GDPR compliance -->
```
asbiin: It's more like "policy acceptance check"

resources/views/people/index.blade.php
asbiin: Well, you left this in your merge ..

PR #1308, opened 2018-05-11
Fixes #1209 Add journal entry with date != today does not work

### LINKED ISSUE #1209: Add journal entry with date != today  does not work
Adding a journal entry with a different date is not possible.

Entries can only be created for the same day. Thus, the datepicker is ignored 

version 2.0.1

turtles2:
This fixes #1209 by getting that date when the Journal entry is an instance of Entry


djaiss:
image

sfeu:
Already looking forward to the next release :-)

PR #1309, opened 2018-05-11
fix: postgresql specific fix for casting a varchar to int

i3anaan:
### Checklist
Closes https://github.com/monicahq/monica/issues/137 and fixes a crash during migration, attached at the bottom.

See: https://stackoverflow.com/questions/13170570/change-type-of-varchar-field-to-integer-cannot-be-cast-automatically-to-type-i

Not sure what tests to run, as they would need to be reconfigured for Postgres usage... Works for me? :)
(I'm also a bit tired now, if more tests are required apart from the automated ones I'll have a look at it tomorrow)

**Happy to finally have Monica working with a postgres db!**

```
monicahq_1  |   SQLSTATE[42804]: Datatype mismatch: 7 ERROR:  column "about_object_id" cann  
monicahq_1  |   ot be cast automatically to type integer                                     
monicahq_1  |   HINT:  You might need to specify "USING about_object_id::integer". (SQL: AL  
monicahq_1  |   TER TABLE gifts ALTER about_object_id TYPE INT)                              
monicahq_1  |                                                                                
monicahq_1  | 
monicahq_1  | In PDOStatement.php line 143:
monicahq_1  |                                                                                
monicahq_1  |   SQLSTATE[42804]: Datatype mismatch: 7 ERROR:  column "about_object_id" cann  
monicahq_1  |   ot be cast automatically to type integer                                     
monicahq_1  |   HINT:  You might need to specify "USING about_object_id::integer".           
monicahq_1  |                                                                                
monicahq_1  | 
monicahq_1  | In PDOStatement.php line 141:
monicahq_1  |                                                                                
monicahq_1  |   SQLSTATE[42804]: Datatype mismatch: 7 ERROR:  column "about_object_id" cann  
monicahq_1  |   ot be cast automatically to type integer                                     
monicahq_1  |   HINT:  You might need to specify "USING about_object_id::integer".           
monicahq_1  |                                                                                
monicahq_1  | 
```


database/migrations/2017_10_14_083556_change_gift_column_structure.php
```diff
@@ -15,7 +15,12 @@ class ChangeGiftColumnStructure extends Migration
     public function up()
     {
         Schema::table('gifts', function (Blueprint $table) {
+          if (get_class(DB::connection()) == "Illuminate\Database\PostgresConnection") {
```
asbiin: You can do `if (DB::connection()->getDriverName() == 'pgsql')`

i3anaan: Ah thanks! I was actually looking for this but was unable to find something like it. I will update the PR!

asbiin:
Reviewed this PR, it's OK ! @i3anaan you just have to fix StyleCI syntax and we can go.

i3anaan:
@asbiin Should be good now!

asbiin:
@i3anaan thank you !

Issue #1224, opened 2018-04-28
Is it possible to use the default local mail server?


Erwyn:
Hello there,

I would like to know if it's possible not to configure an external mail server but use the default direct one? How to do so?

turtles2:
@Erwyn I don't know enough about email to answer your question directly. But I know that Monica is based on Laravel. Which offers many ways to send emails. Read this [doc](https://laravel.com/docs/5.6/mail) to learn about the various ways laravel can send email out of the box. 

@djaiss Correct me if I am wrong. But Monica does use the Laravel Mail feature?

Erwyn:
@turtles2 I'm gonna take a look at this thanks

djaiss:
@turtles2 yes, Monica uses the built-in Mail feature - so you can use https://github.com/monicahq/monica/blob/master/config/mail.php to configure with a variety of providers.

djaiss:
@Erwyn have you succeed?

Erwyn:
Hello Régis,

Sorry haven't got the time to try yet, I'll let you know when I'm on it.

2018-04-30 1:29 GMT+02:00 Régis Freyd <notifications@github.com>:

Erwyn:
So haven't found relevant documentation yet, so I think I'm going to use the Gmail SMTP instead.

djaiss:
I wonder if Gmail will allow it.

Erwyn:
How do you guys do it? you use mailchimp?

djaiss:
[Postmark](https://postmarkapp.com/). Always loved them.

cruisemaniac:
I run my own mail server and just set it up to work with monica. Like its mentioned above in the comments, all you're required to do is to put in the corresponding settings in the `MAIL_` section in the `.env` file.

This really doesnt have anything to do with monica per-se and is more of the laravel standard feature set.

Issue #870, opened 2018-01-31
Add ability to send emails to Monica and associate it with a contact


djaiss:
Also called inbound email processing.

We forward an email to Monica and it automatically saves it to the sender's contact entry on the server.

wearpants:
+1 - mainly using monica for business leads, this would seriously streamline my workflow

turtles2:
@djaiss I would like to work on the back-end for this. Since the hosted version already uses Postmark my plan is to work with there API unless I can find a good Laravel package.

turtles2:
I am thinking that we use the sending address on the email to tie to a user as we are assuming its being forwarded from there email correct?

djaiss:
@turtles2 that would be excellent. I found a package for this a while ago, don't remember the name...

I don't know how you want to do that, but here is what I think we should do: follow what [Highrise does](https://help.highrisehq.com/email/email-intro/#how-do-you-find-your-dropbox-address). I don't see any other way actually. Can you explain how you want to do this? This will also require some sort of MX field to configure, I assume...

turtles2:
I am on Mobile so I won't go into details but hopefully tonight or tomorrow
I will push to the PR and you will be able to see what I have in mind. It
is going to be a rough proff of concept.

tomgs:
So, just to wrap up, I guess this can be closed due to https://github.com/monicahq/monica/pull/1311 (well, due to https://github.com/monicahq/monica/pull/1311#issuecomment-475104214 specifically), right?

tomgs:
I find that one of the foremost issues for me is actually following up on e-mail conversations. If it's WhatsApp or LinkedIn then I make notes, but e-mail... I'd much prefer to just forward the e-mails to monica.

PR #1311, opened 2018-05-11
Backend Code for #870 Add ability to send emails to Monica and associate it with a contact [WIP]

turtles2:
This PR will add the backend code for #870. This PR only is for getting the emails and parsing them into the DB. This code assumes the inbound processor is Postmark as that is what the hosted version uses per #1224 and I could not find a good library to add support for others. It also assumes that users are forwarding the email into Monica from there account. This could be changed but we need a way to link an email to a user.

- [x] Add Middleware to Whitlist Postmark IPs for security 
- [ ] Phrase the email body to get data which includes contact and message data 
- [x] Add table in DB to store email information
- [ ] Add tests where needed


turtles2:
@djaiss Can you take a look at this PR right now it is a very rough proof of concept. What I would like focused on is the workflow lets not worry too much about the specifics of the code at this juncture. Here is the current workflow. 

1. User Forwards Email to Inbound Address one address for all users
2. Receive Webhook from Postmark
3. Find user based on Source of Forward Email
4. Extract Date, To, From form the most recent email in the chain
5. Find all of the users contacts that have the to and from emails
6. Save the email to the DB it will save the plain text of the whole chain similar to printing an email
7. Link the contacts to the Email. It is a Many to Many

djaiss:
Great approach.

We need to make sure that the emails do not appear in the wrong account, as users could potentially have the same contact email address for a similar contact... This is the top priority I think.

1. I think this feature should be a premium feature (many businesses will use it).
1. Unfortunately we need to create a new `env` variable for this feature, as we need to provide an email address for the instance (which will come from Postmark), something like `INBOUND_EMAIL_ADDRESS`. That means if the env variable is not there, this feature should not work and should also degrade gracefully - in the sense that it should not crash the instance. We should simply refuse the webhook.
1. I don't think many users will be happy that we support a paid external service for this feature. That being said, we haven't found a great parsing library yet. Postmark offers 100 emails per month for free. [Mailgun](https://www.mailgun.com/pricing) seems to offer 10 000 emails for free. Could we do a system where we could accept mutiple services like that easily? I know it's not a trivial task. Just asking in case you would think of something.
1. I think emails should be put in the same section as the notes in a contact sheet (even though the objects are different by nature, but I have an idea of how we could represent it). I would like to provide a design for this. Could you simply make a list of those emails above the Notes section and I will do a design in a separate branch for this?
1. Even though I won't review the code right now (just a bit), I've seen that you are calling the object `Email`. I think it's a bit confusing for new comers who won't get immediately what an Email is. What do you think of calling the class `InboundEmail` instead?

routes/api.php
```diff
@@ -2,6 +2,8 @@
 
 use Illuminate\Support\Facades\Route;
 
+Route::post('/emailhook', 'InBoundEmailController@new');
```
djaiss: `/inboundemail` might be a better name.

turtles2: I agree

turtles2:
@djaiss Some of the points you made have left me questioning how you implemented the hosted version. Do you have a running instance for each user? If not comfortable answering publicly let me know and we can take that discussion privately because for this feature I think it is key for me to understand how you are running things. Below I will respond to the various points you raised. 

1. I agree
2. No as Postmark dose not send any emails to the server. It simply posts a webhook to an address you set in Postmark. We keep things secure by only allowing requests from Postmark IPs
3. We could we would just need a separate method on the controller for each service. I only chose Postmark as that it what the hosted version uses from my understanding. Again this is why I need a better understanding of how you are running things. Myself I prefer Mailgun I have used them in other projects.  We could IMAP or POP which there are libraries for but I think that it would lead to a sub par user experience as set up would be difficult on there end. Not to mention the load on the servers needing to be requesting and processing every email on the regular instead of waiting for a webhook. 
4. On this PR I would prefer not to do anything with the front end to keep things simple on my end. I would like to leave it all to you or somebody else.
5. I have not done anything on the API side yet for this project but I will give it a shot. Can you clarify on using the APIs for the front end?
6. I agree

I will be awaiting a response before preceding with any work on this PR.

On a side note I have some experience with DNS and infrastructure work not too long ago I was very close to starting my own ISP I was just short on funding and had some bad breaks. So I have a solid grasp on what is at play here.

djaiss:
@turtles2 Sorry for the delay - I'll check this over the weekend 😀

turtles2:
@djaiss No worries, I feel like this one might take a bit to get right.

app/Http/Controllers/InBoundEmailController.php
```diff
@@ -0,0 +1,101 @@
...
+     * Create a new controller instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        $this->middleware('webhook');
+    }
+
+    public function new_postmark(Request $request)
```
djaiss: I'm not convinced to call this method `new_postmark`.

I don't think we treat the problem the right way. The way we associate an email to an account doesn't seem to be super secure.

We can't rely on the from address to find an user account - this can be super easily spoofed and then we'll associate an email to the wrong account.

In my opinion:
1. we should create an unique ID for each user in the database, that we will append to the email address people will use to contact Postmark after the `+` sign, like `dropbox+HASHID@postmark`.
1. Postmark will give us the `hashid` in the `MailboxHash` parameter in the JSON they send back.
1. This will let us identify, with the combination of the `from` email address, the user we need to associate the email with.
1. Then we proceed.

routes/api.php
```diff
@@ -2,6 +2,8 @@
 
 use Illuminate\Support\Facades\Route;
 
+Route::post('/inboundemail/postmark', 'InBoundEmailController@new_postmark');
```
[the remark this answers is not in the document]
turtles2: I put it there as API routes don't have CSRF protection this is the convention Laravel recommends
https://laravel.com/docs/5.6/csrf#csrf-excluding-uris

app/Http/Controllers/InBoundEmailController.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $email_data = [
+          'from_email' => '',
+          'to_email' => '',
+          'subject' => '',
+          'datetime' => '',
+        ];
+
+        $pattern = '/From:.{1,}<(.{1,})>/m';
+        if (preg_match($pattern, $email_contents, $matches)) {
```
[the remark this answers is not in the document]
turtles2: We need to this as the from email we get from the JSON is the email that forwarded the message this gets us the messages from email.

app/Http/Controllers/InBoundEmailController.php
turtles2: The method only works for postmark that needs to be clear. Later we could add new_mailgun and other services. Perhaps new_inbound_postmark?

Good point I agree on the hashing. I would prefer to use the existing hash method it would be easy to expand for users. It might be harder to make it shorter though. Do you think we should aim for a short hash or not care?

turtles2:
@djaiss I have commented  on your review. Please go ahead and respond where needed. Hopefully I will do some work on this over weekend

app/Http/Controllers/InBoundEmailController.php
djaiss: @turtles2 the problem with the current hash is the fact that it's prefixed with `h:` and I don't know how the `+` sign will work with a `:` after it... so that means we would need to get rid of this `:`. By the way, I think the longer the key after the `+` sign, the better.

routes/api.php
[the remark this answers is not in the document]
djaiss: Interesting, now that you mention it, it makes sense.

app/Http/Controllers/InBoundEmailController.php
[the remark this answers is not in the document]
djaiss: Ohhhhh INSIDE the email content, right. I get it now (I should call myself a turtle as I'm slow).

app/Http/Controllers/InBoundEmailController.php
djaiss: Also, @asbiin is in the process of adding email verification to Monica. We should make sure that we will only accept (as the `from` address) an email that has been verified already.

djaiss:
@turtles2 I think you can proceed. Thanks for your awesome work and continuous contributions.

app/Http/Controllers/InBoundEmailController.php
turtles2: It has been a bit since I implemented it but I feel like adding a method that meets what we need for the email will be easy I will did into it this weekend hopefully

turtles2: @djaiss You are correct that : dose not work. But a . will work. We could change the preflix or do a replace on the string before giving it to hashIds?

turtles2:
@djaiss I have fixed the naming issues as well as implemented HashIDs for the mailbox hash. It uses the raw ID without the prefix. If you could go ahead and take another look that would be awesome. All the added code will have tests once I am confident in the behavior I need to be testing.

djaiss:
@turtles2 great, thanks a lot. I'll review all this very soon!

turtles2:
@djaiss or @asbiin Git is not my strong suite. Can you help get this branch even with master? Because of the file structure change it is very out of date.

djaiss:
@turtles2 I'll take care of it. Yeah master has changed drastically.

turtles2:
@djaiss When you are helping me out with that, can you go ahead and review this, as I have been waiting on that.

djaiss:
https://laravel-news.com/laravel-inbound-email

djaiss:
We are going to replace this by https://github.com/beyondcode/laravel-mailbox instead.

I'm really sorry @turtles2 that we didn't follow through with this awesome PR.

PR #1312, opened 2018-05-12
Add trusted proxies middleware to allow monica running behind ssl terminating loadbalancer (#1287)

### LINKED ISSUE #1287: Connection not secure with production enabled
**Describe the bug**
When connecting to self-hosted Monica, Firefox complains about an unsecure connection even though app_env=production is enabled and SSL is configured through traefik. It says that only some information is unsafe, such as submitting forms or images.

**Screenshots**
If you can, add screenshots to help explain your problem. An image is always helpful.
version 2.1.0

**Additional context**

I have my monica instance behind a reverse proxy, so basically it is monica.personalserver.com > https > traefik reverse proxy > http > monica docker with production app_env enabled.

any ideas how to fix this?

danielpieper:
This fixes #1287 by adding the expected behavior. I'm not sure if this feature should be enabled by default or by an option.


danielpieper:
I would also go with an empty default value. I'm just facing the issue that it's not possible to add multiple proxies to the environment variable. Either limit them to * or one ip address or allow a comma separated list...

asbiin:
"allow a comma separated list" yes this is right!

asbiin:
@danielpieper very nice! Thanks !

PR #1313, opened 2018-05-12
Run postgresql tests with CircleCI

asbiin:


i3anaan:
Nice, I am very much in favor of adding tests for postgresql :)

asbiin:
This is not working. Postgresql support must be review from the beginning.

PR #1333, opened 2018-05-13
Remove id from Address tests

asbiin:
This follow comment here: https://github.com/monicahq/monica/pull/1292#discussion_r187802050

### DIFF
--- a/tests/Api/ApiAdressesTest.php
+++ b/tests/Api/ApiAdressesTest.php
@@ -157,7 +157,6 @@ class ApiAdressesTest extends ApiTestCase
 
         $response->assertJsonFragment([
             'object' => 'address',
-            'id' => 1,
             'name' => 'address name',
             'country' => [
                 'object' => 'country',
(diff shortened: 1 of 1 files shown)


tests/Api/ApiAdressesTest.php
```diff
@@ -168,14 +167,16 @@ public function test_address_post()
             'street' => 'street',
             'postal_code' => '12345',
         ]);
+
+        $address_id = $response->json('data.id');
+        $this->assertGreaterThan(0, $address_id);
```
djaiss: If we want to be picky, it should be `addressId`, not `address_id` as it's not [PSR-2 compliant](https://www.php-fig.org/psr/psr-2/).

djaiss:
👍

tests/Api/ApiAdressesTest.php
asbiin: I agree with you: you're too picky ;)

PR #1284, opened 2018-05-06
[WIP] CalDAV / CardDAV

kidk:
WIP for #582

This is what I have after spending a couple of hours on it, the basic UI of Sabre works, but the two clients I've tested give me errors. 

- Evolution because of a misconfiguration in the client, so it sends an empty request size. [Ref](https://github.com/agendav/agendav/issues/191)
- IOS 12 still unknown. I've reverted to doing own authentication, instead of using the Laravel system, but that hasn't resolved the issue. Next up is looking at the [technical details](http://sabre.io/dav/clients/ios/).

There is still a lot of debug code in there, but I wanted to already share this to see if people have comments or requests. I've added a TODO below to give you an idea on the work that is still needed.

[] Adding more Contacts testing data
[] First working client (Evolution and IOS)
[] Read only CardDAV access 
[] Adding events and reminders to CalDAV
[] Read only CalDAV access

PR #1340, opened 2018-05-15
FakeContentTableSeeder now adds meaningfull data in ContactFields

kidk:
This was added for #1284 as I needed meaningful data to test with.


kidk:
Thanks @asbiin I now retrieve a random row from ContactFieldType and generate according to that.

What do you think about adding a fake phone number and fake email by default, but still randomly generating the others? This would match the real usage of the tool more I think.

asbiin:
@kidk "What do you think about adding a fake phone number and fake email by default, but still randomly generating the others? This would match the real usage of the tool more I think." Well it's interesting to have also cases of contacts without any ContactField as well. So your solution is elegant, I will merge it.

kidk:
Ok for me. Thanks 👍

PR #1341, opened 2018-05-15
Fixes #1338 Contact tags duplicate when filtering contacts. [Needs Review]

### LINKED ISSUE #1338: Contact tags duplicate when filtering contacts.
**Describe the bug**
Go to monicahq.com/people.
Click on a tag.
Click on the same tag again.
The tag is now duplicated in the filter instead of A) removed from the filter or B) nothing.
For example it says: "Showing all contacts tagged with: [- Red] [- Red]"
It appears that there is no limit to the number of duplicates and that using other tags has no effect on them.

Self Host V 2.1.1
Fresh install, performed a git clone and docker build within the last 2 hours.
I have not tried it against the public version.

turtles2:
This fixes #1338 by checking if we already have used that tag, unfortunately the duplicate tags need to stay in the URL unless we want to add more logic to the index method but I believe having a less than pretty URL is better then having a more complicated method. 

@djaiss When ever got get a chance to review.

### DIFF
--- a/app/Http/Controllers/ContactsController.php
+++ b/app/Http/Controllers/ContactsController.php
@@ -49,7 +49,9 @@ class ContactsController extends Controller
                             ->where('account_id', auth()->user()->account_id)
                             ->get();
 
-                $tags = $tags->concat($tag);
+                if(!($tags->contains($tag[0]))) {
+                    $tags = $tags->concat($tag);
+                }
 
                 $url = $url.'tag'.$count.'='.$tag[0]->name_slug.'&';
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/ContactsController.php
```diff
@@ -49,7 +49,9 @@ public function index(Request $request)
                             ->where('account_id', auth()->user()->account_id)
                             ->get();
 
-                $tags = $tags->concat($tag);
+                if (! ($tags->contains($tag[0]))) {
+                    $tags = $tags->concat($tag);
+                }
 
                 $url = $url.'tag'.$count.'='.$tag[0]->name_slug.'&';
```
asbiin: Should this line be in the `if` too ?

djaiss:
Thank you @turtles2

PR #1361, opened 2018-05-21
Remove url form text to translate

asbiin:
These url should not be in the text to translate.
Also added a hreflang to inform the user of the language of the destination text.


djaiss:
Oh good one @asbiin !

PR #1372, opened 2018-05-21
Replace queries with hardcoded "monica" database name to use the current default connection database

### LINKED ISSUE #1213: SQL export assumes a "monica" database

tomzx:
Fixes #1213


djaiss:
Nice work @tomzx !!

app/Helpers/DBHelper.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Illuminate\Support\Facades\DB;
+
+class DBHelper
+{
+    public static function getTables()
+    {
+        return DB::table('information_schema.tables')
+            ->select('table_name')
+            ->where('table_schema', '=', static::getDefaultConnectionConfiguration('database'))
```
asbiin: You can do: `DB::connection()->getDatabaseName()` to get the database name.

tomzx: :+1:

asbiin:
@tomzx have you seen my comment about `array_filter()` function ?

tomzx:
I saw it, but I forgot apply a fix for it. I'll do that ASAP.

tomzx:
@asbiin After reviewing the code, I suggest we remove the filtering altogether...

https://github.com/monicahq/monica/blob/0468845208280bc49952be5bf43cd9bc17146e55/app/Jobs/ExportAccountAsSQL.php#L135-L139

So it's doing a bunch of logic just to basically get the "accounts" string into `$tableName`...

asbiin:
Thanks @tomzx !

PR #1373, opened 2018-05-21
Add debts on the dashboard

djaiss:
This PR adds support for displaying debts on the dashboard, regardless of the contacts, in order to have a complete overview.

image

#### Code-related tasks
- [x] Impact on the seeders.
- [x] Impact on the API.

- [x] Impact on account export.
- [x] Impact on account reset and deletion.

#### Other tasks


resources/assets/js/app.js
```diff
@@ -174,6 +174,14 @@ import axios from 'axios';
 import VueI18n from 'vue-i18n';
 Vue.use(VueI18n);
 
+// Moments
+import moment from 'moment';
+Vue.filter('formatDate', function(value) {
+    if (value) {
+        return moment(String(value)).format('LL')
```
asbiin: Great, but it must support locale. Just add `moment.locale(lang);` on line 210 after `loadLanguageAsync(window.Laravel.locale, true).then((lang) =>`

djaiss: Good point. I was lazy. Thanks.

djaiss: I've added it and tested it, it seems to work. Can you double check?

djaiss: @asbiin waiting for your validation here!

asbiin: Seems good!

asbiin:
What about this ?
"Perhaps it should be more usefull to see the total for 1 person instead of all details?"

djaiss:
> "Perhaps it should be more usefull to see the total for 1 person instead of all details?"

Yes, but it's also more work :-D But yeah, it would be interesting. But we can ship this as is, then improve it in another PR.

PR #1374, opened 2018-05-21
Add email confirmation on signup

### LINKED ISSUE #1183: Add an option to confirm email address on signup
Europe is launching a new law in May, called GDPR, which is great, but comes with a lot of constraints for anyone having a webservice.

One of the thing we need to do to be compliant is:

_With e-mail double opt-in, a new user first needs to verify their e-mail address before completing the sign-up process._

I've tried to prevent doing this on Monica since the beginning - not because I wanted to collect your email address and do nasty thing with it, but because I wanted people to sign up with fake email addresses if they wanted to.

This law will force us to confirm email addresses on signup.

We will create a new `.env` variable, called `SIGNUP_DOUBLE_OPTIN`, which will default to `TRUE`. On our hosted version, it will be set to true, but you will do whatever you want on your own instance.

Note: to be clear, we do have an email newsletter, but we only contact people who have subscribed manually and specifically to receive this newsletter. The email address used on signup is never used otherwise, other than sending reminder emails.

Changes to do:
- [ ] Create a email confirmation process
- [ ] Keep track of when the user has validated his email address
- [ ] Display this information in the Settings panel under a new section (called `Compliance`)
- [ ] Add a method in the User object, something like `isCompliant`, that will tell if users have confirmed emails and accepted Terms & Privacy. If not, we should block email sending (reminders, notifications,...) automatically. This will be done in https://github.com/monicahq/monica/issues/1184 so you don't have to do it here.

cc @asbiin @Mokto

asbiin:
This will close #1183 

#### Code-related tasks

- [X] Impact on account export.

#### Other tasks
- [X] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG) entry added, if necessary, under `UNRELEASED`.


.env.example
```diff
@@ -59,6 +59,8 @@ APP_DEFAULT_LOCALE=en
 # Ability to disable signups on your instance.
 # Can be true or false. Default to false.
 APP_DISABLE_SIGNUP=true
+# Enable user email verification.
```
djaiss: It's more elegant if you put an empty line above it.

app/Http/Requests/EmailChangeRequest.php
```diff
@@ -0,0 +1,30 @@
...
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'email' => 'required|string',
+            'password' => 'required|string',
+            'newmail' => 'required|email|max:2083|unique:users,email,'.$this->id,
```
djaiss: Why `this->id` at the end?

app/Jobs/SendNewUserAlert.php
```diff
@@ -33,6 +33,10 @@ public function __construct(User $user)
      */
     public function handle()
     {
-        Mail::to(config('monica.email_new_user_notification'))->send(new NewUserAlert($this->user));
+        $email = config('monica.email_new_user_notification');
```
djaiss: 👍

app/Notifications/ConfirmEmail.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        return (new MailMessage)
+            ->subject(__('confirmation::confirmation.email-title'))
```
djaiss: What is `__('confirmation...` ?

Translations?

config/monica.php
```diff
@@ -12,6 +12,16 @@
     */
     'disable_signup' => env('APP_DISABLE_SIGNUP', false),
 
+   /*
+    |--------------------------------------------------------------------------
+    | Disable User registration
```
djaiss: Isn't it `Activate double optin on signup` instead?

app/Providers/RouteServiceProvider.php
```diff
@@ -19,19 +19,25 @@
...
+
     /**
      * This namespace is applied to your controller routes.
      *
      * In addition, it is set as the URL generator's root namespace.
      *
      * @var string
      */
-    protected $namespace = 'App\Http\Controllers';
+    protected function getNamespace()
```
djaiss: I don't understand this part of having to change namespace. Can you explain?

.env.example
asbiin: OK

app/Http/Requests/EmailChangeRequest.php
asbiin: Indeed, I copied it from https://github.com/monicahq/monica/blob/master/app/Http/Requests/SettingsRequest.php#L29

app/Jobs/SendNewUserAlert.php
asbiin: 👍

app/Notifications/ConfirmEmail.php
asbiin: Yes. This file is a stub copied from the new package.
I guess `__()` is an equivalent of `trans()`. The prefix 'confirmation::' is mandatory, it gets the translation from the package itself.

config/monica.php
asbiin: OK

app/Providers/RouteServiceProvider.php
asbiin: It uses the trait `DetectsApplicationNamespace` from Laravel to get the root namespace, which I think is nice.

asbiin:
image

app/Http/Controllers/Auth/EmailChangeController.php
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Display a listing of the resource.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function index(Request $request)
+    {
+        if ($request->session()->has('user_id')) {
+            $model = config('auth.providers.users.model');
```
djaiss: I understand why you've done this, but in our case, I don't think we'll ever change the default `users.model` in an installation... so why do this?

config/app.php
```diff
@@ -2,6 +2,19 @@
...
+    | Application Name
+    |--------------------------------------------------------------------------
+    |
+    | This value is the name of your application. This value is used when the
+    | framework needs to place the application's name in a notification or
+    | any other location as required by the application or its packages.
+    |
+    */
+
+    'name' => env('APP_NAME', 'Monica'),
```
djaiss: Holy cow. How come we didn't have this variable already?

resources/lang/en/auth.php
```diff
@@ -59,4 +59,10 @@
     'register_policy' => 'Signing up signifies you’ve read and agree to our <a href=":url" hreflang=":hreflang">Privacy Policy</a> and <a href=":urlterm" hreflang=":hreflang">Terms of use</a>.',
     'register_invitation_email' => 'For security purposes, please indicate the email of the person who’ve invited you to join this account. This information is provided in the invitation email.',
 
+    'confirmation_again' => 'If you want to change your email you can <a href="'.url('settings/emailchange').'" class="alert-link">click here</a>.',
+    'email_change_current_email' => 'Current email address:',
+    'email_change_title' => 'Change your email',
+    'email_change_new' => 'New email address',
+    'email_change_password' => 'Confirm your password',
+    'email_changed' => 'Your email has been changed. Check your mailbox to validate it.',
```
djaiss: `Your email address`

resources/lang/en/auth.php
```diff
@@ -59,4 +59,10 @@
     'register_policy' => 'Signing up signifies you’ve read and agree to our <a href=":url" hreflang=":hreflang">Privacy Policy</a> and <a href=":urlterm" hreflang=":hreflang">Terms of use</a>.',
     'register_invitation_email' => 'For security purposes, please indicate the email of the person who’ve invited you to join this account. This information is provided in the invitation email.',
 
+    'confirmation_again' => 'If you want to change your email you can <a href="'.url('settings/emailchange').'" class="alert-link">click here</a>.',
```
djaiss: `If you want to change your email address`

resources/lang/vendor/confirmation/en/confirmation.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+return [
+    'email-title' => 'Monica – Email verification',
+    'email-intro'=> 'To validate your email click on the button below',
+    'email-button' => 'Email verification',
```
djaiss: If it's a button it should be `Verify email address` or something.

app/Http/Controllers/Auth/EmailChangeController.php
asbiin: Indeed. It's a copy/paste

config/app.php
asbiin: Good question ...

asbiin:
2fa problems are fixed. Everything should work now.
@djaiss can you review this a last time ?

PR #1353, opened 2018-05-20
Typehead

Dagolin:
Add `typeahead` component to meet autocomplete requirements, the PR includes.

* add [vue2-bootstrap4-typeahead](https://github.com/r0bdiabl0/vue2-bootstrap4-typeahead) 
* replace `select` with `typeahead` on `add a new relation` page

Adding a relationship from existing contact could be frustrating if one has many contacts, image a dropdown with 100+ items, to solve the issue, my idea is to replace the dropdown box with  `typeahead` component. 

a

ref #1058

PR #1379, opened 2018-05-22
Add Vue-select component

Dagolin:
Ref : #1353, use  [vue-select](https://github.com/sagalbot/vue-select) instead of [vue2-bootstrap4-typeahead](https://github.com/r0bdiabl0/vue2-bootstrap4-typeahead).

### Warning, vue-select does not support HTML `required` attribute for now (https://github.com/sagalbot/vue-select/issues/477)

I came out with some idea but none of them is good enough to go

* Include another tool,  like [monterail/vuelidate](https://github.com/monterail/vuelidate), sounds like overkill
* Wrap whole form as a new vue-component and check `submit` event, which breaks the consistent of code

Any suggestion?

#### Preview

a

#### Submit with no selected data

b


resources/assets/js/components/people/ContactSelect.vue
```diff
@@ -0,0 +1,77 @@
...
+        methods: {
+            search(keyword, loading) {
+                this.getContacts(keyword, loading, this);
+            },
+            getContacts: function (keyword, loading, vm) {
+                axios.post(this.src, {
+                    needle: keyword,
+                    accountId: $('body').attr("data-account-id")
+                },{
+                    headers: {
```
djaiss: You don't need to add the headers when you are making your calls. They are added automatically, see https://github.com/monicahq/monica/blob/master/resources/assets/js/bootstrap.js#L29

resources/assets/js/components/people/ContactSelect.vue
```diff
@@ -0,0 +1,77 @@
...
+            userContactId: {
+                type: String
+            }
+        },
+        data () {
+            return {
+                src : '/people/search',
+                results : [],
+                filterable : false,
+                selected: { label: 'Search and select an existing contact', id: '' },
```
djaiss: You can use the translation service here instead of hardcoding text.

Dagolin:
Interesting idea, like having two different behavior in the dropdown box? 

* Populate default contact list and use it to perform searching , when < 100, no ajax
* Ajax searching when > 100, no default contact list

Dagolin:
Oh I guess I got you wrong, you are saying to have some default data for dropdown, so the user will be able to select what he wants without doing a search.

Agree, if I only have few contacts, that would be pretty annoying to perform the search every single time.

djaiss:
@Dagolin the idea is to always display the two possibilities:

- let user search if he wants to
- let him see a reasonable amount of contacts when the page loads, inside this dropdown.

That way, it's immediately clear what the search will do and how it will behave, and it won't be weird for users to have a dropdown without actual contacts on page load.

djaiss:
@Dagolin let me know when you think you are ready! 😀

Dagolin:
@djaiss Ready to review, Sir! Please let me know if there is anything I can make it better.

app/Http/Controllers/Contacts/RelationshipsController.php
```diff
@@ -44,7 +45,7 @@ public function new(Request $request, Contact $contact)
         foreach (auth()->user()->account->relationshipTypes as $relationshipType) {
             $arrayRelationshipTypes->push([
                 'id' => $relationshipType->id,
-                'name' => $relationshipType->getLocalizedName($contact, true),
+                'label' => $relationshipType->getLocalizedName($contact, true),
```
djaiss: This broke the relationship dropdown 😀

image

djaiss:
Awesome work 👍 
Can you also run `yarn run production` and commit the results? Then I'll merge it!

Dagolin:
A million thanks @djaiss 🙇

djaiss:
@Dagolin somehow CircleCI had an error that is not linked to your code. I can't merge it until the tests pass :-/ Can you retrigger the tests? I would do it if I could but I don't have rights on your repository.

djaiss:
@asbiin it keeps failing at this stage of the CI process. Do you know why?

djaiss:
Thanks so much, @Dagolin for your work on this. Sorry about the huge time it took to actually merge it. But we've done it ❤️

PR #1394, opened 2018-05-26
Fix for utf8mb4 and mysql < 5.7.7

asbiin:
Readd the `Schema::defaultStringLength(191);` if mysql version < 5.7.7


tomzx:
@asbiin Where does the magic number 191 comes from?

PR #1149, opened 2018-04-11
Remove map files from assets

asbiin:
Close #1120

PR #1407, opened 2018-05-30
Remove map files from assets (again)

asbiin:
This was already done in #1149


djaiss:
Hum perhaps it's me, I don't know. If it is, I'll be careful from now on @asbiin

PR #1414, opened 2018-05-31
Fix wrong url to send the confirm your email message

djaiss:


asbiin:
Indeed. Well, I prefer not to put an url in the translations. I will fix this

PR #1460, opened 2018-06-06
[WIP] POC : couchdb

Mokto:
The idea behind this proof of concept is to enable offline first (with monica's mobile app at start) management of contacts.


app/Console/Commands/MigrateCouch.php
```diff
@@ -0,0 +1,129 @@
...
+        $progress = new ProgressBar($output, count($accounts));
+        $progress->start();
+        foreach ($accounts as $account) {
+            // create its database
+            $client = CouchDbHelper::getAccountDatabase($account->id);
+            $client->createDatabase();
+            CouchDbHelper::admin(CouchDbHelper::getAccountDatabaseName($account->id))->addDatabaseMemberRole(CouchDbHelper::getAccountDatabaseRoleName($account->id));
+
+            //setting note design documents
+            $favNotesFn = "function(doc) { if (doc.type === 'note' && doc.is_favorited === true) { emit(doc.created_at, null); } }";
```
djaiss: Will we have to do this kind of things often? If so we would need to create methods for this in a class.

app/Console/Commands/MigrateCouch.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->info('');
+        $this->info('');
+
+        $notes = DB::table('notes')->get();
+        $this->info('✓ Migrating '.count($notes).' notes');
+        $output = new ConsoleOutput();
+        $progress = new ProgressBar($output, count($notes));
+        $progress->start();
+        foreach ($notes as $note) {
+            $client = CouchDbHelper::getAccountDatabase($note->account_id);
```
djaiss: Is `$client` the right name for this variable?

app/Console/Commands/MigrateCouch.php
```diff
@@ -0,0 +1,129 @@
...
+
+            $new_doc = (object) [
```
djaiss: In a real programming language like PHP, we use camelcase.

app/Console/Commands/MigrateCouch.php
```diff
@@ -0,0 +1,129 @@
...
+                '_id' => 'note-legacy-'.$note->id,
+                'type' => 'note',
+                'contact_id' => $note->contact_id,
+                'created_at' => $note->created_at,
```
djaiss: In theory `$note->created_at` returns a Carbon instance. Does it store a string here or the Object?

app/Helpers/CouchDbHelper.php
```diff
@@ -0,0 +1,38 @@
...
+
+use PHPOnCouch\CouchAdmin;
+use PHPOnCouch\CouchClient;
+
+class CouchDbHelper
+{
+    public static function client($dbName)
+    {
+        $config = config('database.connections.couchdb');
+        // dd($config);
```
djaiss: Delete

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -47,12 +48,20 @@ public function login(Request $request)
...
-            ]));
+            ]);
+            $userId = User::getUserIdFromAccessToken($proxy['access_token']);
+            $user = new CouchUser((array) CouchUser::getOneById($userId));
+            $proxy['user'] = $user->toArray();
+            $proxy['couchdbAddress'] = config('database.connections.couchdb')['frontAddress'];
+
+            return response()->json($proxy);
         } catch (Exception $e) {
+            dd($e);
```
djaiss: Remove?

app/Http/Controllers/Contacts/NotesController.php
```diff
@@ -15,13 +17,17 @@ class NotesController extends Controller
...
+        $notes = $response->rows;
 
-        foreach ($notes as $note) {
+        foreach ($notes as $noteArray) {
+            $note = new CouchNote((array) $noteArray->doc);
             $data = [
-                'id' => $note->id,
-                'parsed_body' => $note->parsedbody,
+                '_id' => $note->_id,
+                'parsed_body' => $note->getParsedBodyAttribute(),
```
djaiss: Laravel uses this weird stuff for accessors. That means you can access it with `$note->parsed_body`, no need to call the method directly.

app/Http/Controllers/Contacts/NotesController.php
```diff
@@ -15,13 +17,17 @@ class NotesController extends Controller
      */
     public function get(Contact $contact)
     {
+        $client = CouchDbHelper::getAccountDatabase($contact->account_id);
+
         $notesCollection = collect([]);
-        $notes = $contact->notes()->latest()->get();
+        $response = $client->limit(7)->startkey([$contact->id, []])->endkey([$contact->id])->descending(true)->include_docs(true)->getView('notes', 'byContact');
```
djaiss: Could we use a method for this? It looks so complicated, and if we have to enter it again and again...

app/Http/Controllers/Contacts/NotesController.php
```diff
@@ -41,65 +47,65 @@ public function get(Contact $contact)
...
      * @return \Illuminate\Http\Response
      */
-    public function destroy(Contact $contact, Note $note)
+    public function destroy(Contact $contact, CouchNote $note)
     {
         $note->delete();
 
-        $contact->events()->forObject($note)->get()->each->delete();
+        // $contact->events()->forObject($note)->get()->each->delete();
+        $contact->logEvent('note', $note->_id, 'delete'); // is this what I should do here ?
```
djaiss: We don't log deleted objects for now.

app/Http/Controllers/DashboardController.php
```diff
@@ -101,27 +104,45 @@ public function calls()
      */
     public function notes()
     {
+        $client = CouchDbHelper::getAccountDatabase(auth()->user()->account->id);
```
djaiss: Is this the right name for the variable?

app/Http/Controllers/DashboardController.php
```diff
@@ -101,27 +104,45 @@ public function calls()
...
+
         $notesCollection = collect([]);
-        $notes = auth()->user()->account->notes()->favorited()->get();
+        $response = $client->descending(true)->include_docs(true)->getView('notes', 'favorites');
```
djaiss: Could the `$client` come from the Account object instead? So you would call it with `auth()->user()->account->client` or something, instead of having to request the client from the helper every time?

I wonder if we could create a method in the $client object to get the objects we want from the DB, like `$client->get('notes', 'favorites);`... but would we cover all the cases we need... I don't know.

app/Models/CouchNote.php
```diff
@@ -0,0 +1,96 @@
...
+
+        return $client->deleteDoc($this);
+    }
+
+    /**
+     * Get a note from his _id.
+     *
+     * @return CouchNote
+     */
+    public static function getOneById($accountId, $noteId)
```
djaiss: We'll probably need to do this for every class.

Perhaps we could put this method in the parent?

app/User.php
```diff
@@ -384,4 +384,32 @@ public function getAllCompliances()
 
         return $terms;
     }
+
+    /**
+     * Get the userId from the access_token.
+     *
+     * @return string
+     */
+    public static function getUserIdFromAccessToken($access_token)
```
djaiss: 👍

app/Console/Commands/MigrateCouch.php
Mokto: This function is stored when a user is created and for migrations. You're right, I'll refactor this.

app/Console/Commands/MigrateCouch.php
Mokto: See http://php-on-couch.readthedocs.io/en/latest/api/couchclient/database.html#getting-started

But yeah, I guess we could name it `accountDatabase`.

app/Console/Commands/MigrateCouch.php
Mokto: It stores a string here. Do we need to store timezone here too ? Can't we use GMT timezone everywhere ?

app/Http/Controllers/Contacts/NotesController.php
Mokto: Right.

app/Http/Controllers/Contacts/NotesController.php
Mokto: That's what I thought but as the note is not a laravel model, but is inherited from `Jenssegers\Model\Model`, it does not work.

app/Http/Controllers/DashboardController.php
Mokto: 1) `auth()->user()->account->client` is a great idea.

2) I'll think about this.

app/Models/CouchNote.php
Mokto: Right ! I'll create a super class.

djaiss:
@Mokto I think it's safe to say we won't finish this one, right ;-) I'll close the PR and we'll reopen it needed.

Issue #1437, opened 2018-06-02
SQL error when adding a new person (Monica 2.2.1, PostgreSQL)


catatonik:
**Describe the bug**
The only options I have when adding a new person are: First name, Surname, Gender. If I populate all 3 I get the below SQL error:

```
(3/3) QueryException
SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "default_avatar_color" violates not-null constraint
DETAIL: Failing row contains (8, 4, Firstname, null, Surname, null, null, null, null, null, 2018-06-02 13:29:07, 2018-06-02 13:29:07, null, null, null, null, null, null, local, f, null, f, f, f, null, null, null, null, null, 1, null, null). (SQL: insert into "contacts" ("account_id", "gender_id", "first_name", "last_name", "updated_at", "created_at") values (4, 1, Firstname, Surname, 2018-06-02 13:29:07, 2018-06-02 13:29:07) returning "id")

(2/3) PDOException
SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "default_avatar_color" violates not-null constraint
DETAIL: Failing row contains (8, 4, Firstname, null, Surname, null, null, null, null, null, 2018-06-02 13:29:07, 2018-06-02 13:29:07, null, null, null, null, null, null, local, f, null, f, f, f, null, null, null, null, null, 1, null, null).

(1/3) PDOException
SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "default_avatar_color" violates not-null constraint
DETAIL: Failing row contains (8, 4, Firstname, null, Surname, null, null, null, null, null, 2018-06-02 13:29:07, 2018-06-02 13:29:07, null, null, null, null, null, null, local, f, null, f, f, f, null, null, null, null, null, 1, null, null).
```

**Which version are you using:**
 - A server you maintain yourself
-- Monica Version: 2.2.1
-- PHP 7.2.5-1
-- PostgreSQL 10.4
-- Debian 9.4

**Additional context**
More of the error from debug:

```
[...]

PR #1464, opened 2018-06-07
Set the default_avatar_color property before saving a contact model.

tomzx:
Related to #1437.

This should make sure that we set a default_avatar_color before we call `$contact->save()`, which should hopefully fix issues with database constraints where `default_avatar_color` cannot be `NULL`.

This removes the call to `$this->save()` in `\App\Contact::setAvatarColor()` and moves calls to `\App\Contact::setAvatarColor()` before the first call to `\App\Contact::save()` in various methods.


asbiin:
Thanks @tomzx !

PR #1470, opened 2018-06-09
Remove delay on tooltips

### LINKED ISSUE #1457: Avatar tooltip feels laggy
When you hover over avatars on the Dashboard and Journal page, the tooltip remains visible for too long which makes the tooltips overlay on top of each other. It also looks like the tooltips for avatars with photos aren't displaying correctly.

tomzx:
Fixes #1457.

By default, the delay on all tooltips will be 0, which means that as soon as you move your mouse away, the tooltip will disappear.


djaiss:
🕺thanks a lot!!

PR #1471, opened 2018-06-09
Set the contacts.last_name column to NULL if the last name sent to the Contacts controller is empty

### LINKED ISSUE #1441: Clearing the last name on Edit contact form doesn't clear the name
**Describe the bug**
* Edit a contact who has a first and last names
* Clear the last name field
* Save
* The profile keeps displaying the last name

tomzx:
Fixes #1441.


tests/Feature/ContactTest.php
```diff
@@ -212,6 +212,26 @@ public function test_a_contact_can_have_food_preferences()
...
+        list($user, $contact) = $this->fetchUser();
+
+        $data = [
+            'firstname' => $contact->first_name,
+            'lastname' => '',
+            'gender' => $contact->gender_id,
+            'birthdate' => 'unknown',
+        ];
+
+        $this->post('/people/'.$contact->id.'/update', $data);
```
asbiin: I think it should use `hashID` instead of `->id`

tomzx: Should I take the opportunity to update all other code (within the ContactTest class) so that it stays consistent?

asbiin: @tomzx Oh I didn't saw it's in a test. Well, I think it's not a bad idea to fix these tests, go ahead !

asbiin: @tomzx we will fix the contactid uses in test later...

PR #1476, opened 2018-06-11
Move models into their respective folders

djaiss:
The idea is to clean up the structure of the app a little bit. It'll look less messy.


djaiss:
@asbiin ready for review but.... almost impossible to review.

djaiss:
I need to make sure this hasn't broken registration.

PR #1491, opened 2018-06-16
Dates are now stored in UTC

djaiss:


djaiss:
@asbiin could you review this?

app/Http/Controllers/ContactsController.php
```diff
@@ -160,7 +160,7 @@ public function show(Contact $contact)
             $query->orderBy('updated_at', 'desc');
         }]);
 
-        $contact->last_consulted_at = Carbon::now(auth()->user()->timezone);
+        $contact->last_consulted_at = Carbon::now();
```
asbiin: You could use the `ǹow()` helper

app/Models/Contact/Contact.php
```diff
@@ -1509,9 +1509,9 @@ public function updateStayInTouchFrequency($frequency)
      * @param int $frequency
      * @param string $timezone
      */
-    public function setStayInTouchTriggerDate($frequency, $timezone)
+    public function setStayInTouchTriggerDate($frequency)
     {
-        $now = Carbon::now($timezone);
+        $now = Carbon::now();
```
asbiin: You could use the `ǹow()` helper

database/migrations/2018_06_14_212502_change_default_name_order_user_table.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function ($table) {
+            $table->string('name_order')->default('firstname_lastname_nickname')->change();
```
asbiin: This has nothing to do with the UTC timezone, right ?

djaiss: No indeed. It's a bug fix for something else. I should put it out and create a separate PR for this.

asbiin: OK I have moved it in #1585

app/Http/Controllers/ContactsController.php
djaiss: Done.

app/Models/Contact/Contact.php
djaiss: Done.

PR #1532, opened 2018-07-02
comment fix: disable signup default is true

privong:
#### Before submitting the PR


asbiin:
Nope ! It's false by default:
https://github.com/monicahq/monica/blob/master/config/monica.php#L13
... but the value in .env.example is true.
@djaiss Maybe we should change the value in .env.example?

privong:
@asbiin i see. then the example file should be clarified, perhaps as something like: "Is false by default but here is overridden as true"

PR #1548, opened 2018-07-07
fix(docker-compose): mount top-level storage dir

tlvince:
The oAuth keys are stored in `/var/www/monica/storage`, so will not be persisted if only the public directory is bound as before. See https://github.com/monicahq/chandler/issues/239


asbiin:
Thank you @tlvince !

PR #1554, opened 2018-07-08
Update timestamps timezone

asbiin:
Since #1527 timestamps are stored in UTC, but older timestamps has to be migrated.


djaiss:
@asbiin I understand what you are trying to do here. I was wondering whether we should do it.
Let me explain.

Let's say the app was in 'US/Eastern' before and the user entered a birthdate `01/01/1980`.

The birthday should always remain `01/01/1980` regardless of the timezone. But if we do what you want to do here, depending on the timezone, we might end up having `02/01/1980` for instance, which we don't want.

This means... I'm not sure this is the right idea here. Perhaps we shouldn't convert each date we stored so far.

djaiss:
Dismiss my comment. You are only affecting `created_at` and `updated_at` entries. My bad.

PR #1556, opened 2018-07-09
First step for activities customization [wip]

djaiss:
TBD


djaiss:
@asbiin I don't understand why those tests fail. If I make them work locally, it doesn't work on circleCI. It seems we don't have the same configuration while testing locally and on CircleCI.

database/migrations/2018_07_03_204220_create_default_activity_type_groups_table.php
```diff
@@ -0,0 +1,268 @@
...
+            $table->integer('default_activity_type_category_id');
+            $table->string('translation_key');
+            $table->string('location_type');
+            $table->timestamps();
+        });
+
+        // SIMPLE ACTIVITIES
+        DB::table('default_activity_type_categories')->insert([
+            'translation_key' => 'simple_activities',
+            'created_at' => now(),
```
asbiin: Like in old `ActivityTypesTableSeeder` class, you should calculate `now()` only once.

resources/views/people/show.blade.php
```diff
@@ -0,0 +1,66 @@
...
+      <div class="{{ Auth::user()->getFluidLayout() }}">
+
+        <div class="row">
+          <div class="col-xs-12 col-sm-3 profile-sidebar">
+
+
+            <p><a href="{{ url('/people/'.$contact->hashID().'/vcard') }}">{{ trans('people.people_export') }}</a></p>
+            <p>
+              {{ trans('people.people_delete_message') }}
+              <a href="#" onclick="if (confirm('{{ trans('people.people_delete_confirmation') }}')) { $('#contact-delete-form').submit(); } return false;">{{ trans('people.people_delete_click_here') }}</a>.
```
asbiin: `$('#contact-delete-form').submit();` I though you wanted to get rid of jQuery ?

resources/views/people/show.blade.php
```diff
@@ -0,0 +1,66 @@
+@extends('layouts.skeleton')
```
asbiin: I think this file is not used

asbiin:
Just 2 fixes for this PR:
[2018-07-03-new-profile.txt](https://github.com/monicahq/monica/files/2182315/2018-07-03-new-profile.txt)

asbiin:
> @asbiin I don't understand why those tests fail. If I make them work locally, it doesn't work on circleCI. It seems we don't have the same configuration while testing locally and on CircleCI.
Yes we have the same configuration, you should not have modified the test.

resources/views/people/show.blade.php
djaiss: I actually didn't change this. Yes we need to get rid of jQuery but perhaps it's not the time here. I'll check if I can change this though.

resources/views/people/show.blade.php
djaiss: You are actually right.

djaiss:
@asbiin Ready to merge

resources/views/people/show.blade.php
asbiin: Always ;)

djaiss:
We shouldn't merge yet, I forgot something, hang on.

PR #1574, opened 2018-07-15
Format date in input (vuejs) with right locale format

asbiin:
Use right format to display to date (only vuejs date in `form-date` component).

image
- date is right formatted
- if needs, monday is first day of the week

It uses `moment`.


djaiss:
👍 excellent work

Issue #839, opened 2018-01-21
Birthday of someone who's birthday is today is not shown on the dashboard


rickpastoor:
I have a contact who's birthday is today, but it is not shown, while others (further in the future) are shown. I think it is useful to see who's birthdays I actually have to do something about -- today ;)

tlwt:
Fully agree on that and I think this can be merged with #740

djaiss:
Feedback from user:

One thing i've noticed is that my dashboard only shows upcoming reminders and not current ones (i.e. reminders for today)

djaiss:
I have hard time reproducing this bug. The thing is, the dashboard lists all reminders in the current month (regardless of the day we are in the current month).

Do you have any other details?

rickpastoor:
Ping! I keep missing things because of this :(

djaiss:
Ok. I'm on it!

rickpastoor:
This seems to be an issue still, sorry for brining it up again :)

elanosilveira:
This seems to be an issue still, sorry for brining it up again :) [2]

hwmrocker:
This seems to be an issue still, sorry for bringing it up again :) [3]

CyberMuz:
This seems to be an issue still, sorry for bringing it up again :) [5]

mwolfaardt:
I have the same issue if I use the CalDAV API

JacTra:
Unfortunately this is still a (quite significant) issue - today's events are not shown on the dashboard. :(

aspoet:
**This is still a very significant issue.** Checking the dashboard on any given day to see who's birthday it is today is the single most important usecase for entering birthdays in the first place. Since monica has no calendar-view or a birthday list, this is crucial. Will this ever be fixed ?

Any information on a workaround would be greatly appreciated. This makes the software unusable for my specific usecase (not missing peoples birthdays by checking everyday).

Version: [4.1.2]

abinadi:
Anyone know if this will be fixed in v5?

PR #1632, opened 2018-07-30
Show reminders for the current day on the dashboard

### LINKED ISSUE #740: Dashboard should list reminders for today
Feedback:

One thing i've noticed is that my dashboard only shows upcoming reminders and not current ones (i.e. reminders for today)

cmorbitzer:
**Edit** see comment below for discussion point

Fixes #740 and #839 

This delays the SetNextReminderDate job by 48 hours so that reminders'
next_expected_date does not change until after the current day has
passed for users in all timezones. This also changes the condition in
getRemindersForMonth() on the Account model so that reminders for the
current month will be shown from the start of the current day instead
of the current time.

This should not interfere with any recurring reminders because the
minimum interval for a reminder is 1 week.

Users also should not see duplicate reminders because isTheRightTimeToBeReminded() does not return true for past reminders but only when a user is supposed to receive a reminder at that very hour.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 UNRELEASED CHANGES:
 
 * Add ability to define custom activity types and activity type categories through the API and the UI
+* Show reminders for the current date on the dashboard
 
 RELEASED VERSIONS:
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -34,3 +34,4 @@ Theo Mathieu @Mokto
 Daniel Pieper @danielpieper <monica@daniel-pieper.com>
 Andrew Paul Smith <github@chucklesthebeard.com>
 Brian Clemens <brian@tiuxo.com>
+Chris Morbitzer @cmorbitzer <chris.morbitzer@gmail.com>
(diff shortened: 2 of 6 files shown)


cmorbitzer:
After sleeping on this, I'm not enamored with this solution, because it means the next occurrence of a reminder won't appear until 2 days after the last one. Even though this won't interfere with reminders on the day of the reminder, the user wouldn't have as accurate a list of upcoming reminders.

The best solution I can see from here is to also select reminders where `last_triggered` is the current date in the user's timezone. The challenge is sorting the results, because a database-sorting solution would require some raw SQL that would be compatible with most if not all DBMS, but not guaranteed to be so. (something like `ORDER BY LEAST(COALESCE(last_triggered, '9999-12-31'), next_expected_date)`) Sorting the results on the server side with the Laravel collection helper methods wouldn't work so well because the ordering is used in the definition of the `Contact::reminders()` relation and by at least one API method.

I'm willing to implement this-- I've already started. Thoughts?

djaiss:
@cmorbitzer thanks for your work.

Reminders are a tricky beast. Just so you know, right now, for reasons I can't understand yet, some users do not even receive their reminders, as the cron job sometimes miss items in the queue (and I have no idea why). So everything that touches reminders makes me nervous.

That being said, is this the right solution at all? Perhaps we could use another new table for this purpose? Let's talk about this in an issue so the conversation doesn't get lost.

Check here https://github.com/monicahq/monica/issues/1638

cmorbitzer:
Sounds good. I'll close this for now.

PR #1640, opened 2018-08-01
fix(package): update tachyons to version 7.0.1-3

asbiin:
Closes #1567


asbiin:
Actually, this update is not a good idea, as a lot of things changed with our current version:
image
image
image

PR #1660, opened 2018-08-05
fix: Fix idHasher decode function

asbiin:


asbiin:
Sentry issue: [PRODUCTION-APP-6S](https://sentry.io/monica-lv/production-app/issues/630370449/)

PR #1662, opened 2018-08-06
Fix missing storage link

TheZenti:
Fixes #1622.

If migrations fail, the storage directory will not be linked to $MONICA_DIR/public/storage, which breaks uploaded avatars. This PR moves the linking command so migrations will be executed at the end.

#### Before submitting the PR
- [X] Read the [CONTRIBUTING document](https://github.com/monicahq/monica/blob/master/CONTRIBUTING.md) before submitting your PR.
- [X] If the PR is related to an issue or fix one, don't forget to indicate it.
- [X] Make sure that the change you propose is the smallest possible.
- [n/a] Screenshots are included if the PR changes the UI.
- [n/a] If you change the UI, make sure the user experience is consistent with the current interface.

#### Code-related tasks
- [n/a] Tests added for this feature/bug. - Storage linking was disabled in testing environments before.
- [n/a] Impact on the seeders.
- [n/a] Impact on the API.

#### Other tasks
- [X] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG) entry added, if necessary, under `UNRELEASED`.
- [X] [CONTRIBUTORS](https://github.com/monicahq/monica/blob/master/CONTRIBUTORS) entry added, if necessary.
- [n/a] If it's relevant and worth mentioning, create a changelog entry for this change. The changelog entry will appear inside the UI for all users to see. To know if your change is worth the creation of a changelog entry, [read the documentation](https://github.com/monicahq/monica/blob/master/docs/administrators/tips.md#when-is-it-relevant-to-create-a-changelog-entry).

PR #1663, opened 2018-08-06
Fix reminders for single digit hours (00:00-09:00)

TheZenti:
Saving a default reminder time using the web frontend saves that time in 24-hour format with leading zero. (At least the standard Docker deployment does.)
However, this time was compared to the current hour without leading zero. Therefore, all reminders between 00:00 and 09:00 were never sent on my deployment.

This PR changes the current hour formatting to 24-hour with leading zero, and adjusts the associated tests to consistently set the default hour with leading zero, which fixes the aforementioned problem.

Feel free to reject this PR if this problem isn't actually present on platforms other than Docker. Still, making all tests associated with reminders, notifications and such consistent would still be a sensible goal IMO.

#### Before submitting the PR
- [n/a] Screenshots are included if the PR changes the UI.
- [n/a] If you change the UI, make sure the user experience is consistent with the current interface.

#### Code-related tasks
- [X] Tests added for this feature/bug.
- [n/a] Impact on the seeders.
- [n/a] Impact on the API.

- [n/a] Impact on account export.
- [n/a] Impact on importing data with `vCard` and `.csv` files.
- [n/a] Impact on account reset and deletion.

#### Other tasks
- [X] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG) entry added, if necessary, under `UNRELEASED`.
- [X] [CONTRIBUTORS](https://github.com/monicahq/monica/blob/master/CONTRIBUTORS) entry added, if necessary. - already happened in different PR (#1662)


asbiin:
Thanks so much @TheZenti !
Actually, this could fix some not sent reminders ... 
I wrote [this script](https://github.com/monicahq/monica/blob/master/database/migrations/2018_05_24_160546_fix-inconsistant-reminder-time.php) to set every hours *with* leading zero. But I missed the fix on `isTheRightTimeToBeReminded` function. 

Cc @djaiss

djaiss:
@TheZenti you might have finally found a bug that I've been trying to find the cause for a few weeks now. Huge 👍 .

PR #1684, opened 2018-08-10
chore: enable turkish locale

djaiss:


asbiin:
@djaiss To enable a new language you must update `https://github.com/monicahq/monica/blob/master/config/lang-detector.php` file too.

djaiss:
@asbiin I didn't know. I'll update the doc.

PR #1688, opened 2018-08-12
fix(command): Cache configuration on Update in production

asbiin:


asbiin:
I will add some doc in installation/update process before finish this PR

PR #1689, opened 2018-08-12
feat: Improve automatic route binding

asbiin:
# Improvements
* `IdHasher::decodeId` now throws a `App\Exceptions\WrongIdException` Exception when `$hash` is not a real hash. The will block getting model (Contact, Activity, etc.) with Hashing, by id number directly.
* Some improvement were made in `ActivitiesController` to look for the list of contact only 1 time, instead of 2 times for all linked contacts (store and update function)
* `Hasher` trait is gone. Models now extends `ModelBindingHasher`.
* All models with previous special route in `RouteServiceProvider` now extends `ModelBinding` or `ModelBindingWithContact` which implements the resolveRouteBinding function.

# Fix
* A new binding for 'otherContact' was created
* Model binding are fixed for : ActivityType, ActivityTypeCategory, Address, Call
* 'relationships' route is not used
* in `ContactsController`, using now the right method call `route('people.show', $contact)`
* unused route '/settings/personalization/activitytypes/{activityTypeCategoryId}' was removed
* old `ID_hasherTest` class is now PSR4 compliant


djaiss:
This is a huge change 😱

asbiin:
@djaiss yes I know :D Introduce HashID is not very usefull if old ids still works.
Thus, I prefer to use `resolveRouteBinding` function instead of adding the same routine in `RouteServiceProvider` x times.

djaiss:
@asbiin I've fixed the failed Cypress tests. You can merge when you want! Thanks!

PR #1701, opened 2018-08-15
fix: Pluralize activities profile subtitle

asbiin:
Fix recent activities report.

- pluralize people.activities_profile_subtitle text
- fix people.activities_profile_number_occurences: in english, 0 times is "plural" form (well, it whould never display "0 activities" here, but voilà)
- Carbon::formatLocalized is not working, the LC locale must be installed in the system to work. We should never use it. I've created a new `DateHelper::getFullMonthAndDate` function to replace this.


app/Helpers/DateHelper.php
```diff
@@ -134,9 +134,24 @@ public static function getShortMonth($date)
...
     }
 
+    /**
+     * Return the month and year of the date according to the timezone of the user
+     * like "October 2010", or "March 2032".
+     *
+     * @param Carbon $date
+     * @return string
+     */
+    public static function getFullMonthAndDate($date)
```
djaiss: @asbiin where is the test for this 😀 ?

PR #1703, opened 2018-08-15
feat: Get first existing gravatar of all emails of the contact

asbiin:
Add ability to get gravatar, not for the first email, but for every emails: first gravatar detected is returned.

### DIFF
--- a/app/Models/Contact/Contact.php
+++ b/app/Models/Contact/Contact.php
@@ -994,7 +994,7 @@ class Contact extends Model
                 $query->where('type', '=', 'email');
             })
             ->get();
-        
+
         foreach ($emails as $email) {
             if (is_null($email) || empty($email->data)) {
                 continue;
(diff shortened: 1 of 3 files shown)


app/Models/Contact/Contact.php
```diff
@@ -982,31 +982,40 @@ public function updateGravatar()
...
-     * Get the gravatar, if it exits.
+     * Get the first gravatar of all emails, if found.
      *
      * @param  int $size
      * @return string|bool
      */
     public function getGravatar($size)
     {
-        $email = $this->getFirstEmail();
+        $emails = $this->contactFields()
```
djaiss: This is basically the content of `getFirstEmail` method.
This method is now not used anywhere in the app.

asbiin: Indeed. Do you wish to remove it?

djaiss: Yes, if it isn't used...

PR #1629, opened 2018-07-29
feat: Conversations

djaiss:
This PR adds the concept of **conversations**.

Right now we can log phone calls. It's useful, but it's limited. What if we could log every communication with a contact? Each communication is a conversation. A conversation has one or more messages. The idea is to let users record conversations made on social media (FB, Whatsapp,...), interactions through SMS, meeting notes, phone calls, email,...

All this presented in a nice and friendly way. That way, I can remember for instance that I've had a nice conversation with a good friend of mine at this date, and we've talked about X.

This will close https://github.com/monicahq/monica/issues/917

For this first iteration, **we won't migrate existing phone calls** and we'll keep the calls feature, otherwise it will break the mobile application

- [x] Order conversations list by date
- [x] Crop last message if it's too long for display
- [x] Make sure it works in responsive mode
- [x] When deleting a contact field type, make sure we delete conversations too
- [ ] Add Conversations on dashboard
- [x] Daily Statistics calculation on the instance
- [x] Cypress tests
- [x] Make sure all exceptions are taken into account in the frontend
- [x] Impact on notion of modules

image

## Technical discussion

This PR uses the notion of service, and especially the notion of [single action classes](https://medium.com/@remi_collin/keeping-your-laravel-applications-dry-with-single-action-classes-6a950ec54d1d). The idea is to have one class per action.
It's very useful as it will let us write less code at the end. By using a class separated from your model, we can have the logic at one place, and call this logic from both the API and the web version, without having to write the backend twice.

#### Code-related tasks
- [x] Tests added for this feature/bug.
- [x] Impact on the seeders.
- [ ] Impact on the API.

#### If the code changes the SQL schema
- [x] Impact on account export.
- [x] Impact on importing data with `vCard` and `.csv` files.
- [x] Impact on account reset and deletion.

#### Other tasks
- [x] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG) entry added, if necessary, under `UNRELEASED`.

PR #1747, opened 2018-08-27
feat: prepare AWS S3 upload for avatars

### LINKED ISSUE #1742: Store images in AWS S3
We plan to move images for app.monicahq.com in S3 - see #1735.

This issue is to check if everything is ready to store images in S3.
See https://laravel.com/docs/5.6/filesystem

- [x] create an [S3 bucket](https://s3.console.aws.amazon.com/s3/home) for Monica
- [x] create a new [strategy](https://console.aws.amazon.com/iam/home?#/policies) - see [s3strategy.txt](https://github.com/monicahq/monica/files/2342251/s3strategy.txt) : replace %BUCKET% with the name of the bucket
- [x] create a new [group](https://console.aws.amazon.com/iam/home?#/groups) and apply the strategy
- [x] create a new [user](https://console.aws.amazon.com/iam/home?#/users) in the newly group
- [x] update `.env` with `DEFAULT_FILESYSTEM=s3`, and AWS keys
- [x] test uploading an avatar in the new bucket 
- [x] put our instance down in production: `php artisan down`
- [x] run command `php artisan monica:moveavatars -vvv`
- [x] put our instance back up in production: `php artisan up`

asbiin:
Fix the feature of upload files to AWS S3 for avatars.
This is linked to #1742


app/Console/Commands/MoveAvatars.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
+
+class MoveAvatars extends Command
```
djaiss: This is something that should be done only once, right? Maybe we could indicate it in the description.

app/Console/Commands/MoveAvatars.php
```diff
@@ -0,0 +1,78 @@
...
+                }
+            });
+    }
+
+    private function moveAvatarSize($contact, $size = null)
+    {
+        $filename = pathinfo($contact->avatar_file_name, PATHINFO_FILENAME);
+        $extension = pathinfo($contact->avatar_file_name, PATHINFO_EXTENSION);
+
+        $avatar_file_name = 'avatars/'.$filename.'.'.$extension;
```
djaiss: `$avatar_file_name` -> `$avatarFileName` for PRS4 compliance.

app/Console/Kernel.php
```diff
@@ -19,6 +19,7 @@ class Kernel extends ConsoleKernel
         'App\Console\Commands\ImportCSV',
         'App\Console\Commands\ImportVCards',
         'App\Console\Commands\LangGenerate',
+        'App\Console\Commands\MoveAvatars',
```
djaiss: Perhaps we could put the commands that should only be done once if a proper folder, like `Commands\OneTime` or something.

app/Models/Contact/Contact.php
```diff
@@ -937,14 +938,50 @@ public function getAvatarURL($size = 110)
             return $this->avatar_external_url;
         }
 
-        $original_avatar_url = Storage::disk($this->avatar_location)->url($this->avatar_file_name);
+        $original_avatar_url = $this->avatar_file_name;
```
djaiss: PSR4

app/Models/Contact/Contact.php
```diff
@@ -937,14 +938,50 @@ public function getAvatarURL($size = 110)
...
         $avatar_extension = pathinfo($original_avatar_url, PATHINFO_EXTENSION);
         $resized_avatar = 'avatars/'.$avatar_filename.'_'.$size.'.'.$avatar_extension;
 
         return asset(Storage::disk($this->avatar_location)->url($resized_avatar));
     }
 
+    /**
+     * Delete avatars files.
+     */
+    public function deleteAvatars()
```
djaiss: I don't understand why we need to create a function in Contact object just for this, especially if this process will only happen once for this migration. Is it really useful? Can we move this function elsewhere ?

app/Console/Commands/MoveAvatars.php
asbiin: As you know, the location of an avatar is stored in ``Contact -> avatar_location`` property. The ``MoveAvatars`` command aims to move avatars (contacts with ``has_avatar`` to true) from current location to the - possibly different - default driver, as ``config('filesystems.default')``. This way, you can run the commands multiple times: if the current location of the avatar is the default one, it's not touched.

app/Models/Contact/Contact.php
asbiin: When a contact already has an avatar (not gravatar), if you update the image, today the old image stays in the store forever. I'm not sure it's a voluntary behavior. I've created this function to remove the old image, but we can imagine an other behavior...

asbiin: And today a user can flood us by uploading as much as images he can, and they will stay in the store...

app/Console/Commands/MoveAvatars.php
djaiss: But will we have to run the commands multiple times?

app/Models/Contact/Contact.php
djaiss: This is true. Thanks for the explanation.
I wonder if we should create a service for this method instead, like what I've been started to do here: https://github.com/monicahq/monica/pull/1629/files#diff-b3c40f0913df027e06af970085f889ed

The idea: have a class called, for instance, `deleteAvatarForContact`. This is [explained here](https://medium.com/@remi_collin/keeping-your-laravel-applications-dry-with-single-action-classes-6a950ec54d1d).

What do you think?

app/Console/Commands/MoveAvatars.php
asbiin: We will have to run the command one time for this migration, but it could be run multiple times, every time you update the default filesystem store.

app/Models/Contact/Contact.php
asbiin: After a bit reflexion: this does not apply here, even I think Services are great, but it's used for Controller, to handle request the same way for htpp and api. In this case, I already have the $contact object, and I just pass it to the Service: there is no interest.

djaiss:
:+1:

PR #1753, opened 2018-08-28
fix: fix search with wrong search field

asbiin:
Fix search with a search field that not exists


asbiin:
Sentry issue: [PRODUCTION-APP-71](https://sentry.io/monica-lv/production-app/issues/651879087/)

PR #1766, opened 2018-09-02
fix: fix email maximum size on settings

### LINKED ISSUE #1381: Email validation form differ
Validation of email field is not the same on register form and settings form:
[SettingsRequest.php#L29](https://github.com/monicahq/monica/blob/master/app/Http/Requests/SettingsRequest.php#L29)
`'email' => 'required|email|max:2083|unique:users,email,'.$this->id,`
[RegisterController.php#L72](https://github.com/monicahq/monica/blob/master/app/Http/Controllers/Auth/RegisterController.php#L72)
`'email' => 'required|email|max:255|unique:users',`

Wich one is right?

asbiin:
This fix #1381

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@ UNRELEASED CHANGES:
 * Fix search with wrong search field
 * Fix gift recipient relation
 * Fix subscription cancel on account deletion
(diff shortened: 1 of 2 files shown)


djaiss:
Good one.

PR #1767, opened 2018-09-02
fix: fix reminder link in email sent

asbiin:
Fix the link in reminder mails, currently the host appears twice:
image

PR #1768, opened 2018-09-02
feat: update mail template

### LINKED ISSUE #1339: Email reminders are very shouty
When a reminder email is received it looks something like:
```
Hi Matt,

YOU WANTED TO BE REMINDED OF:

Wish Joe happy birthday.

FOR:

Joe Bloggs
```
This comes across as very shouty where simply making these lower-case would help.

Looking at the source this is defined in `monica/public/js/langs/en.json`, `monica/resources/lang/en/mail.php` and `monica/resources/lang/es/mail.php`. I'm not sure why it is repeated in these files but changing the content of `want_reminded_of` to `'You wanted to be reminded of'` and `'for'` to `'for:'` would do the trick.

As far as I can tell, the rest of the messages are not full upper-case so this would perhaps make it more consistent too.

asbiin:
Example:
image

This fix #1339 and will close #1767


djaiss:
Excellent. It doesn't "fix" #1339 though - just partially. We should apply this to the reminder themselves, not only the stay in touch, if we want to close #1339.

asbiin:
@djaiss indeed. I've rewrite the other reminders, but we can certainly go further.

asbiin:
@djaiss can you review this please?

PR #1770, opened 2018-09-03
feat: add language selection in registration page

### LINKED ISSUE #1003: Add langage listbox in registration page

asbiin:
This close #1003
image

It adds another feature: you can switch to any supported language by adding "?lang=XX", where XX is the 2-letters code, anywhere in the application. This could be very useful on development!

This will close #1178 too


djaiss:
Let me just update the layout just a little bit before merging.

djaiss:
Updated

image

PR #1794, opened 2018-09-08
fix: do not send confirm email when signup_double_optin is false

asbiin:


app/Services/User/EmailChange.php
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @param array $data
+     */
+    public function execute(array $data)
+    {
+        if (! $this->validateDataStructure($data, $this->structure)) {
+            throw new MissingParameterException('Missing parameters');
+        }
+
+        $user = auth()->user();
```
djaiss: I don't think it's a good idea that you use `auth()->user()` in a service. This assumes that we call the service in the context where we are authenticated which might not be always the case.

Instead, you need to pass the user id as a parameter and then find the user, etc...

More DB queries but at least, the service is not dependant on something else.

asbiin:
> Also, don't forget the test for the service ;-)

Yes that was my intention ;)

asbiin:
@djaiss tests done!

PR #1797, opened 2018-09-08
feat: parse FN property when import vcards

### LINKED ISSUE #1796: vCard import doesn't support FN tag
My vCard export from Nextcloud uses the `FN` tag instead of the `N` tag therefore no contacts get imported. 

As the `FN` tag is required in the [new vCard specification](https://en.wikipedia.org/wiki/VCard#Properties) we should also parse this field.

I'll provide a PR which fixes this issue.

fabwu:
Fixes #1796


djaiss:
Thanks so much for your help @fabwu 
I'll review this PR in the coming days, sorry for the time it takes to do so.

djaiss:
@fabwu 
Thanks so much for your work.
if you could just fix the two issues reported by Sonar, we'd be good to go!

djaiss:
@asbiin here the job on Jenkins seems stuck. Do you see how to unstuck this?

PR #1826, opened 2018-09-19
fix: remove useless try catch

asbiin:
These try/catch are useless. Use the `MissingParameterException` instead of `\Exception` as well.


djaiss:
@asbiin much better this way. Thanks.

djaiss:
@asbiin I'll have to rewrite some of the methods in the life events feature, then, to match what you've done here.

PR #1842, opened 2018-09-25
fix: test if user is authenticated on error 500 page

asbiin:
Forgot to test that


asbiin:
Sentry issue: [PRODUCTION-APP-7X](https://sentry.io/monica-lv/production-app/issues/701010542/)

PR #1849, opened 2018-09-26
feat: add a display name for email notifications

asbiin:
The permit to not use APP_NAME to display an application name on email notifications


config/app.php
```diff
@@ -15,6 +15,17 @@
 
     'name' => env('APP_NAME', 'Monica'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Application Display Name
+    |--------------------------------------------------------------------------
+    |
+    | This value is the name of your application displayed in email notifications.
```
djaiss: `This is the name of the application that will be displayed in the notification emails`.

PR #1875, opened 2018-10-04
 fix: fix life event 'got a new vehicle' misspell

asbiin:


asbiin:
Need to rename the svg file too.

PR #1881, opened 2018-10-07
fix(i18n): improve text when adding contact, not just "significant other"

Flimm:
The contact could be another family member.


asbiin:
@djaiss it needs to run `php artisan lang:generate` and `yarn run production` before merge.

asbiin:
Thanks @Flimm !

Issue #1880, opened 2018-10-06
Table monica.default_life_event_types on creation of user


anmol26s:
On creation of first user I am getting this error : 
Something went wrong:
`
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'monica.default_life_event_types' doesn't exist (SQL: update `default_life_event_types` set `migrated` = 1)`

Further I have disabled `APP_SIGNUP_DOUBLE_OPTIN` but still I get email verification notice on the new user registration.

asbiin:
This might be related to #1882

anmol26s:
@asbiin 
Is there temporary sql fix for this, so that I can have new user?

djaiss:
2.8.1 is out

@anmol26s does it fix the issue?

anmol26s:
@djaiss 
Yes \o/ 
This push fixes the issue. Thanks for fast merge and release.

PR #1885, opened 2018-10-07
fix: fix use of 'json' column type

### LINKED ISSUE #1882: Type json not supported on mysql < 5.7
The new life event feature introduced in v2.9.0 (#1850) includes a [migration script](https://github.com/monicahq/monica/blob/master/database/migrations/2018_08_31_020908_create_life_events_table.php] which uses the `json` Eloquent function.
This type is only supported for [mysql 5.7.8](https://dev.mysql.com/doc/refman/5.7/en/json.html) and later.

With mysql 5.6, the script failed, but the first table is already created. So when we run the script a second time, the error message is displayed : `Base table or view already exists: 1050 Table 'default_life_event_categories'` (see #1865)

We have to replace the json type with a simple string.

asbiin:
Fix using the type 'json' for a column type, which is available on mysql > 5.7.8 only.

This fix #1882

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@ UNRELEASED CHANGES:
 * Add ability to set a reminder for a life event
 * Stop reporting OAuth exceptions
 * Replace karakus/laravel-cloudflare with monicahq/laravel-cloudflare to fix dependencies issues
+* Fix use of 'json' mysql column type
 
 RELEASED VERSIONS:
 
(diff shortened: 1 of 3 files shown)


anmol26s:
If this fixes https://github.com/monicahq/monica/issues/1880
It would be nice to merge it and have a minor release. This has break fresh install on Yunohost.

asbiin:
@anmol26s we still have to test it on mysql > 5.7 and on our official instance. I agree with you: we should release this asap!

djaiss:
@asbiin @anmol26s I'll release something in the next hour. Sorry for the trouble.

anmol26s:
@djaiss 
Thanks for the release.

PR #1888, opened 2018-10-07
chore: fix vuejs v-if and v-else usages

asbiin:
This should not modify any feature


djaiss:
@asbiin this is a dangerous change. We need to test it thoroughly for each one of those changes. I’ll help you testing that. Because I’m sure we’ve broken some stuff here.

asbiin:
@djaiss it's [pure vuejs logic](https://fr.vuejs.org/v2/guide/conditional.html). Don't you believe it can works? ;)

djaiss:
@asbiin two errors on Cypress 😀

One in Activity Types in the Settings, the other when managing a task.

I don't have time tonight but I can help to check what's going on.

asbiin:
@djaiss I also had bizarre error in adding a life event...

PR #1894, opened 2018-10-09
Add PHPStorm project folder to .gitignore

fabwu:
PHPStorm project folder won't show up in staging with this change.


asbiin:
No.
I think it's better if you create a personal global gitignore for your machine, rather than adding specific IDE in our gitignore.
See https://help.github.com/articles/ignoring-files/

fabwu:
I didn't know that. Thanks for the tip!

PR #1903, opened 2018-10-10
feat: hide deceased people in dashboard / people list

### LINKED ISSUE #1345: Do not show deceased people on dashboard & list
Currently, it shows deceased people right on the dashboard.

I use this feature for people I stop contacting. But if it would be real deceased people, it will remind about them always, which is not good.

So I suggest to remove them in a dashboard at least. And, maybe, from the list of contacts too (could make a small link to show them back, as they like in some sort of archive)

tcmal:
Fixes #1345 

Dashboard remains unchanged.

People List:

image
image


djaiss:
Thanks so much @tcmal 

Going to merge this one.

PR #1912, opened 2018-10-12
feat: add ability to upload documents

djaiss:
This PR adds the ability to upload documents to a contact. In this PR, we can't upload a document through the API yet.

2018-10-27-documents

- [x] Delete a document
- [x] When resetting an account, delete all documents
- [x] When deleting an account, delete all documents
- [ ] Increment counter whenever someone downloads the document

### Front-end changes
- [ ] Front-end tests have been written with Cypress.

#### Backend/models changes
- [x] If you change a model, make sure the FakeContentTableSeeder is updated. We need seeders to develop locally and generate fake data.

- [x] Make sure your changes do not break importing data with `vCard` and `.csv` files.

#### Other tasks
- [x] If it's relevant and worth mentioning, create a changelog entry for this change. The changelog entry will appear inside the UI for all users to see. To know if your change is worth the creation of a changelog entry, [read the documentation](https://github.com/monicahq/monica/blob/master/docs/administrators/tips.md#when-is-it-relevant-to-create-a-changelog-entry).


resources/views/people/documents/index.blade.php
```diff
@@ -0,0 +1,5 @@
+<div class="{{ \App\Helpers\LocaleHelper::getDirection() == 'ltr' ? 'fl' : 'fr' }} w-100 pb3 pt1 pl3 pr3">
```
asbiin: Use the helper `htmldir()`
```suggestion
<div class="{{ htmldir() == 'ltr' ? 'fl' : 'fr' }} w-100 pb3 pt1 pl3 pr3">
```

djaiss:
@asbiin I need your help.

Deleting a document doesn't work. The error is about WrongIdException - meaning the hashID of the contact is not right when making the call.

But I don't see what I do wrong.

app/Models/Contact/Document.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace App\Models\Contact;
+
+use App\Models\Account\Account;
+use Illuminate\Support\Facades\Storage;
+use App\Models\ModelBindingHasher as Model;
```
asbiin: @djaiss don't use `ModelBindingHasher` here, unless you send hashid for Documents in vue/axios. Use `ModelBinding` instead, and it's working.

djaiss: Ahhh.

PR #1916, opened 2018-10-13
feat: add export:all command to export as sQL

### LINKED ISSUE #1523: Automatically export SQL
In order to backup my data I'd like to be able to export SQL automatically for example with an API call or a command line. 

The button in Settings is nice but not really usable in an automated way.

Maybe it's possible to call directly ExportAccountAsSQL.php ?

tcmal:
used for backing up the entire database, as opposed to just user content

fixes #1523


djaiss:
Thanks again @tcmal !!

PR #1928, opened 2018-10-19
test: add api tests

asbiin:
Add some tests for api


djaiss:
Wow. That's a huge increase!!

Issue #1819, opened 2018-09-13
Ability to add a description next to a name


djaiss:
Feedback from user:

> I often don’t know someone’s last name. I only know how I met them and their first name. For example: “met at john’s bbq”. I need a way to find these people later. EIther create a description field that is searchable and less temporal than activities, or make activities searchable. If you go with a description field, putting it on the contact list view would make that view more useful. Right now, it’s impossible to distinguish between two people with the same first name and whose last name I don’t know. That goes for both search and list views.

PR #1930, opened 2018-10-20
feat: made contact description field searchable

lattlay:
Makes the description field on a contact searchable.

Feat: https://github.com/monicahq/monica/issues/1819
(Note: description field already appears in contact list view)


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
asbiin: What is this file ?

lattlay: This is the project file of the phpstorm IDE. Would you like me to remove it?

djaiss: I don't think we need to remove it. The .gitignore contains a bunch of stuff like this already anyway.

asbiin: NotI disagree. You can easily add a global gitignore for your needs. See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

lattlay: Cool I learned something new today! I've now removed that line from the .gitignore file

asbiin:
Thank you @lattlay !

PR #1932, opened 2018-10-22
fix: fix api routes

asbiin:
Fixes:
- `POST /activities/` and `PUT /activities/:id` now checks for ActivityType `activity_type_id` to be from the same account
- `DELETE /activities/:id` now works
- `GET /activitytypes/` is filtered to the account
- `GET /lifeevents/:id` now works

Changes:
- `POST /contacts/:contact_id/pets` is replaced with `POST /pets/`, with the use of `contact_id` (the route already exists)
- `PUT /contacts/:contact_id/pets/:id` is replaced with `PUT /pets/:id`, with the use of `contact_id` (the route already exists)
- Every validator fails now send a HTTP 400 code (was 200) with the error 32
- Every Invald Parameters errors  now send a HTTP 400 (was 500 or 200) code with the error 41
- Every QueryException now send a HTTP 500 code with the error 40


djaiss:
@asbiin don't forget the CHANGELOG though!

Issue #1896, opened 2018-10-09
prefix for table names not working


Schrigga:
I use the latest release (2.8.1) and try to get it run. In my configuration I use a prefix for the table names. When I run ``php artisan setup:production`` I will get an error, that the tables doesn't exist. In the message the prefix is used twice. Running the command without the prefix in the configuration works. 

The part of the configuration for the db looks like this: 
```
DB_CONNECTION=mysql
DB_HOST=127.0.0.1
DB_PORT=3306
DB_DATABASE=siw_23121_2
DB_USERNAME=XXX
DB_PASSWORD=YYY
DB_PREFIX=monica_
```

The errors look like this: 
```

In Connection.php line 664:
                                                                                                                                                                                                                  
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'monica_monica_information_schema.schemata' doesn't exist (SQL: select `DEFAULT_CHARACTER_SET_NAME` from `monica_monica_information_schema`.`schemat  
  a` where `schema_name` = siw_23121_2)                                                                                                                                                                
                                                                                                                                                                                                                  

In PDOConnection.php line 82:
                                                                                                                       
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'monica_monica_information_schema.schemata' doesn't exist  
                                                                                                                       

In PDOConnection.php line 80:
                                                                                                                       
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'monica_monica_information_schema.schemata' doesn't exist  
```

PR #1938, opened 2018-10-23
fix: Prefix for table names (DB_PREFIX) is now working

lattlay:
Setting DB_PREFIX in your .env file will now create the appropriate tables and work as expected.

Fix for: https://github.com/monicahq/monica/issues/1896


lattlay:
Unsure why this is failing the `continuous-integration/styleci/pr` test, if you look at the code that I actually committed it is not the same as what is being shown in `continuous-integration/styleci/pr`  😄

app/Console/Commands/Update.php
```diff
@@ -117,12 +117,12 @@ private function migrateCollationTest()
 
         $databasename = $connection->getDatabaseName();
 
-        $schemata = $connection->table('information_schema.schemata')
-                ->select('DEFAULT_CHARACTER_SET_NAME')
-                ->where('schema_name', '=', $databasename)
-                ->get();
+        $schemata = DB::select(
```
asbiin: Not sure why we need to rewrite this part?

database/migrations/2018_10_07_120133_fix_json_column.php
```diff
@@ -20,21 +20,15 @@ public function up()
...
-                                ['table_schema', '=', $databasename],
-                                ['data_type', '=', 'json'],
-                            ])
-                            ->whereIn('table_name', [
-                                'default_life_event_types',
-                                'life_event_types',
-                                'life_events',
-                            ])
-                            ->get();
+        $columns = DB::select(
```
asbiin: This too: why we need to rewrite this part?

app/Console/Commands/Update.php
lattlay: Sure, because when you use the ->table method Laravel automatically appends the DB_PREFIX to the database name. The thing is that information_schema does not have a prefix as it is a common metadata table so it will try to make a query to {DB_PREFIX}information_schema.schemata which always throws an error. I rewrote it to ensure that there will not be a prefix added to this specific query only.

The same thing applies to the other location where I had this.

database/migrations/2018_10_07_120133_fix_json_column.php
lattlay: See above comment, the same thing applies here.

app/Console/Commands/Update.php
asbiin: OK I see! Thanks for the expanations.

Issue #1939, opened 2018-10-23
Contact List to Ability to Open Contact in New Tab


chrisspiegl:
When navigating to the `/people` page on Monica (online and self-hosted), I can only click the contacts one by one. The usual "CMD+Click" to open a contact in a new tab does not work, neither does a "Right Click + Open in New Tab".

Ideal behavior would be that both of these work as per usual to make managing multiple contacts easier without having to go back and forth all the time.

Thanks 💪

asbiin:
Yes, this is done in #1888

djaiss:
This has been merged tonight. Will be in production in the next days, and in a release soon.

PR #1947, opened 2018-10-24
feat: Made contact list support right clicks

lattlay:
Contact list now supports right click. 

Primary use case is for the ability to open contacts in new tabs without leaving the contacts page

Feature: https://github.com/monicahq/monica/issues/1939


djaiss:
Thanks a lot @lattlay for your help!

Issue #526, opened 2017-08-09
Seeding currencies should be performed in a seeder.


aejnsn:
I see that currencies are introduced via a JSON file as a migration. This should be refactored into a seeder. Migrations are for controlling the schema, not the data within said schema.

I realize this change would impact the update path, however supporting multiple currencies is something that would likely involve change to any underlying data as well. The update path could be sustained by constructing a console command ran upon updating or having users updating their installation by manually running an ideally separate seeder class (`php artisan db:seed --class=SeedCurrencies`). Moving forward in later versions, that seeder class would be included in the top-level `DatabaseSeeder` class, just like the production block consisting of `ActivityTypesTableSeeder` and `CountriesSeederTable`. This would allow it would be seamless for new installations just as the migrations presently exist.

See merged PR changes here: https://github.com/monicahq/monica/pull/509

djaiss:
I agree that it should be moved to a seeder instead, definitely.

PR #1956, opened 2018-10-26
Moved currency seeding to a seeder class

pschmidt88:
Removed the currency seeding migration and moved it to a laravel seeder class (as suggested in issue https://github.com/monicahq/monica/issues/526)


djaiss:
Thanks for your PR @pschmidt88 

There is one problem though: while you've moved this to a seeder, you are not actually calling the seeder anywhere, so it won't be called, ever.

pschmidt88:
Yeah, you're right. Forgot to add it to the `DatabaseSeeder`. Add that in the last commit.

Issue #1959, opened 2018-10-27
Bug: Click Chinese Tags cannot get the result.


Wangjunyu:
I have some tags are in Chinese, When I am on Contact list click a English tag like "AI", the result is ok, and the link is '''https://app.monicahq.com/people?tag1=ai'''

But, if I click Chinese one like "合伙人", the link is '''https://app.monicahq.com/people?tag1=''' and cannot get the right result.

PR #1961, opened 2018-10-27
fix: tag filter for non-english characters works now

lattlay:
Previously, adding a non-English tag was not saving the `name_slug` field correctly. This patch fixes that, and included is a migration to fix all currently affected tags.

Fix for: https://github.com/monicahq/monica/issues/1959

### DIFF
--- a/database/migrations/2018_10_27_230346_fix_non_english_tab_slugs.php
+++ b/database/migrations/2018_10_27_230346_fix_non_english_tab_slugs.php
@@ -1,0 +1,17 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+class FixNonEnglishTabSlugs extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        DB::table('tags')->update(['name_slug' => DB::raw('`name`')]);
+    }
+}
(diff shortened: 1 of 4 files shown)


database/migrations/2018_10_27_230346_fix_non_english_tab_slugs.php
```diff
@@ -0,0 +1,17 @@
...
+class FixNonEnglishTabSlugs extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        DB::table('tags')->update(['name_slug' => DB::raw('`name`')]);
```
asbiin: This updates every tags->name_slug with the name of tag.
I don't think it's what is intended?

You should do:
```suggestion
        Tag::chunk(200, function ($tags) {
            foreach ($tags as $tag) {
                if (empty($tag->name_slug)) {
                    $tag->forceFill([
                        'name_slug' => htmlentities($tag->name),
                    ])->save();
                }
            }
        });
```
Don't forget to add the use: `use App\Models\Contact\Tag;`

lattlay: Good catch! I've updated now

djaiss:
@lattlay I'm sorry - this PR was made at the same tag the tags management has been completely changed.

Basically, you need to add `html_entities` here: https://github.com/monicahq/monica/blob/master/app/Services/Contact/Tag/AssociateTag.php

I would do it myself - but if you still want to contribute, let me know 😀

PR #1964, opened 2018-10-28
fix: fix activity buttons spacing

andreasjacobsen93:
The activity buttons broke off into two lines because of the width and the icons was spaced a little weird.

The button now looks as below, which I think is a little nicer.
screenshot 2018-10-28 at 21 39 46


resources/assets/sass/people.scss
```diff
@@ -628,6 +628,7 @@
       a.edit {
         position: relative;
         top: 1px;
+        margin-right: 5px;
```
asbiin: Carefull about the orientation here: RTL langages might need a `margin-left` (to be verified)
```suggestion
    @if $htmldir == ltr {
        margin-right: 5px;
    } @else {
        margin-left: 5px;
    }
```

asbiin: Good question! I don't know yet.
Tip: you can test by adding your langage in (this list )[https://github.com/monicahq/monica/blob/master/app/Helpers/LocaleHelper.php#L72] temporarly.

asbiin:
@andreasjacobsen93 still need to change layout for LTR/RTL. What do you think ?

resources/assets/sass/people.scss
andreasjacobsen93: I think this makes sense.. still need to test it though

andreasjacobsen93:
I think the last change makes sense.. still need to test it tho

asbiin:
@andreasjacobsen93 thank you!

PR #1968, opened 2018-10-29
fix: fix to display non latin names on the dashboard

wgh000:
Quick fix to show latin initials for non-latin names.

Example:
screenshot 2018-10-29 at 18 06 58

Example:
screenshot 2018-10-29 at 18 08 15


djaiss:
Thanks a lot for your help @wgh000

Issue #1966, opened 2018-10-28
AWS_SERVER environment variable ignored.


allanlw:
With commit feded44625af43b63c936daa1d2c26a65dcabe72 the glue in for using non-S3 servers using the S3 protocol (via the AWS_SERVER environment variable) was removed.

This means that if you clone the default config and modify that, to point to e.g. GCP, it will unexpectedly connect to S3.

The obvious simple fix is to just [re-add the glue](https://github.com/allanlw/monica/commit/79c924f562a6595ccbb5a8d39b7b7e3fb4a57c65) which *seems to work for me* but likely would break other installations. I think the more elegant option would be to only set the endpoint option when necessary, but I'm not really sure when that is.

As an aside, with that (small patch) Monica does seem to support Google Cloud Storage just fine (in interop mode), so maybe adding that to the documentation and/or sample config would be worthwhile too.

Happy to open a PR with both once I better understand how to write the patch in a way that won't break existing installs.

PR #1971, opened 2018-10-30
fix: aws server endpoint configuration handling

allanlw:
This fixes bug #1966 by correctly handling the AWS_SERVER environment variable that is outlined in the .env.example

I tested this manually in the following cases:

- S3 configuration with AWS_SERVER unset
- S3 configuration with AWS_SERVER set to the empty string
- S3 configuration with AWS_SERVER explicitly set to the S3 endpoint for my bucket's region
- Google Cloud Storage with AWS_SERVER set to storage.googleapis.com

In all four cases, uploading and viewing avatars works as expected, so I think it's unlikely that this will cause any breakage.


asbiin:
Nice work, thank you for your help @allanlw ! I had removed the 'endpoint' value, because I though it was useless, but it wasn't ...

PR #1996, opened 2018-11-03
feat: vcard services for import/export [wip]

asbiin:


app/Http/Middleware/CheckLimitations.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Support\Facades\Auth;
+
+class CheckLimitations
```
djaiss: I would just change this because it's not super clear what it does.

```suggestion
class CheckAccountLimitations
```

app/Services/VCard/ExportVCard.php
```diff
@@ -0,0 +1,208 @@
...
+    {
+        // The standard for most of these fields can be found on https://tools.ietf.org/html/rfc6350
+
+        // Basic information
+        $vcard = new VCard([
+            'UID' => $contact->hashid(),
+            'SOURCE' => route('people.show', $contact),
+        ]);
+
+        $this->exportNames($contact, $vcard);
```
djaiss: Instead of repeating those two attributes everywhere, isn't it simpler to create  Contact and vCard objects at the class level at the beginning of this file? Like `private Contact $contact` and tell call `$this->contact`?

app/Services/VCard/ImportVCard.php
```diff
@@ -0,0 +1,677 @@
...
+    private function process(array $data) : array
+    {
+        $behaviour = $data['behaviour'] ?: self::BEHAVIOUR_ADD;
+
+        $entry = Reader::read($data['entry']);
+
+        if (! $this->canImportCurrentEntry($entry)) {
+            return [
+                'error' => 'ERROR_CONTACT_DOESNT_HAVE_FIRSTNAME',
+                'reason' => $this->ErrorResults['ERROR_CONTACT_DOESNT_HAVE_FIRSTNAME'],
```
djaiss:
```suggestion
                'reason' => $this->errorResults['ERROR_CONTACT_DOESNT_HAVE_FIRSTNAME'],
```

app/Services/VCard/ImportVCard.php
```diff
@@ -0,0 +1,677 @@
...
+        return array_get($this->genders, $genderCode);
+    }
+
+    /**
+     * Get the gender by name.
+     *
+     * @param  string  $name
+     * @return Gender|null
+     */
+    private function getGenderOfName($name)
```
djaiss:
```suggestion
    private function getGenderByName($name)
```

app/Services/VCard/ExportVCard.php
asbiin: Well, it's more thread safe like this, but I agree there are other solutions.
But like this, it's way way easier to test it!

PR #2015, opened 2018-11-11
fix(avatar): display configured max upload size

### LINKED ISSUE #1988: Upload avatar dialog states incorrect limit
**Describe the bug**
The photo/avatar upload widget claims that the max upload size is "10 Kb." Kb = kilobit, so this would be 1.25 KB, which would be insanely small for a photo file.

Fortunately, Monica doesn't enforce a limit anywhere close to this (I uploaded a 258 KB file fine). Suggest removing this warning or matching it to the actual limit.

**Screenshot**

image

 - A server you maintain yourself: v.2.9.0

ronaldevers:
The form help text was using a hard-coded 10 character limit. I copied
the expression for the configured max size from
`resources/views/settings/imports/upload.blade.php`.

Fixes #1988.


asbiin:
@ronaldevers thanks a lot!

PR #2036, opened 2018-11-14
fix: convert avatar deletion issues into a warning

allanlw:
Although the Contact model does silently ignore FileNotFound exceptions, any other sort of exception getting triggered by the underlying avatar deletion operation currently prevents saving the contact entirely. 

Instead, make this a warning, and log it, but don't stop the entire contact saving process. If there's an error with the avatar deletion, it won't cause data loss, so it's strange to consider this an error.

This recently bit me while trying to get avatars working correctly with S3.


asbiin:
Thank you @allanlw

Issue #2038, opened 2018-11-14
Editing a conversation throws  MethodNotAllowedHttpException


allanlw:
Editing a conversation throws a  MethodNotAllowedHttpException. I am on v2.10.1

To reproduce:

1. create a conversation.
2. Go to edit it.
3. Attempt to save it and receive an exception:

```
(1/1)                                     MethodNotAllowedHttpException
--
in RouteCollection.php line 255
at RouteCollection->methodNotAllowed(array('PUT', 'PATCH', 'DELETE'))in RouteCollection.php line 242
at RouteCollection->getRouteForMethods(object(Request), array('PUT', 'PATCH', 'DELETE'))in RouteCollection.php line 176
at RouteCollection->match(object(Request))in Router.php line 633
at Router->findRoute(object(Request))in Router.php line 622
at Router->dispatchToRoute(object(Request))in Router.php line 611
at Router->dispatch(object(Request))in Kernel.php line 176
at Kernel->Illuminate\Foundation\Http\{closure}(object(Request))in Pipeline.php line 30
at Pipeline->Illuminate\Routing\{closure}(object(Request))in TrustProxies.php line 33
at TrustProxies->handle(object(Request), object(Closure))in Pipeline.php line 151
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in Pipeline.php line 53
at Pipeline->Illuminate\Routing\{closure}(object(Request))in TrustProxies.php line 57
at TrustProxies->handle(object(Request), object(Closure))in Pipeline.php line 151
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in Pipeline.php line 53
at Pipeline->Illuminate\Routing\{closure}(object(Request))in TransformsRequest.php line 31
at TransformsRequest->handle(object(Request), object(Closure))in Pipeline.php line 151
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in Pipeline.php line 53
at Pipeline->Illuminate\Routing\{closure}(object(Request))in TransformsRequest.php line 31
at TransformsRequest->handle(object(Request), object(Closure))in Pipeline.php line 151
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in Pipeline.php line 53
at Pipeline->Illuminate\Routing\{closure}(object(Request))in CheckForMaintenanceMode.php line 62
at CheckForMaintenanceMode->handle(object(Request), object(Closure))in Pipeline.php line 151
[...]

PR #2040, opened 2018-11-14
fix: fix editing conversations

allanlw:
Fixing method errors for editing conversations (#2038) by using the right method in the form.


asbiin:
Thank you so much @allanlw !

Issue #2049, opened 2018-11-14
The 10 contact limit on the free plan takes partial account into accounts


djaiss:
... which it shouldn't.

djaiss:
Thanks @tbirrell and @asbiin

PR #2058, opened 2018-11-15
fix: exclude partials from contact limit #2049

tbirrell:
To correct the problem mentioned in Issue #2049, I have added the `real` scope to the `hasReachedContactLimit` check to exclude any partial contacts from the contact limit.


database/factories/ModelFactory.php
```diff
@@ -91,6 +91,9 @@
         },
     ];
 });
+$factory->state(App\Models\Contact\Contact::class, 'partial', [
```
asbiin: I didn't this feature! Thanks

PR #2067, opened 2018-11-17
feat: add ability to create tasks that are not linked to contacts [wip]

djaiss:
This PR lets users add tasks that are not associated with contacts. Users can now add tasks straight from the dashboard.

This PR also 
* creates new services to manage tasks
* rewrite all unit tests about tasks
* remove unused images from the dashboard


djaiss:
Remains to be done:
- [ ] Rewrite how tasks are managed on the contact page (this is now broken)
- [ ] Use filters in Vue on the dashboard

PR #2075, opened 2018-11-21
fix: display avatars in search

### LINKED ISSUE #1985: Undefined image in quick search results
**Describe the bug**
When doing a quick search the results show an undefined image. This happens regardless of whether or not the contact has an Avatar assigned.

**Screenshots**
screen shot 2018-10-31 at 9 24 12 pm

- Local environment on Master branch

**Additional context**
On line 54 of search.js we're checking if avatar_url is not equal to null when we should be checking if it is `undefined`. However, I think the issue might be a little more big than that. I don't believe that field is being returned via the API, at least on master branch.

### LINKED ISSUE #2057: Missing avatars in autocomplete
Hi,

I found the following issue with avatars in the contact search autocomplete results:

**Steps to reproduce:**

1. Start typing an existing contact first or last name in the top search box
2. Observe avatars in autocomplete results box

**Expected behavior:**
Users avatars are displayed 

**Actual behavior:**
Users avatars aren't displayed

**Additional notes:**

- Users avatars are properly displayed elsewhere
- I'm using self-hosted 2.10.1 version
- Here is the actual HTML workup xnip2018-11-15_12-17-52

asbiin:
This fix #1985 and fix #2057


djaiss:
Woooot

PR #2091, opened 2018-11-23
fix(lang): corrected variable typo in german language

fbuchner:
Fixed a typo so the name variable is displayed correctly on the German people site. Before it displayed an arbitrary string as plaintext instead of the contact name.

This is my first PR here, please feel free to correct me.


asbiin:
@fbuchner Thanks, but our translations are handled in [crowdin](https://crowdin.com/project/monicahq). Please report your fix in crowdin, and we well include in the next release.

PR #2099, opened 2018-11-26
fix: Make FakeContentTableSeeder runnable several times

AdrienPoupa:
I wanted to run the FakeContentTableSeeder several times but I figured it would only work once with a fresh DB. That is because it is first creating a new account, adding the contacts, then creating a second account.

When relaunching it, it would randomly get a ContactFieldType instance, that could belong to account number 2 whereas we run the seeder as account number 1. Then, the following failure is triggered:

ContactFieldType::where('account_id', $data['account_id'])
                        ->findOrFail($data['contact_field_type_id']);

because it tries to access a ContactFieldType owned by account 2 whereas it queries account_id=1.

The fix I propose scopes the request to the current account and only creates the account admin@admin.com if it does not exist yet.

As a result, one can now run FakeContentTableSeeder more than once regardless of the DB state.


djaiss:
Thanks for your help @AdrienPoupa 
But the question is: why would you want to run it several times  😮 ?

AdrienPoupa:
Hi,
Say you want to insert 10 random contacts, then 10 others... Seeders should be runnable several times imo.
Now, I have a special use case where I want to study the behavior of Monica with different set of data and a growing number of records. I wanted to insert 20000 records (people) and it is not possible to do it all at once since the random user api is limited to 5000 entries. I discovered a loading issue on the people page and I'll do my best to fix it (ie paginate it). Thanks!

djaiss:
I’ll merge it now. Awesome work.

PR #2121, opened 2018-12-01
fix: fix editing relationship not working

djaiss:
This reverts #2051


asbiin:
Indeed.
The *method* was not right, not the route, my bad.

djaiss:
:-D

PR #2132, opened 2018-12-04
feat: add ability to upload photos

djaiss:
This PR adds photo upload capabilities to a contact. It's a fork of the PR #2112 that was too big.

@asbiin I haven't done the API methods for this, like I did for documents. I don't think we should allow the upload of files through the API for now (simply because... I don't know how to deal with uploads through an API) 😀

What has been done:
- add a new tab under the profile of a contact, called photos
- this tab is empty by default
- you can upload photos if your account allows you to
- there is the same upload mechanism that the one used on the documents
- the Storage tab under Settings has been updated to include the photos now
- the global method that checks if the user is above its account storage limit has also been updated.


asbiin:
OK, for the API part we can do this later then.

djaiss:
@asbiin just updated the description at the top to indicate what has been done in this PR.

PR #2135, opened 2018-12-06
feat: paginate the Contacts page and improve database performance

### LINKED ISSUE #1992: Pagination on the people page (/people)
Feature Request:
-----------------
Pagination (using jquery datatables or some other similar construct) on the /people page so it only loads 10/25/50/100 contacts at a time and the user can page through and/or search for specific people or sort in a specific order.

Background:
-------------
Just got started with Monica: I'm looking at a number of potential uses, but one of my use cases is trying to maintain a fairly massive contact list (I am up to 18000 contacts). I set it up self-hosted on a little VM and configured it alright, then wrote some Node.JS code to read my contacts (in a custom format) and import them into Monica, creating contact fields for the phone/email, addresses for the contact addresses and a note with the raw json so I'd be able to refer to it in the future for fields that don't have a natural storage place (I have a UUID attached to each contact that I currently store in the food_preferencies field as that is the only field searchable via the API that isn't in use for something else).

screen shot 2018-11-02 at 11 23 10 am

My import is still in progress (not the most efficient code, but I am at 16726 contacts as of this very moment). However, whenever I visit /people it tries to load all 16726 contacts into a single page: it takes quite a while to do, slows the browser down, and I may only be interested in a single contact.

AdrienPoupa:
This commit uses [Vue Good Table](https://github.com/xaksis/vue-good-table), already used for the conversations logs, to display the contacts in a datatable that is paginated on the server side. This was suggested in #1992.
A new route is created to supply the data to the datatable: /people/list
image
Here is the new datatable:
image
One can select the number of contacts to display and search for a contact:
image
Current features such as ordering the table or filtering it with a tag still works, but needs a reload of the page. If we wanted to do it without reload, that would be possible but we should display the contact information in different rows (such as firstname row, lastname row etc so that we could click on it). I don't know if this behavior is wanted or not.
I also improved the tag retrieval mechanisms by creating a query that lists the number of contacts for a single tag:

> select COUNT(tag_id) AS contact_count, name, name_slug from `contact_tag` inner join `tags` on `tags`.`id` = `contact_tag`.`tag_id` where `tags`.`account_id` = 1 group by `tag_id`

This way, instead of having 470+ queries on the contacts page for 152 tags and 134 contacts, only 16 queries are necessary now.
I did my best to respect Monica's conventions, but since those changes are significant, please review and tell me if some of them were not followed.


AdrienPoupa:
I published new assets but CircleCI still fails, I'm not sure why.

djaiss:
Thanks for your help @AdrienPoupa this is pretty impressive.

Tests fail because you forgot to run `php artisan lang:generate` and `yarn run production` before submitting. Also perhaps you weren't up to date with master and the assets there.

We are going to review this PR over the next days. I think I'll also have to tweak the design a little bit. We'll keep in touch!

AdrienPoupa:
That's weird, I did precisely that during my [Update assets commit](https://github.com/monicahq/monica/pull/2135/commits/030955619bb7ceee05868e4bffb87b20675d2761)

Anyway, I'm glad you find my PR useful!

app/Http/Controllers/ContactsController.php
```diff
@@ -570,4 +547,118 @@ public function archive(Request $request, Contact $contact)
...
+                $count++;
+            }
+            if (! is_null($tags)) {
+                $contacts = $contacts->tags($tags);
+            }
+        }
+
+        // filter out deceased if necessary
+        if ($showDeceased != 'true') {
+            $contacts = $contacts->where('is_dead', 0);
```
asbiin: You may use the `scopeAlive` you've added on `Contact` model:
```suggestion
            $contacts = $contacts->alive();
```

app/Http/Controllers/ContactsController.php
```diff
@@ -570,4 +547,118 @@ public function archive(Request $request, Contact $contact)
...
+        if ($user->contacts_sort_order !== $sort) {
+            $user->updateContactViewPreference($sort);
+        }
+
+        $tags = null;
+        $url = '';
+        $count = 1;
+
+        $contacts = $user->account->contacts()->real();
+        $contacts = $contacts->active();
```
asbiin: Merge these 2 lines

app/Http/Controllers/ContactsController.php
```diff
@@ -570,4 +547,118 @@ public function archive(Request $request, Contact $contact)
...
+                'contacts.nickname',
+                'contacts.description',
+                'contacts.job',
+            ];
+            $queryString = StringHelper::buildQuery($searchable_columns, $request->get('search'));
+            $contacts = $contacts->whereRaw($queryString);
+        }
+
+        // get the number of contacts per page
+        $perPage = $request->has('perPage') ? $request->get('perPage') : 30;
```
asbiin: Should we put the default `30` in an application config setting?

resources/views/people/index.blade.php
```diff
@@ -104,73 +104,14 @@
...
-                   @endif
-                  @endif
-                  <span>
-                    {{ $contact->name }}
-                  </span>
-
-                  <span class="i {{ htmldir() == 'ltr' ? 'ml3' : 'mr3' }}">{{ $contact->description }}</span>
-                 </a>
-                </li>
+                <div class="{{ \App\Helpers\LocaleHelper::getDirection() == 'ltr' ? 'fl' : 'fr' }} w-100">
```
asbiin: You can use the monica helper `htmldir()`:
```suggestion
                <div class="{{ htmldir() == 'ltr' ? 'fl' : 'fr' }} w-100">
```

AdrienPoupa:
I think that should be it, let me know if anything should be modified. I put the 30 value in a config setting, and applied your htmldir suggestion to other views that were using getLocale.

AdrienPoupa:
There is a test failure for "can see contacts", I thought it was because of the getDirection() removal so I reverted it but apparently not... I'm not sure what causes it.

asbiin:
@AdrienPoupa `htmldir()` is an alias for `LocaleHelper::getDirection()`, so it's the exact same comportment.
Anyways, @djaiss wanted to review this PR too.

asbiin:
@AdrienPoupa Thanks for your contribution!

AdrienPoupa:
I'm glad to see it merged, thanks!

PR #2138, opened 2018-12-06
fix: reduce the number of queries in the dashboard by eager loading the contacts

AdrienPoupa:
The following pull request eager loads the contacts for the reminders in the dashboard, in `getRemindersForMonth`

Since contacts are accessed several times from the _monthReminder view, load them from the beginning.

On my installation, adding the with line does the following:

| Before  | After |
| ------------- | ------------- |
| 179 queries | 111 queries  |
| 7.8Mb       | 7.2Mb  |
| 185ms       | 140ms  |

As a bonus, if I had not any partial contact, the number of queries would drop to 27, 84ms and 6.60Mb


djaiss:
That is awesome ❤️ 

I love those optimizations.

AdrienPoupa:
You're welcome! However, before merging, be aware that I added a new commit that reduces the number of queries even more for partial contacts (41 queries now on the same installation, 120ms)

asbiin:
Thank you very much @AdrienPoupa !

PR #2144, opened 2018-12-09
feat: add foundation to support emotions and ability to log emotions during a phone call

djaiss:
This PR adds the foundation for supporting emotions throughout the platform. That means we'll be able to use emotions to describe how someone/something made us feel.

I've created an object `Emotion` to represent this. An emotion is universal and therefore not customizable through the Settings, nor does it belong to an Account.

For instance you could add emotions to a phone call, representing that it made you feel happy and discouraged at the same time.

Later on, we would be able to have interesting data, like how many times you are happy after a call, or who makes you happy the most, etc...

The list of emotions is based on http://changingminds.org/explanations/emotions/basic%20emotions.htm

and comes precisely from 

Shaver, P., Schwartz, J., Kirson, D., & O'Connor, C. (2001). Emotional Knowledge: Further Exploration of a Prototype Approach. In G. Parrott (Eds.), Emotions in Social Psychology: Essential Readings (pp. 26-56). Philadelphia, PA: Psychology Press.

image

As a proof of concept, this PR also adds the ability to add emotions when you log a call.


djaiss:
This is what it currently looks like to log emotions for a phone call.

image

keunes:
I saw some relevant strings pop up in CrowdIn, but not all emotions. @djaiss Can users tag primary, secondary and tertiary emotions, or only (a) particular group(s)? And shouldn't these emotions be translated also? Would be interesting to see if they are translated already in academic literature.

PR #2150, opened 2018-12-11
Test activation of AutoScan

bellingard:


bellingard:
Sorry guys for the noise, please forget this PR, it was not created on the right repository.

Issue #2152, opened 2018-12-12
Feature Request: Add Support for Maiden Names


tbirrell:
This is something I've added to my personal instance. But I think it would be good to add to the main project. At the moment I just have an extra colum in my DB that I edit from my IDE. But from the front, it looks something like this:

![image](https://user-images.githubusercontent.com/17275888/49890925-e894e400-fe0a-11e8-8a85-b87ce13549eb.png)

If whoever works on this (assuming its not me) wants my code to use as a foundation, I have a migration for the DB, some extra methods in `Contact.php`, and the change to `_header.blade.php` (which is visible in the above picture). What I don't have is any sort of form for adding/editing the information, or any code in the Controller to save it.

mccollam:
Does it make sense to generalize this some?  People change names for several reasons, sometimes first name as well.  I've been using the description field to capture this now, but it's free-form (e.g. "birth name was XXX").

This may be too specific to my contacts but I could see a use for tracking name changes and then allowing searches even on old names.

tbirrell:
@mccollam That would probably require a significant redesign of `monica.contacts`. Possibly even moving names completely out to a pivot table. This is enough of a change to justify its own feature request ticket, I think.

asbiin:
I disagree with the term "maiden name" which is associated only to women, and thus could be taken as a sexist term. I prefer something ile "birth name".

mccollam:
Actually "birth name" would capture a lot of my cases as well (someone changes their name, changes gender, etc.)

tbirrell:
> I disagree with the term "maiden name" which is associated only to women, and thus could be taken as a sexist term. I prefer something ile "birth name".

It's a legal term though. So should not be a problem. It is also very specifically for recording pre-marriage names, as you can see from how I display it on the frontend. 

If we want to put this on hold until we can come up with a more comprehensive solution for alternative names, I'm cool with that.

trwnh:
"Former name" would cover more use cases than "birth name". As an edge case, consider a case where a person has a legal name change, twice.

Alwaysin:
Was searching for where to input maiden names, didn't think it wouldn't be there :(
I put the info in the description of the contact for the time being.

immanuelfodor:
I've just run into this issue as I was looking for a way to store former name for a person who changed name after marriage. I also have some people who legally changed names even as male because they didn't like their given name. Former name +1

PR #2164, opened 2018-12-17
feat: add support for maiden names [wip]

tbirrell:
Per #2152, here is the beginning of the work for maiden name support.

I wanted to make the Maiden Name field toggle based on the Gender, but aside from Monica, I have no experience with Vue and couldn't figure it out. And since custom genders are supported now, it might not be worth trying to pick and choose.

Aside from adding in maiden names, the only other change was moving nicknames down to a new row. This was for two reasons. First, so it could fit. And second, so that when we eventually add the middle name field, it can also fit.

This is what the current edit form looks like

screenshot-monica tapl io-2018 12 17-12-10-42

I'm not sure what tests, if any, you want for this. I tried to find other tests for adding a DB field, but couldn't. So if you could point me in the right direction, that would be appreciated.

If anyone else wants to contribute, I am fine with this.
Remaining work includes:

- [x] Saving maiden names
- [ ] Writing Tests (see above)


tbirrell:
I'm not sure why circleci is still failing with:

```
Monica asset are not up to date.
Please update the Monica assets yourself by running:
 ~ php artisan lang:generate
 ~ yarn run production
Exited with code 2
```

I've run both those commands.

asbiin:
"Maiden" is only for women, right?
It would be more interesting to have a solution for everyone, like "birth name"?

tbirrell:
@asbiin, I have responded to your comments over in [the issue](https://github.com/monicahq/monica/issues/2152#issuecomment-448268791).

tbirrell:
Regarding the discussion in #2152, it seems as if we will be wanting to go a slightly different route for this. If you want to use this code in the interim, that's fine by me. Otherwise, we should probably close this PR.

PR #2115, opened 2018-11-28
fix: add asset tags to resources so they load properly in non-root directories

bendauphinee:
I plan to make an entry once I've got more of this fixed. I wanted to get this small PR in first to make sure I've got it right about contributing.

(Partially) Fixes #139

PR #2274, opened 2019-01-08
fix: add more asset/url tags to resources/links so they load properly…

guessit-dev:
… in non-root directories

Follow up of #2115 and part of #139 which should be reopened because after this PR axios request still won't work in a subdirectory.

Some images in *.vue files also need to be made relative.

But I don't have the knowledge to do the other parts for now.


asbiin:
@bendauphinee What is the problem with this solution with more `assets` and `url`?

bendauphinee:
> @bendauphinee What is the problem with this solution with more `assets` and `url`?

More calls more complexity :)

PR #2283, opened 2019-01-10
fix: fix label for attributes

asbiin:
Fix all <label for=... attributes: some are missing, and other are not working.
- fix Input.vue, Select.vue and Textarea.vue


resources/assets/js/components/partials/form/Select.vue
```diff
@@ -59,7 +69,7 @@ export default {
       type: Boolean,
       default: true,
     },
-    formClass: {
+    iclass: {
```
djaiss: why this name?

Issue #2242, opened 2019-01-02
Adding long contact causes notes, reminders to drop to bottom of page


aheinricy:
If you add a contact that is 39 characters or greater it causes the main section to drop down to the bottom of the page.

Steps to reproduce:
1. Under "Contact Information" click edit.
2. Click Add
3. Add contact type of "Facebook"
4. Add content of 39 characters (eg. https://www.facebook.com/xxxxxxxxxxxxxx)
5. Notice that main area moves down to bottom of page.

djaiss:
Sorry about this. It kept happening but I never knew how to fix this. Now I know.
I'm going to fix it.

PR #2288, opened 2019-01-12
feat: use contact field name instead of value

djaiss:
Fix https://github.com/monicahq/monica/issues/2242


asbiin:
Well I'm not happy with that. I would have been better to trunc the result with some defined length, and add `...` if it's too long. The url will be the same.

PR #2296, opened 2019-01-12
 chore: replace php-vcr by guzzle when mocking external api calls

djaiss:
https://github.com/php-vcr/php-vcr was used to mock external api calls.

But in practice it worked like 24% of the time in the CI, and almost never locally.

We are now mocking external api calls with Guzzle instead. It's a bit of more work, but it just works. So...


djaiss:
@asbiin read how I do now!

PR #2301, opened 2019-01-13
chore: add paratest for parallel tests [wip]

djaiss:
@asbiin check this out.

This reduces drastically the tests by running the tests in parallel with https://github.com/paratestphp/paratest (using `./vendor/bin/paratest` to run them). However by doing so, there is one deadlock.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,7 @@
     },
     "require-dev": {
         "barryvdh/laravel-debugbar": "^3",
+        "brianium/paratest": "^2.1",
         "enm1989/chromedriver": "^2.36",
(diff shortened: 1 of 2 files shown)


djaiss:
Lol it works here, but not locally because of deadlocks.

djaiss:
Doh I'm stupid. It works because it ran phpunit, not paratests.

asbiin:
It's really not working.
- there are a lot of Deadlock
- it cannot run only our 4 testsuites. it wants to run Dusk tests too ...

djaiss:
That was just a try. Might be interesting to revisit in the future though.

PR #2304, opened 2019-01-14
feat: support CalDAV for birthdays and tasks

asbiin:


app/Http/Controllers/CardDAV/CardDAVController.php
```diff
@@ -3,19 +3,22 @@
 namespace App\Http\Controllers\CardDAV;
```
djaiss: Shouldn't the name of this file `CardDAVController.php` be renamed to something else? Both Carddav and Caldav are from the WebDav protocol, so perhaps it would be a better name if we are to have one controller.

asbiin:
In addition to this PR, I want to move both CardDAV and CalDAV to 
- a namespace "DAV"
- a url "/dav" : that will be a breaking changes, but carddav is only alpha feature for now.

PR #2223, opened 2018-12-28
feat: refactor reminders

djaiss:
This PR completely refactors how reminders are managed.

- The `Notification` object has been completely removed.
- There are 3 tables for reminders:
    - `reminders`
    - `reminder_outbox`
    - `reminder_sent`
- Special Dates do not have a `reminder_id` column anymore
- A reminder is now only defined by an `initial_date`.
- A reminder in itself is not sent - once it's scheduled, we create a ReminderOutbox object that will be sent.
- Think of ReminderOutbox as the queue for reminders that need to be sent.
- A ReminderOutbox object has a `nature` field which can contain two values: `reminder` and `notification`. `notification` means that it's a reminder that needs to be sent a few days the actual reminder is sent - these correspond to the Reminder Rules set in the account.

**New processes**
- Once a reminder is created, we create ReminderOutbox objects
    - one of the `reminder` type
    - two for each reminder rules setup in the account.
- Once a reminder outbox is sent:
    - we create a new corresponding entry in the `reminder_sent` table and delete the Reminderoutbox object.
    - we schedule the new reminder outbox if it's a recurring reminder
    - we mark the corresponding reminder if it's a one time reminder
    - we delete the reminder outbox
- A reminder is scheduled for a specific user.
- The only cron that remains with this new system is the `send:reminders` command which takes care of everything.
- Special dates do not have a `reminder_id` anymore. It shouldn't be special dates' responsibilities to know if they have a reminder associated with them. All reminders should be linked to contacts. Therefore, we got rid of the `setReminder` method in the `specialDate` class.
- We don't store anymore on the reminders table the next expected date of the reminder - we only store the initial date, then we always calculate the next date on the fly.
- Some reminders shouldn't be deleted from the UI - for instance, reminders about birthdays. For those birthdays, we don't display the Edit and Delete buttons in the list of reminders - this is controlled by the new `delible` field in the reminder object.

**Other changes**
- Create all services for reminder management
- Add foreign keys for reminders, contacts, reminder rules, special dates

**Limitations**
- No API methods for managing ReminderOutbox and ReminderSent objects yet.
- Reminders that have been sent are not displayed in the UI yet.

--- 
- [x] Make sure existing data is migrated

PR #2318, opened 2019-01-19
fix: fix reminders

asbiin:
This is a fix to recent #2223 
- use user Timezone in `calculateNextExpectedDate` to get the right next expected date, and add a test for it


asbiin:
@djaiss OK is it better now ?

PR #2348, opened 2019-01-27
chore: refactor activity type categories [wip]

djaiss:
- Rewrite activities type management with services
- Change location of activity type category and activity type objects from the `Contact` folder to `Account`, as activities belong to an account instead of a contact. Indeed, we'll be able to add activities without specifying contacts (you could do activities alone).

To not forget:
- [x] use singleton in tests


djaiss:
Please review @asbiin then I can finalize activities.

PR #2368, opened 2019-01-30
chore: add laravel-mix-purgecss

asbiin:
This reduce the size of our .css file from 200Kb to 50Kb


djaiss:
This is mainly because of Bootstrap I think. Lots of unused classes.

Issue #2032, opened 2018-11-13
Dashboard avatars of inconsistant size and shape


scheidm:
Image avatars show up as an inconsistent size, depending on how many images v.s. initials are in the list of last consulted.

[A list heavy with images looks ALMOST normal](https://photos.app.goo.gl/bZAVN77AbdNd83X56)

[A list with only 4 images is barely useful](https://photos.app.goo.gl/pP1p4hKPUQ7u9Beh7)

[A list containing only 3 reduces the images to useless slivers.](https://photos.app.goo.gl/TUkFoXk371tjvNW57)

Note I am using the hosted version on https://app.monicahq.com

Thanks for listening!

scheidm:
not knowing the stack very well, I did some monkeying in dev tools before digging into the code.
I moved 'Last Consulted' to before the flex container and added the following css rules:
.mw9 .flex.items-center {    flex-wrap: wrap; }
.mw9 .pr2 {    float: left; }

[The result isn't perfect](https://photos.app.goo.gl/NcfwsNZ5fiec7zpA6) but looks [passable at multiple sizes](https://photos.app.goo.gl/JLRNYZdJWHTBmoVJA}

I didn't see any break point references in the css, and I'm not sure how you all are handling reflow, but I took a [quick stab at further cleanup](https://photos.app.goo.gl/YfbEhFKNzXbEMN678), were there a breakpoint to work with:
.mw9 .pr2 {    padding:0.25rem; }
.mw9 .w-70  {   display: flex;
                          flex-direction: column; 
                          width: 100%;
                          padding: 0.5rem; }

djaiss:
Indeed. These look terrible. This will be completely fixed by https://github.com/monicahq/monica/pull/2112

pqhf5kd:
It looks like this was fixed by #2374. Can it be closed? Or are you waiting on #2112?

Issue #792, opened 2018-01-11
Mobile view of new dashboard is broken


erdmenchen:
The new dashboard design (released with version 1.6.0) is not only not mobile friendly, but also breaks the layout:

![monica - broken mobile dashboard]()

Issues:
* Avatars are overflowing (and too big)
* Avatars with profile pictures are not shown
* Misplaced "Add someone" button
* Content layout is too narrow for 2 column setup (see statistics container) -> would be better to use single column setup for mobile view

tbirrell:
I also have experienced this. On the iPhone SE it doesn't try to do two columns, but it is still very broken.

jarradh:
+1 mobile layout is broken, i want to use Monica on the go 90% of the time

pqhf5kd:
This is still an issue on Monica Version: 2.11.2 and Safari 12.0.1

Bounz:
Hi @djaiss.
As the problem is still there I dug a little bit and it seems that the problem is around flexbox (without wrap) with last consulted contacts.
Here are some screenshots.
iPhone:
![actual_iphone](https://user-images.githubusercontent.com/3062744/51801244-3dba7700-224c-11e9-82b6-56fcdb501bbf.png)

iPhone (the page scrolls horizontally 😥 ):
![actual_iphone2](https://user-images.githubusercontent.com/3062744/51801249-41e69480-224c-11e9-9ef6-312762f16b17.png)

iPad
![actual_ipad](https://user-images.githubusercontent.com/3062744/51801263-56c32800-224c-11e9-891e-de564122d6de.png)

iPad (landscape):
![actual_ipad_landscape](https://user-images.githubusercontent.com/3062744/51801270-60e52680-224c-11e9-9c73-c7cc32923d99.png)

I spent a couple of hours playing with the layout and trying to show all last consulted contacts, but there were problems with a unification of the layout between mobile, tablet and desktop.
Finally, I propose to abandon flexbox and replace it with div with `overflow: hidden` and make avatars a floatable divs. And also organize "Last consulted" block in one column on mobiles. This will result in following:
iPhone:
![proposed_iphone](https://user-images.githubusercontent.com/3062744/51801310-000a1e00-224d-11e9-90eb-bc0cd302471b.png)

iPad:
![proposed_ipad](https://user-images.githubusercontent.com/3062744/51801315-0b5d4980-224d-11e9-94f8-8499d3e4ecb2.png)

iPad (landscape):
[...]

PR #2374, opened 2019-01-31
fix: fix dashboard layout on mobile and tablet

Bounz:
Hi @djaiss and @asbiin.
As the problem stated in #792 is still there I dug a little bit and it seems that the problem is around flexbox (without wrap) with last consulted contacts.
Here are some screenshots.
iPhone:
actual_iphone

iPhone (the page scrolls horizontally 😥 ):
actual_iphone2

iPad
actual_ipad

iPad (landscape):
actual_ipad_landscape

I spent a couple of hours playing with the layout and trying to show all last consulted contacts, but there were problems with a unification of the layout between mobile, tablet and desktop.
Finally, I propose to abandon flexbox and replace it with div with `overflow: hidden` and make avatars a floatable divs. And also organize "Last consulted" block in one column on mobiles. This will result in following:
iPhone:
proposed_iphone

iPad:
proposed_ipad

iPad (landscape):
proposed_ipad_landscape

iPhone (RTL):
proposed_iphone_rtl

iPad (RTL):
proposed_ipad_rtl

Also, this solution fixes the issue with squeezed avatar images (like those in #2032).
What do you think about this?


asbiin:
This is really great! Thank you @Bounz !!

PR #2389, opened 2019-02-07
refactor: use parent class to avoid duplicate code

Bu4ak:
This PR will help keep the code cleaner. Those form requests that have custom logic of `authorize()` method must extend the base `Request` class. All others - `AuthorizedRequest`


.gitignore
```diff
@@ -20,3 +20,4 @@ php-extensions-*.tar.bz2
 monicadump.sql
 .scannerwork/
 .sentry-release
+.idea/
```
asbiin: I don't know this directory, but if it's platform specific, you may want to add it to your personal global gitignore.
See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

Bu4ak: When you use the PhpStorm IDE, all settings for project are stored under the `.idea` folder

PR #2406, opened 2019-02-09
refactor:  fix commands registration

Bu4ak:
Using `:: class` instead of hard-coded strings. Many IDE link these class names. This will help avoid errors after refactoring code.
And it looks more readable.


app/Console/Kernel.php
```diff
@@ -13,25 +32,25 @@ class Kernel extends ConsoleKernel
...
+        SendReminders::class,
+        SendStayInTouch::class,
+        SentryRelease::class,
+        SetupProduction::class,
+        SetupTest::class,
+        SetupFrontEndTest::class,
+        SetPremiumAccount::class,
+        SetUserAdmin::class,
+        Update::class,
+        MigrateDatabaseCollation::class,
```
asbiin: We should sort this list alphabetically, could you just do it?

Bu4ak: Is done. But I hope that in the future, other developers will follow the sorting when adding commands)

asbiin:
Thank you @Bu4ak!
We will indeed check the sorting of commands in the future.

PR #2428, opened 2019-02-12
ci: upgrade to psalm v3

muglug:
Was previously blocked on this because Psalm used a clashing version of `sabre/event` for its language server. It now uses `amphp/amp`, so everything is good.

I'm using the opportunity to write a Laravel plugin - support for that will likely come in a forthcoming PR.


asbiin:
@muglug Thank you! We could not used v3 indeed, so that's nice!

PR #2432, opened 2019-02-12
fix: fix contact information update

### LINKED ISSUE #2390: Description is not saved
I tried creating new contacts and editing contacts in the hosted version of Monica. I found the following bugs:
- photo uploads fail, an S3 URL is generated but that URL shows "Access Denied" when trying to access it directly
- any value entered in the description field is not saved and/or not displayed

asbiin:
This fix #2390
I've also removed all `$contact` services properties.

### DIFF
--- a/app/Services/Contact/Contact/CreateContact.php
+++ b/app/Services/Contact/Contact/CreateContact.php
@@ -82,7 +82,7 @@ class CreateContact extends BaseService
         $contact->save();
 
         // we query the DB again to fill the object with all the new properties
-        $contact->refersh();
+        $contact->refresh();
 
         return $contact;
     }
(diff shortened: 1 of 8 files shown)


app/Services/Contact/Contact/CreateContact.php
```diff
@@ -72,44 +70,47 @@ public function execute(array $data) : Contact
...
 
-        $this->contact->setAvatarColor();
-
-        $this->contact->save();
+        $contact->setAvatarColor();
+        $contact->save();
 
         // we query the DB again to fill the object with all the new properties
-        return Contact::find($this->contact->id);
+        $contact->refersh();
```
guessit-dev: refresh?

asbiin: :D Of course !

asbiin: Thank you @guessit-dev

PR #2434, opened 2019-02-12
Set up CI with Azure Pipelines [wip]

asbiin:


djaiss:
What are the benefits? Integration with GitHub?

asbiin:
@djaiss I don't know yet, it just a try here.
But, we could have 10 parallel working build ...

PR #2435, opened 2019-02-13
refactor: fix bad return types with Psalter

muglug:
Psalm comes with a tool called Psalter that fixes issues in your codebase.

I ran

```
vendor/bin/psalter --issues=InvalidReturnType --safe-types=true
```

with a few small manual adjustments.


asbiin:
Thank you @muglug !
What's wrong with all the `@return \Illuminate\Http\Response` ? It should be right from what I saw.

muglug:
> What's wrong with all the `@return \Illuminate\Http\Response` ? It should be right from what I saw.

In short, it's not right. For example, `\Illuminate\Http\RedirectResponse` does not extend `\Illuminate\Http\Response`

djaiss:
Nice insights, thanks for sharing.

asbiin:
Should we run this command on the CI to check the return type, or is the command launched with psalm?

muglug:
@asbiin Psalm will emit an error if return types are incorrect from here on out - the PR includes a Psalm config change, which tells Psalm that two issues that were once warnings should now stop the build if Psalm encounters them: 

screen shot 2019-02-15 at 14 45 53

asbiin:
Thansk you @muglug for this improvement!

muglug:
You're welcome!

Issue #835, opened 2018-01-20
SSL termination issues


jwynn6:
I have a SSL termination/load balancer/reverse proxy that sits in front of Monica.  When trying to enter data, the site now warns that data may be sent insecurely, upon clicking 'ok' the site reloads and nothing is inserted. 

Suggestions?
<img width="478" alt="screen shot 2018-01-19 at 11 45 19 pm" src="https://user-images.githubusercontent.com/6757260/35179942-f6a69636-fd72-11e7-8121-db746c344788.png">

degan6:
Interesting, looks like your on a mac but let's start with some other questions. 
What browser are you using?
Do you connect to your reserver proxy over https?

t-rascal:
Hello, I am experiencing a similar issue and believe it is related to the closed issue #479. 

I am receiving a warning about an insecure form submission when trying to add a significant other to a contact, and clicking submit then leads to the "Whoops, something went wrong" screen. 

According to issue #479 it's due to use of HTTP calls... The noted work around is to enable HTTPS in the docker container directly. I've been trying to do that but being a docker newbie no luck as of yet (I've never built my own image before - and the base image doesn't include a2enmod so I'm stuck!)

Anyway other info - 
Client browser: tested on both Chrome/Win10, Safari/Mac and Safari/iPhone. 
Host: official MonicaHQ docker image on RancherOS. 
Reverse Proxy: NGINX with SSL termination proxying to HTTP Monica port. 

Please let me know if I can provide any useful logs etc. 

Finally my apologies if this is a separate issue, in which case please disregard the above!

jwynn6:
I was on a Mac. I get the same result mobile or on Linux too. Firefox and Chrome. 

Browser ----(https)--- nginx proxy ---- (http)--- Monica

Hope that diagram helps.

On January 19, 2018 10:10:41 PM PST, degan6 <notifications@github.com> wrote:
>Interesting, looks like your on a mac but let's start with some other
>questions. 
>What browser are you using?
>Do you connect to your reserver proxy over https?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/monicahq/monica/issues/835#issuecomment-359148877

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

t-rascal:
Today I experimented by changing the reverse proxy setup as follows:

Browser -----(http) --- nginx proxy ---- (http) --- Monica.

[...]

PR #2439, opened 2019-02-14
fix: switch to HTTPS schema early

FranPregernik:
When using nginx as a https termination reverse proxy the assets and
URL in the html were generated as HTTP causing the browser not to
trust them, especially the login form.

This fixes the issue by switching to HTTPS early in the app startup
allowing laravel to generate URLs to have the HTTPS schema.

Issue #835


asbiin:
Thank you @FranPregernik !

PR #2442, opened 2019-02-15
Removed --no-suggest option from composer

tomgs:
The option does not exist.

### DIFF
--- a/docs/installation/ubuntu.md
+++ b/docs/installation/ubuntu.md
@@ -97,7 +97,7 @@ exit
 
 1. `cp .env.example .env` to create your own version of all the environment variables needed for the project to work.
 1. Update `.env` to your specific needs. Don't forget to set `DB_USERNAME` and `DB_PASSWORD` with the settings used behind.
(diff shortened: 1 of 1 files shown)


asbiin:
@tomgs Thank you for this suggestion.
But ... it exists:
https://getcomposer.org/doc/03-cli.md#install-i
Should we keep it? That's another story.

tomgs:
Alright, so it's a version thing. The version installed by default from the ubuntu repo is outdated and does not contain that option. Should we force the user to get the latest composer, or at least one that contains that option?

asbiin:
Is there an error with this old version?
The option is only display with `composer install --help`.

tomgs:
I've just received it when trying to install composer....

asbiin:
Maybe we can suggest to update composer before installing packages instead?

tomgs:
Yeah, probably the right way to go.

djaiss:
Should we do something more here then?

tomgs:
Delete this, as a person coming with a clean install will probably not see the error. It's on me for using an outdated version.

PR #2462, opened 2019-02-20
fix: deleting a LifeEvent no longer deletes the associated Contact

### LINKED ISSUE #2415: trying to delete life event deletes contact
**Describe the bug**
how to reproduce: open contact page, open live events. Click "delete". In the Pop up that opens, click the green "Delete" Button. Nothing seems to happen. Reload the page or click cancel and reload the page.

**Screenshots**
image

image

After Reload:

image

 - Hosting version on https://app.monicahq.com

mechanarchy:
This PR resolves #2415. There were two bugs present:
* The parameter in the web routing file was called `lifeevent` when it should have been `lifeEvent`. This caused a null argument to be given to `\App\Http\Controllers\Contacts\LifeEventsController::destroy`
* The null argument caused the `DestroyLifeEvent` call in `LifeEventsController::destroy` to fail, which resulted in the `catch` block being entered; this naively calls `back()` which issues an `HTTP 302` response. This is not an issue for `GET` requests, but for other methods including `DELETE` the correct response is `HTTP 303`. **This bug ~is~ may still be present in other areas.**

This is only my second PR so once again I hope I've done everything correctly! I will try and audit the code for other erroneous `back()` calls in the next few days.


asbiin:
Wow that's a really good job @mechanarchy , thank you

PR #2465, opened 2019-02-21
fix: Journal view now includes Activities as intended

### LINKED ISSUE #2423: Journal does not load
When I go to the journal view, the page is empty, it only shows that it loads content. But I can confirm with developer tools that the entries were fetched from `/journal/entries` correctly, they're just not displayed.

mechanarchy:
Resolves #2423. The Vue code was looking for the wrong type string, and so Activity entries never rendered.


djaiss:
Thanks a lot for your help!

JackAction:
I still seem to have this problem.
Just created a subscription and imported my contacts. 
Created an activity:
image

But the journal does not show it, it only shows a manually created entry:
image

sigsec:
These changes are in the code base but are not released yet; they will be in the next version after 2.12.0

mechanarchy:
@JackAction These changes are in the code base but are not released yet; they will be in the next version after 2.12.0

casidiablo:
I'm running `2.12.1` and still can't see activities in my journal.

PR #2468, opened 2019-02-24
Fix contact missing create form

### LINKED ISSUE #2464: Contact creation form on "Person not found" doesn't work
**Problem**
If you try to go to a contact page that no longer exists, you'll be redirected to `/people/notfound`. This page suggests that you add a new contact, using a form like this:

image

Trying to create a contact using this form doesn't work, and it also doesn't allow you to select any custom genders. I suspect this bug was introduced in #879 or a related change.

The standard form for adding new contacts (at `/people/add`) looks like this:

image

**Suggested resolution**

I'd suggest that the `/people/notfound` page should display the same form as `/people/add`, and the old form code is removed.

As a quicker alternative, the `people/notfound` page could have no form and instead a link that redirects you to `/people/add`.

asbiin:
This fix #2464

Create and missing forms are now the same.


mechanarchy:
Thanks! This is so much better than what I would have done!

djaiss:
@asbiin I've made a small change - it should be good to go now

PR #2469, opened 2019-02-24
fix: gift label ordering is now consistent

### LINKED ISSUE #2292: Inconsistent Order: Gift idea   Gift received   Gift offered
Gift idea, Gift received and Gift offered are in a different order on the profile and Gift Management pages.

Profile:
screenshot 2019-01-12 at 14 30 56

Gift Management:
image

I'm using the hosted version

mechanarchy:
This PR resolves #2292 by making the order of the 'Gift' options consistent between the profile and the Gift Management page: firstly ideas, then offered, then received. No `CHANGELOG` entry for this because it seems pretty trivial.

There might be a way to merge the translation strings for the profile page ("Gift ideas (11)") with the gift management page ("Gift idea") but this doesn't account for singular/plural and RTL languages.


asbiin:
Thank you for your work @mechanarchy

PR #2475, opened 2019-02-27
 fix: add default protocol links to LinkedIn, Twitter contact field types

### LINKED ISSUE #1030: Adding Contact Information gives bogus link
I added a contact information piece and it automatically added a link to:
https://app.monicahq.com/people/whateverIhadinthecontactinfo

Which doesn't make sense.

### LINKED ISSUE #2044: Link to Twitter profile
The link to Twitter profile is wrong on the version on https://app.monicahq.com.

screenshot from 2018-11-14 11-24-57

The link points to `https://app.monicahq.com/people/jane.doe` instead of `https://twitter.com/jane.doe`.

### LINKED ISSUE #2045: Link to LinkedIn profile
Similar to monicahq/monica#2044 but for LinkedIn. The link should be something like `https://www.linkedin.com/in/jane.doe/`.

mechanarchy:
This PR should resolve #2044 and resolve #2045. While I could have added Facebook links as well, I chose not do this because they are not always consistent - generally `facebook.com/{profile}` but I also often find `facebook.com/profile.php?id={id}`.

I'm not 100% sure that the way I've done this is correct - I've **updated the database migration files** as these were the only locations I could find creating/updating the fields. I do also notice that the VCard importer and exporter rely on fixed URL strings to operate. I haven't modified these but it's something to make note of. 

I've also **adjusted how the links are rendered** to prevent accidental internal links (resolves #1030). The new logic is:
* If a `protocol` is provided, generate a link to `protocol + data`
* Else, if `data` contains `://`, generate a link to `data`
* Else, output `data` without creating a link

I hope these changes are suitable!


mechanarchy:
I've just had a thought - this could potentially produce a breaking change for users if they expect the (erroneous) links to be produced and it suddenly stops happening. I doubt this will be a problem but better to be aware of it in case there are any questions raised.

tbirrell:
You wouldn't want to make changes to old migration files, you would want to create a new migration file with the update. Because if my DB is current, I'll never run your changes. 

Also, you would want to make sure the contact field exists (it'll break the migration if I've deleted my LinkedIn or Twitter fields) and that the user has not already populated it with something already. 

If your new logic works how I think it work, thats a good idea.

mechanarchy:
Thanks for the information! Yeah I haven't done it that way so I'm clearly on the wrong track. I'll look into making a new migration file today with the changes you've suggested.

My hope was to only adjust the fields for _new_ users; it's likely that existing users have either fixed the field protocols themselves or have put in full URLs into the contact form, so changing the protocol for them would produce an incorrect result.

tbirrell:
If I understand your logic correctly, if it finds a URL in the data already, then it doesn't add the protocol. So that _should_ be fine. If not, it's trivial to check before rendering.

mechanarchy:
@tbirrell Regarding your earlier comment with the logic:
>If I understand your logic correctly, if it finds a URL in the data already, then it doesn't add the protocol. 

This is not correct. If a protocol is set (ie not null and not empty), it creates a hyperlink to `protocol + data`. If this check passes, it does not examine whether the data is a complete URL or not. So it is possible that a user could set a protocol of `https://facebook.com/` and a data of `https://facebook.com/zuck`; this would lead to the HTML tag `<a href="https://facebook.com/https://facebook.com/zuck">`.

tbirrell:
In that case, create an [accessor](https://laravel.com/docs/5.8/eloquent-mutators#defining-an-accessor), and put something like `return $protocol . preg_replace($protocol, '', $data);`. 

Also, this branch history is a mess. It might be worth pulling master, checking out a new branch, then selectively copying over your editing files. That would look something like this:
```
git checkout master
git pull monicahq master
git checkout -b profile-field-links
git checkout 2044-profile-field-links -- path/filename.php
```

See [this StackOverflow Q/A](https://stackoverflow.com/questions/2364147/how-to-get-just-one-file-from-another-branch/2364223) to see more details.

mechanarchy:
Closing this pull request to create a new one without all the mess

Issue #929, opened 2018-02-24
Brainstorm: export account data


djaiss:
Right now it's possible to export your account's data in SQL. That way you can take your data and use them on a server you own.

However, this is rather technical. We should be able to export in another, more user-friendly format.

The problem with this: which format? A contact have a lot of different kind of data, and I can't just put everything in one gigantic CSV for instance.

We could have one CSV per model but is it useful? Let's brainstorm the simplest way to achieve this.

asbiin:
Good idea. SQL is not a user friendly model.
JSON or XML is the best way to do this. We could have a model for this export data and use the same model for the API ?

asbiin:
JSON export was already proposed in #32

jonasvp:
JSON and XML are fine for developers but those are probably already served well with the SQL dump.

For end users I would propose using vCard (VCF) or CSV, like @djaiss proposed. VCF is already supported and it would allow importing contact information e.g. in Thunderbird. A VCF file can contain many vCards for multiple people. It has the advantage of actually being designed to handle that kind of information.

Otherwise CSV is definitely a good choice, since most applications allow importing from CSV. There are multiple formats, however - Google Contacts allows exporting in "Google CSV", "Outlook CSV", etc. So it would probably be best to pick one of those CSV formats - usually applications allow selecting the correct column for every attribute they import.

Most power users are happy with CSV as well, since they can import into Excel/LibreOffice and do whatever data wrangling they want from there.

djaiss:
> CSV is definitely a good choice

Alright, but how? One CSV per table?

jonasvp:
> One CSV per table?

I wouldn't stick to the database schema for that. Have one CSV export for contacts (in Google/Outlook CSV format) and add others for other things (conversations, debts, gifts) as requested. Those would need to be denormalized somewhat, so e.g. for every person that appears, show the name _and_ the ID.

Is there actual user demand for this, though? I see the point of importing contacts into another application, the other items seem somewhat Monica-specific.

asbiin:
I disagree with CSV format.
We need a format to represent object base and their relations (hierarchic-structure).
JSON or XML is very good, it can be used in a script, program, you can do queries to get a part of the file. CSV does not allow that.
Nor VCF.
We need a format to represent all objects on Monica, VCF it to restrictive here.

Eventually we can do both JSON and CSV ...

djaiss:
[...]

PR #2478, opened 2019-02-28
feat: add JSON export support

### LINKED ISSUE #32: Add export function in JSON
It would be useful to have an "export data" function that would export all the data held by an account into a JSON file. Of course this should be API-callable for automation purposes.

mechanarchy:
This PR adds support for exporting a users' data as a JSON file, and resolves #32 and relates to #929. This isn't part of the 2019Q1 roadmap but is a request with community support that has been open for some time.

The code is largely based on the existing SQL exporter and produces a JSON object mapping table name keys to arrays of value objects. Exported data comes nicely formatted with newlines and indentation, which gives greater upfront readability at the expense of slightly larger generated files. My tests with a homestead instance indicate it's 3.9MB for a formatted file vs 3.3MB for a stripped one, operating on a dataset of 164 contacts/167 activities/259 gifts. The same dataset exported as SQL is 3.3MB.

This PR is complete from my limited perspective, but there are a few notes to comment on:
* I have added four new language strings to the `en` file only. Do I need to copy these across to all other languages or will they 'fall back' to the English file if the string is missing?
* I've created a `CHANGELOG` entry, but I haven't added `changelog.json` entry.
* I'd like to deprecate and remove the old endpoint `/settings/exportToSql` in favour of `/settings/export/sql` so we can better 'namespace' any other export styles in the future.

@djaiss can I ask you to please review and provide feedback at your leisure? :smile: 

### A snippet of the exported data:
```
{
  "export_meta": {
    "format": 1,
    "username": "John Doe",
    "filename": "1916093254.json",
    "exported": "2019-02-28 07:01:34"
  },
  "activities": [
    {
      "id": 1,
      "account_id": 1,
      "activity_type_id": 5,
      "summary": "Majesty?' he asked. snip",
      "description": null,
      "date_it_happened": "2018-12-12 00:00:00",
      "created_at": "2019-02-19 00:15:45",
      "updated_at": "2019-02-19 00:15:45"
    }
  ],
  "activity_contact": [
    {
      "activity_id": 1,
      "contact_id": 1,
      "account_id": 1
    }
  ],
  "activity_statistics": [],
  "activity_type_categories": [],
  "...": "skipped a bunch here",
  "term_user": [],
  "weather": [],
  "accounts": []
}
```

### The export interface now looks like:
image


djaiss:
Thanks so much for this PR.
Great work, let me review it officially. Also, to answer your questions:

> * I have added four new language strings to the `en` file only. Do I need to copy these across to all other languages or will they 'fall back' to the English file if the string is missing?

Nope, it will fall back to other locales as well. We just need to update the English source file.

> * I've created a `CHANGELOG` entry, but I haven't added `changelog.json` entry.

You could have 😀 I mean, changelog.json is used to broadcast important changes to the users, and it's one of these changes.



> * I'd like to deprecate and remove the old endpoint `/settings/exportToSql` in favour of `/settings/export/sql` so we can better 'namespace' any other export styles in the future.

Yes!

app/Jobs/ExportAccountAsJson.php
```diff
@@ -0,0 +1,165 @@
...
+    public function handle()
+    {
+        $downloadPath = $this->path.$this->file;
+
+        /** @var User $user */
+        $user = auth()->user();
+        $account = $user->account;
+
+        $json_output = '';
+        $this->generate_json_header($json_output, $user);
```
djaiss: Could you use the camelCase for your method names? We try to use it everywhere in the repository. This apply to all methods in this PR.

mechanarchy: Thank you for the review! I have updated the function names.

djaiss:
@asbiin could you also review this one? I want to be sure, especially since it's about exporting data.

asbiin:
I don't really like the result. It's quite the same function as SQL export, only it's a json file.
A really usefull json export will get the same result as json import/export with the API. This means each table is exported with some specific format, but we cannot iterate over every table like this.

mechanarchy:
Yeah, after some more thought I agree - this PR just adds more code to support without any meaningful benefit. I still think JSON export is a nice feature to have, but without a real user story behind it it's just clutter. I'll close this PR.

Issue #2388, opened 2019-02-07
ErrorException: Trying to get property 'contact' of non-object 


justsomebody42:
**Describe the bug**
This error started to show up on my Dashboard just recently.
The list of birthdays is unchanged to when it was working.

**Which version are you using:**
Selfhosted. master, commit e9b28661

**Additional context**
I added:
`                @if (!is_object($reminderOutbox->reminder))
                    @continue;
                @endif
`
in resources/views/dashboard/_monthReminder.blade.php as a workaround.
When I printed $reminderOutbox->planned_date to see if an existing birthday was missing, I saw that the same birthday showed up twice. 
The first line without reminder object, the second line showed up fine.
Maybe this helps to find the cause...

jatls:
I experience the same bug and have used your same workaround in the meantime.

justsomebody42:
Fixed here: https://github.com/monicahq/monica/pull/2502

PR #2503, opened 2019-03-07
fix: dashboard crash when reminder is empty

justsomebody42:
Fixed that the dashboard could crash with "ErrorException: Trying to get property 'contact' of non-object"

Fixes: https://github.com/monicahq/monica/issues/2388


asbiin:
I'm not sure this could happen anymore, but the change is OK I guess.

asbiin:
Thank you for your help!

PR #2505, opened 2019-03-09
fix: fix reminders not being sent

djaiss:
I _think_ the problem that reminders are not being sent is because the job `NotifyUserAboutReminder` takes a ReminderOutbox object in parameter and triggers the different notifications to be sent by also passing the same ReminderOutbox object to the other jobs.

However `NotifyUserAboutReminder` also deletes the current ReminderOutbox object at the end of the function. Therefore once the other jobs want to process their queues, they can't find the ReminderOutbox object as it has been deleted by the parent job.

I suppose that we haven't seen this behaviour locally because queues are synchronous, whereas they are asynchronous in production.


Bounz:
Hi, @djaiss.
I updated to 2.12.1 (Cloudron installation) but reminders are still not being sent.
The only way I was able to get email was to manually change notification date in the database and then run `php artisan send:reminders`.

Issue #1077, opened 2018-03-30
Telegram prefix/Multiple handlers


yarons:
https://github.com/monicahq/monica/blob/1e1e984baf52156d1c7b6e44ba93cb89643f341d/database/migrations/2017_11_10_174654_create_contact_fields_table.php#L103

Hey guys, I'm not sure `telegram:` is the correct handler, from what I remember it's `tg:`, this is the situation in TDesktop according to this bug:
https://github.com/telegramdesktop/tdesktop/issues/2955
Maybe multiple handlers support would solve this case (`telegram://`, `tg://`, `https://t.me`).

Furthermore, there are the `http://t.me/@telegram` links that might be a better idea in this case.

I think for WhatsApp this is their URL sceme (whatsapp://):
https://www.macstories.net/tutorials/use-whatsapps-url-scheme-with-drafts-launch-center-pro-or-a-bookmarklet/

djaiss:
Thanks @yarons.

It will be pretty easy to fix this.

FranklinYu:
I would vote for t.me links, as they fallback gracefully if the browser doesn’t recognize the Telegram app. By the way, it is http://t.me/username (without the `@` sign). See official documentation here: https://core.telegram.org/api/links

PR #2562, opened 2019-04-03
fix: add default protocols to profile field links

### LINKED ISSUE #1030: Adding Contact Information gives bogus link

### LINKED ISSUE #2044: Link to Twitter profile

### LINKED ISSUE #2045: Link to LinkedIn profile

mechanarchy:
This PR should resolve #2044 and resolve #2045. I have added default protocols for Facebook, Twitter, WhatsApp and LinkedIn. I note that the VCard importer and exporter rely on fixed URL strings to set profile field links - this may be problematic if we need to change the defaults in future.

I've also **adjusted how the links are rendered** to prevent accidental internal links (resolves #1030). The new logic is:
* If a `protocol` is provided, generate a link to `protocol + data`
* Else, if `data` contains `://`, generate a link to `data`
* Else, output `data` without creating a link

This PR replaces #2475 - I don't know what I did but I trashed the branch history (thanks @tbirrell for your help!).


asbiin:
After some reconsideration:
"Protocol" is here a communication protocol, or scheme, like `mailto:`, `tel:`, or even `http:`, `ssh:`, etc.
I'm not sure it's a good idea to save the schema links `https://www.facebook.com/xxx`, `https://twitter.com/xxx` in the same place.
That could take us to a problem, if some people have already change it, and especially for Telegram: there is already a protocol `telegram:` (or `tg:` according to #1077), and there is a schema link `http://t.me/`. How to handle these possibilities?

Maybe we could create another field to save this schema link?

tbirrell:
Well, if you think about it, we are using these links as protocols. Granted they are full URIs (sans ids), but they are still used to specify what to do with the information.

Also, we will never use the https protocol without the rest of the url for FB or Twitter or LinkedIn or whatever, so there is no good reason to separate the values.

asbiin:
@mechanarchy and @tbirrell  Thanks for your help!

PR #2630, opened 2019-04-28
chore: add thecodingmachine/safe package

asbiin:


djaiss:
Hum. Is that really necessary?

asbiin:
@djaiss read this article: https://thecodingmachine.io/introducing-safe-php

PR #2652, opened 2019-05-05
fix: fix schedule run

asbiin:
Fix schedule run in case cron can't run on fix hours.
This is the case on heroku for example.


keywordnew:
Amazing! Thank you for fixing this 🎉

I'm guessing, based on the monthly pattern of releases, this will make it into 2.14.0 or 2.13.1 to be released in the next few days.

tomgs:
Still not working for me, I'm afraid:(

Issue #2513, opened 2019-03-11
Date widget sets date to one day before


mrzv:
**Describe the bug**
When filling the date manually (e.g., when entering somebody's birthday), the date in the calendar popup widget gets set to one day before what's actually entered. When submitting, it's the widget date that gets recorded. If one doesn't catch this, one gets lots of birthdays shifted by one day.

**Which version are you using:**
Self-hosted 2.12.1

asbiin:
I cannot reproduce this.
What is your locale and timezone?

mrzv:
Both client and server are set to Pacific. But I think this is entirely a client-side UI issue. Because even as I enter a given date, the date before it is highlighted in the calendar widget. The client is a Mac, using Chrome, if that helps.

tbirrell:
I'm running into the same issue. I'll set to March 31st and it'll display as March 30th (and if, as in the case of birthdays, I set a reminder, it is also off by a day). Then when I go in to edit it, it says March 30th. Upon saving _again_ it stays. I haven't been able to look into it yet, but I suspect that somewhere between PHP and the DB, a timezone conversion is breaking something.

I'm on version 2.12.0

tbirrell:
Further exploration shows that this is a discrepancy with the Vue Datepicker. If you use the popup to select the date, it is correct. If you type it in, it will actually show the previous date in the popup, then send that (incorrect) date back to the PHP.

asbiin:
This is fixed by #2655

PR #2655, opened 2019-05-06
fix: use vuejs-datepicker, not vuejs-datepicker-tmp (#2513)

SyntaxBlitz:
The [vuejs-datepicker-tmp](https://www.npmjs.com/package/vuejs-datepicker-tmp) library exists only to add "typeable" to vuejs-datepicker. It has the bug mentioned in #2513, though. Now that vuejs-datepicker includes "typeable" support (which isn't broken), we can switch over to using the non-forked version.

Thank you for your work on Monica! I just started using it today. I've never contributed to the repository, so let me know if there's anything I should be doing differently here.


asbiin:
Nice catch @SyntaxBlitz ! Thank you for your help!

Issue #589, opened 2017-09-17
After docker install getting "fatal: Not a git repository" in logs


timjrobinson:
I went through the monica docker setup guide building the docker image myself with `docker-compose build` then `docker-compose up`. After setup I saw the following error message appear whenever I visited the `/dashboard` page. 

```
fatal: Not a git repository (or any of the parent directories): .git
```

This error also appears in /var/log/apache2/error.log inside the monica docker container. 

I fixed it by ssh'ing into the monica container and setting up git in /var/www/monica with:

```
/var/www/monica # git init
/var/www/monica # git remote add origin https://github.com/monicahq/monica.git
/var/www/monica # git reset --hard origin/master
```

This stopped the error from appearing, but I'm not sure where it was coming from in the first place. Is Monica inside the docker container supposed to be a git repository? Or is this some bug where the code assumes it's a git clone but that's not true when doing the docker setup so it freaks out?

kofiav:
Encountering the same issue, tried doing what @timjrobinson did, but now my dashboard does not work.

timjrobinson:
My dashboard isn't working at the moment either :/

RogerSik:
Removing .git from .dockerignore and then rebuilding the image fixed this error for me.

djaiss:
Was this solved?

mgolisch:
works for me without that error.
probably due to #650 
the artisan commands seem to expect a git reposiroty there

asbiin:
Linked to #872, but it's just a warning message, app is running well.

PR #2662, opened 2019-05-09
fix: change cwd before running 'exec' in sentry config

SyntaxBlitz:
Related to #589: when running artisan from outside the git repository (e.g. when running from [cron](https://github.com/monicahq/monica/blob/master/docs/installation/ubuntu.md#3-configure-cron-job-optional)), we get `fatal: Not a git repository (or any of the parent directories): .git`. If we first `chdir` into the script directory, we can run `git` from `exec` the way we want to.


asbiin:
I don't really agree with that.
In production mode, you may not need this, because you would use 'monica:update' command, or 'config:cache' which cache the config, and the problems with cron does not appear.

At least we could chdir inside the 'exec' command?

SyntaxBlitz:
That makes sense, though in that case I would argue that `config/sentry.php` shouldn't try to find the release automatically, but should require that the `.sentry-release` file already exist from an earlier working `sentry-set` call (and die with an informative error otherwise). Since this script should be runnable from outside the git repository (e.g. from cron), it's not very friendly for the script to contain the `git log` functionality that can throw an error when used like this.

I agree that it makes sense to put the directory change inside `exec`. It does work, but I chose not to do this because I was worried about escaping:

    exec('cd '.__DIR__.' && git log --pretty="%h" -n1 HEAD')

I'm not sure if \_\_DIR\_\_ can contain characters that would interfere with this script, but it seems like bad practice to concatenate directly into the `exec` call like that. It looks like we could use `escapeshellarg` for this, though I'm hesitant to say for sure because I don't have a whole lot of experience with PHP's `exec`-related functions.

asbiin:
Closing this because it's not a desire feature.
Use `monica:update` or `config:cache` on production mode, so this command is never run in a cron task.

Issue #2668, opened 2019-05-11
Produce different Logo size


asbiin:
Our great logo is available [here](https://app.monicahq.com/img/monica.svg).

There is a [140px version](https://app.monicahq.com/img/monica_140.svg), but it's buggy:
![image](https://user-images.githubusercontent.com/25419741/57569266-a17b4100-73f2-11e9-8124-24a793f77583.png)
Could you fix it?

The webmanifest file should include a 192px and 512px icon. See [this doc](https://developers.google.com/web/tools/lighthouse/audits/install-prompt).
We should provide a 192px at least.

tbirrell:
The bug is in the open line of the SVG file. You want to use `height/width` to scale the image, and leave `viewbox` alone. The `viewbox` is the canvas (per se), so you want to scale the canvas rather then resize it.

`<svg width="140" height="140" viewBox="0 0 500 500" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">`

PR #2682, opened 2019-05-15
fix: add resized logos and fix scaling

tbirrell:
Per #2668, I have fixed the bug with the 140px logo, and added 192px and 512px logos as independent files.


asbiin:
Thanks @tbirrell !

PR #2707, opened 2019-05-20
fix: Show full name in dashboard reminders(fixes #2608)

### LINKED ISSUE #2608: Ability to see full names instead of just initials
The dashboard is great for showing when there are upcoming events, particularly birthdays, but I struggle to figure out who people are from just their first name and an initial of their last name. I seem to recall this format being used in other places but can't find an example.

It'd be great if there was a way to select the format of names one wants to see across the entire app. Maybe some people would just like initials for first and last, maybe some people would like last name then first name, etc. Personally I'd like to see full names for everyone wherever they are used.

Thanks for your consideration :)

Brice187:
tiny fix, just changed the function in html template


resources/views/dashboard/_monthReminder.blade.php
```diff
@@ -12,7 +12,7 @@
                     @if ($reminderOutbox->reminder->contact->is_partial)
 
                         @php($relatedRealContact = $reminderOutbox->reminder->contact->getRelatedRealContact())
-                        <a href="{{ route('people.show', $relatedRealContact) }}">{{ $relatedRealContact->getIncompleteName() }}</a>
+                        <a href="{{ route('people.show', $relatedRealContact) }}">{{ $relatedRealContact->getNameAttribute() }}</a>
```
tbirrell: Using a [mutator](https://laravel.com/docs/5.8/eloquent-mutators#accessors-and-mutators) means that the method `getNameAttribute()` allows you to call `$relatedRealContact->name;` to get the name.
```suggestion
                        <a href="{{ route('people.show', $relatedRealContact) }}">{{ $relatedRealContact->name }}</a>
```

asbiin:
Closing now: you should consider adding an option to let the user choose the UI.

PR #2735, opened 2019-05-30
chore: reduce js files size

asbiin:


asbiin:
Reduce js files from 4128 kB to 3904 kB

Issue #2843, opened 2019-07-20
Hide "Stay in touch" for deceased contacts


GeorgeHahn:
The contact page shows a 'Stay in touch' button/reminder. It would be nice to hide this button for deceased contacts.

Current UI (Version: 2.14.0):

![image](https://user-images.githubusercontent.com/502747/61583437-53d43080-ab27-11e9-9c48-567dc0176192.png)

tomgs:
I am sorry for your loss, and this is a nice feature to have.

virtadpt:
...this.  Please?

asbiin:
Fixed by #2845

PR #2845, opened 2019-07-22
feat: hide stay-in-touch for deceased contacts

tbirrell:
A quick UI fix for #2843.


asbiin:
Thank you @tbirrell 
So this only hides the "stay-in-touch" form for deceased contacts.
Bu why if I have a normal (alive) contact with stay-in-touch activated, and then I switch it to "deceased"? I will not be able to remove the stay-in-touch then. I could switch it to "not dead" and then remove the stay-in-touch, but this is not realy simple.

tbirrell:
I've added some logic to automatically disable StayInTouch when a contact is marked deceased.

asbiin:
@tbirrell Great job, thank you!

PR #2884, opened 2019-08-09
fix(lang): upaded German translation for users_blank_title

zuphilip:
🇺🇸 You -> 🇩🇪 Du


asbiin:
Hi @zuphilip . Thanks for your contribution, but translation is handled only in crowdin. Please use it to change the translation.
See https://github.com/monicahq/monica/blob/master/docs/contribute/translate.md

zuphilip:
Done. You or someone need to look at it now: https://crowdin.com/translate/monicahq/14/en-de#503 . There are also more translations in German to review.

PR #2958, opened 2019-08-30
feat: better age display

### LINKED ISSUE #2956: Approximate Age
When you enter an approximate age, it presents the birthday as 1/1/19xx and at first glance appears to be an accurate birthday.

It'd be helpful to allow for an approximate age that's present like "approx 35" or "mid-30s."  The vagueness of the age might help since I assume this is just to be able to identify the person when there is no photo (that's what I would find it useful for).

tbirrell:
#2956 made a good point about approximate ages displaying funny. So I fixed that. As well as year-unknown dates having an empty age parenthesis.

If age approximate: `around 35 years old`

If age almost (year-unknown): `Jan 1`

If age exact: `Jan 1, 1984 (35)`

This fix #2956


asbiin:
@tbirrell thank you ! great job !

PR #2982, opened 2019-09-04
fix: push footer to bottom regardless of content height

shroudedcode:
By default the footer is always directly below the page content. This is not an issue if the content is taller than the screen but looks weird if it's not. This change ensures that the footer is always at the bottom of the page:


asbiin:
I'm not sure if it's right or not ... @djaiss what do you think?

djaiss:
This is awesome. Sorry for the very late reply. Could you resolve the conflicts so we can merge this one?

Issue #2951, opened 2019-08-28
Age calculation wrong for deceased person with unknown birthday


sid3windr:
**Describe the bug**
I have a contact which is deceased. The birth year is unknown (but the birthday is not). The date the person died is known. Monica correctly shows the person is deceased, but calculates an age at death which is impossible to know because the birth year is unknown.

I'm not sure where the age 4 comes from... (cf. screenshot)

**Screenshots**
![image](https://user-images.githubusercontent.com/1889999/63874104-2d74a100-c9c1-11e9-8189-faff22af886d.png)

![image](https://user-images.githubusercontent.com/1889999/63874415-c86d7b00-c9c1-11e9-9937-2cb81ac6ba0b.png)

**Which version are you using:**
- Selfhosted latest git as of right now when writing this bug report

asbiin:
@tbirrell thank you for the research on a solution.

PR #2983, opened 2019-09-04
fix: fix broken age calculation for deceased contacts

tbirrell:
This commit fixes the Contact Profile page for deceased contacts as noted in #2951. 

Additionally, I fixed a typo in the method that calculates `Age at Death`


asbiin:
Thank you @tbirrell !!

PR #3011, opened 2019-09-15
fix: center image in pop-up on contacts page

andreasjacobsen93:
When opening an image on the contact page, the image is aligned left. This PR centers the image.


asbiin:
Thank you again @andreasjacobsen93

PR #3012, opened 2019-09-15
fix: hide completed tasks on dashboard

andreasjacobsen93:
All tasks are shown on the dashboard, even completed tasks. This can get very cluttered with many tasks.

I think the dashboard should be a more clean up to date view, so it should only show tasks not completed yet.


asbiin:
Thank you @andreasjacobsen93

PR #3025, opened 2019-09-18
fix: reschedule missed stay-in-touch

asbiin:
Some times the StayInTouch trigger is missed.
This fix the problem in rescheduling the StayInTouch as it was not previously missed.

### DIFF
--- a/tests/Unit/Jobs/ScheduleStayInTouchTest.php
+++ b/tests/Unit/Jobs/ScheduleStayInTouchTest.php
@@ -88,4 +88,35 @@ class ScheduleStayInTouchTest extends TestCase
             'stay_in_touch_trigger_date' => '2017-01-01 07:00:00',
         ]);
     }
+
+    public function test_it_reschedule_missed_stayintouch()
+    {
+        NotificationFacade::fake();
+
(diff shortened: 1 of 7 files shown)


tests/Unit/Jobs/ScheduleStayInTouchTest.php
```diff
@@ -88,4 +88,35 @@ public function test_it_doesnt_dispatches_an_email_if_free_account()
...
+            'timezone' => 'America/New_York',
+        ]);
+
+        dispatch(new ScheduleStayInTouch($contact));
+
+        NotificationFacade::assertNotSentTo($user, StayInTouchEmail::class);
+        NotificationFacade::assertNothingSent();
+
+        $this->assertDatabaseHas('contacts', [
+            'stay_in_touch_trigger_date' => '2019-01-26 07:00:00',
```
djaiss: If the stay in touch frequency is set to 30, why is the date the 26th?

asbiin: because the frequency is not 'one month', it's '30 days' ...

PR #3080, opened 2019-10-12
feat: fix pluralization in vue-i18n

asbiin:
This adds pluralization form for vue-i18n
See https://kazupon.github.io/vue-i18n/guide/pluralization.html#accessing-the-number-via-the-pre-defined-argument


resources/assets/js/pluralization.js
```diff
@@ -0,0 +1,42 @@
...
+  number = Math.abs(number);
+  number = (number == 1) ? 0 : (((number >= 2) && (number <= 4)) ? 1 : 2);
+  return Math.min(number, choicesLength - 1);
+}
+function pluralD (number, choicesLength) {
+  number = Math.abs(number);
+  number = ((number % 10 == 1) && (number % 100 != 11)) ? 0 : (((number % 10 >= 2) && (number % 10 <= 4) && ((number % 100 < 10) || (number % 100 >= 20))) ? 1 : 2);
+  return Math.min(number, choicesLength - 1);
+}
+function pluralE (number, choicesLength) {
```
yarons: This is the plural form for Hebrew, needs conversion to the correct form:
`nplurals=4; plural=(n == 1) ? 0 : ((n == 2) ? 1 : ((n > 10 && n % 10 == 0) ? 2 : 3))`

yarons:
Requires some additional adjustments so it'll work with Hebrew.

asbiin:
Thank you @yarons! Can you provide a PR for it? I took the function from Laravel... If there is a problem, maybe we could fix it too 😉

resources/assets/js/pluralization.js
asbiin: @yarons Could you check these source too?
- http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html
- https://github.com/laravel/framework/blob/master/src/Illuminate/Translation/MessageSelector.php

PR #3087, opened 2019-10-15
fix: update journal date on activity update

### LINKED ISSUE #2993: Journal entry date does not match edited activity date
**Describe the bug**
When creating an activity, it defaults to today's date. If you later edit the date on the activity, it will appear correctly on the contact page. However, the automatically created Journal entry is not updated correctly. It will display the edited date in the calendar icon, but it will be sorted into the list at the wrong position (ie, 'today' at the time it was made).

I am using a server I maintain myself, version 2.14.0 from the docker image.

**Steps to reproduce**
1. Create a Contact, eg "John Doe"
1. On John Doe's page, create an activity and fill in the required fields.. Leave the date set at default (it should be the current date). Save the activity.
1. Edit the activity, and change the date to be at some point in the past.
1. Go to your Journal view.

Expected result: The linked Journal Entry should appear in the list at the correct date.

**Screenshots**
Contact page view (after editing the date):
image

Journal view (after editing the date). Note that the "Monday 22nd" text is correct, but its position in the journal is wrong (it should be in July but it is in September).
image

**Additional info**
Here's a sanitised snippet from the `/journal/entries` API response:
```json
{
  "id": 123456,
  "date": {
    "date": "2019-09-07 00:00:00.000000",
    "timezone_type": 3,
    "timezone": "UTC"
  },
  "journalable_id": 123456,
  "journalable_type": "App\\Models\\Account\\Activity",
  "object": {
    "type": "activity",
    "id": 123456,
    "activity_type": null,
    "summary": "Second interview",
    "description": null,
    "day": 22,
    "day_name": "Mon",
    "month": 7,
    "month_name": "JUL",
    "year": 2019,
  },
  "show_calendar": false
}
```

LukasRos:
Note: I only changed this for the UI controller, not the API controller because the API controller already has this (however it is implemented differently, by deleting and recreating - I'm simply updating here).

Fix #2993


asbiin:
Thank you @LukasRos !

PR #3103, opened 2019-10-19
docs: add heroku sql import instructions + db cleanup script

tomgs:
A few notes, as to why this is WIP:

1. When actually trying to log into my docker heroku app I see: `Trying to get property 'latest_version' of non-object`. Note that I attempted to import data from an old version of Monica to a new one.
2. Which begs the question: what happens if I export from an older version of Monica and import to a new one? The SQL breaks and gives errors like in the bottom of my PR. This is supposedly addressable by migrations, but I'm not sure what's the best way to attack it. It's also possible to just instruct users to first update to the latest version of Monica and then do the export, which I did in this doc.
3. Wasn't aware of the commit guidelines, and thus didn't follow the commit guidelines, sorry about that!

BTW, I still have the dynos up if any one of you is up for some investigation.


tomgs:
@asbiin btw, this is how I got to the issue of the reninders, while trying to fix this up.

tomgs:
@djaiss @asbiin Can we get this merged when you have a moment? Docs mostly.

asbiin:
thank you @tomgs ! It ihas taken a long time, but this is now merged.

Issue #3074, opened 2019-10-08
Avatars in /journal entries missing [v2.15.2]


kotnik:
**Describe the bug**

I visit `/journal` and little contact images are missing. Contacts have working avatars, changing them does not make them appear in Journal. Images from `/journal/entries` where the frontend is fetching entries from are just fine, I can open `object.attendees[].avatar.url` in browser just fine.

**Screenshots**

![](https://bin.kotur.org:9441/UKuBRAqnwYw7Q2.png)

**Which version are you using:**

 - A server I maintain running v2.15.2

**Additional context**

There are no errors in logs.

baptistemarchand:
This seems fixed. Can we close the issue?

PR #3116, opened 2019-10-22
fix: correct reference to journal entry avatars

chrisforrence:
Fixes the issue reported in #3074.


asbiin:
Great job! Thanks @chrisforrence

Issue #3104, opened 2019-10-19
Add CSV + JSON Data Export Format


tomgs:
While SQL can be considered quite universal and time-proof, it also requires a database and is heavily geared towards the schema that is used in said database.

If there was a possibility to export the data into a more universal format, i.e. JSON or CSV, in a way that's more "aggregated" (i.e. a single row in the CSV or a single object in the JSON represent all that the system knows about a person) that would be more future-proof, in my opinion.

It shouldn't be too hard to implement, since it's basically just querying around, some massaging of the data, and then dumping into a file.

By no means I mean you should do it though:) This is more a note to myself.

mattdp:
@tomgs I'm also thinking about preserving this data on a long timescale, so I'd support this addition. If you ever seriously consider taking this on, I'd be happy to take a first draft at what the export should include/what massages would be best. I'm sure there's a lot of edge cases that would benefit from multiple thinkers.

tomgs:
Totally agree. If you have some ideas just jot them down here, and we'll
properly spec them out later on.

tomgs:
@mattdp I've invited you to a private repo with my contact details. Just ping me there and we'll catch up!

tomgs:
Alright, Vagrant box being set up right now on my local machine. Since I'm not entirely sure what's the best approach for this, I'm going to hack out a first thing that does a JSON export in a really simple way:  I will simply be an array of contact objects, with just the first name, last name, and gender of the person. Once there's a function that does this, it should not be very difficult to incrementally add more and more fields until we have the whole thing ready. @mattdp If you're for pushing this through thorough an actual specs process, I'm game, just say the word.

tomgs:
I will work of the #3119 branch, and try and add a flag/option for JSON with just the barebones format. @djaiss if you can not merge it tonight (at least until I'm through with it) that would be awesome.

tomgs:
This is now on hold until I get my box up and running. While I can get the Vagrant box up, I can't seem to have the folder sync back to my machine, so I can actually edit the files and make commits to my fork. Will look into it soon.

Issue #2937, opened 2019-08-25
Unable To Export Data


pqhf5kd:
**Describe the bug**
I'm trying to export my data (just to keep a backup) using the Export to SQL button from the hosted MonicaHQ instance but I'm just getting a blank page. Some weeks back I was getting an error but I didn't submit a ticket at the time.
 
**Screenshots**
Clicking the button beings to load https://app.monicahq.com/settings/exportToSql, it finishes and literally a blank page is returned (no source code)

mattdp:
@asbiin  I am experiencing this issue as well, and it makes me nervous about continuing to put data into the system. Thank you for all your help building Monica!

DieselDuz42:
I too am experiencing this issue. 

I dropped an email Sept 9th and have gotten no response.

tomgs:
This is fixed by now, at least on 2.4.* and 2.5.* which I ran today.

mattdp:
@tomgs I don't believe this is working - I tried on PC Firefox and Mac Chrome. I have 2K+ contacts so I would guess it's timing out for people with large databases and working for smaller ones.

@asbiin I'd be happy to investigate further, but don't know much about PHP or how projects are structured - can you point me in the right direction as to where this function is located in the codebase, and I'll try to do some research from there?

tomgs:
@mattdp Ah, sorry about that, my database is indeed smaller than yours. 
Per your question: 

The route that specific settings pages sits on is `/settings/export`, and it is redirected in the router here:
https://github.com/monicahq/monica/blob/c90354c07aa1e302405d6e224c4281e419fa2f09/routes/web.php#L238

The settings controller is defined here:
https://github.com/monicahq/monica/blob/10e73e35b6160bf8a427ddfabb958970297cfb45/app/Http/Controllers/SettingsController.php#L36

If I'm not mistaken, than this is the thing that actually triggers the correct view inside that file:
https://github.com/monicahq/monica/blob/10e73e35b6160bf8a427ddfabb958970297cfb45/app/Http/Controllers/SettingsController.php#L236-L243

This is that view:
https://github.com/monicahq/monica/blob/master/resources/views/settings/export.blade.php#L1-L51

And this is the button that triggers what is eventually giving you an empty page:
https://github.com/monicahq/monica/blob/master/resources/views/settings/export.blade.php#L41

If I'm not mistaken, this is where the export gets triggered:
[...]

PR #3119, opened 2019-10-23
feat: refactor the export process

djaiss:
This PR tries to refactor the export process which currently doesn't work in production, for an unknown reason (but I believe this is because of some sort of memory limitation).

This new way of exporting is way different that what we used to do. Before we were simply looping over all the tables (except some of them defined in an array at the top of the file). Now we are calling explicitly which tables we would like to export.

The file is pretty huge, around 1300 lines, but it lets us control exactly what we want to export, and what we want to avoid to export --- something we didn't have in the past.

Locally it works fast. I don't know if it will work in production, we have to merge to see it.


tomgs:
This is being done (if I understand correctly) to alleviate https://github.com/monicahq/monica/issues/2937. In addition, as a follow up, I'd like to have an option to not be constrained by the schema of the database, and have a more "massageable" format to work with. I was thinking JSON/CSV - see the discussion on https://github.com/monicahq/monica/issues/3104.

phpstan.neon
```diff
@@ -38,3 +38,4 @@ parameters:
         - */app/Helpers/ComposerScripts.php
         - */app/Http/Controllers/DAV/Backend/CardDAV/AddressBookHome.php
         - */app/Providers/BroadcastServiceProvider.php
+        - */app/Services/Account/Settings/ExportAccount.php
```
asbiin: This is bad!

app/Services/Account/Settings/ExportAccount.php
```diff
@@ -0,0 +1,1366 @@
...
+use App\Models\User\User;
+use App\Services\BaseService;
+use App\Models\Account\Account;
+use App\Models\Contact\Document;
+use Illuminate\Support\Facades\DB;
+use App\Exceptions\NoAccountException;
+
+class ExportAccount extends BaseService
+{
+    protected $sql;
```
asbiin: There are 2 problems here:
- first you should not use a class variable here. It's better to use local variable in `execute` function and pass it to each functions.
- this could link to a memory limit problem, because this `$sql` variable could be huge. We should find a way to write each line in a file directly, so there is no risk of memory limit.

app/Services/Account/Settings/ExportAccount.php
```diff
@@ -0,0 +1,1366 @@
...
+        $this->exportReminderRule($data);
+        $this->exportReminderSent($data);
+        $this->exportReminder($data);
+        $this->exportSpecialDate($data);
+        $this->exportTag($data);
+        $this->exportTask($data);
+        $this->exportTermUser($data);
+        $this->exportUser($data);
+        $this->exportWeather($data);
+        $this->exportContactPhoto($data);
```
asbiin: How are we sure this is the right order of recreating(populate) each table?

app/Services/Account/Settings/ExportAccount.php
```diff
@@ -0,0 +1,1366 @@
...
+
+    private function buildInsertSQLQuery(string $tableName, string $foreignKey, array $columns, array $data)
+    {
+        $accountData = DB::table($tableName)
+            ->select($columns)
+            ->where($foreignKey, $data['account_id'])
+            ->get();
+
+        if (! $accountData) {
+            throw new NoAccountException();
```
asbiin: When this could happen?

app/Services/Account/Settings/ExportAccount.php
```diff
@@ -0,0 +1,1366 @@
...
+        $listOfColumns = $columns;
+        foreach ($listOfColumns as $key => $value) {
+            $listOfColumns[$key] = '`'.$value.'`';
+        }
+        $listOfColumns = implode(',', $listOfColumns);
+
+        foreach ($accountData as $singleSQLData) {
+            $columnValues = [];
+
+            $this->sql = $this->sql.'INSERT IGNORE INTO '.$tableName.' ('.$listOfColumns.') values (';
```
asbiin: We may (?) add the prefix table in here.

phpstan.neon
djaiss: Yeah but try removing it and see how it triggers a really, really stupid error as it doesn't understand the context.

app/Services/Account/Settings/ExportAccount.php
djaiss: I've thought about that.
Let me check how I could enhance it.

app/Services/Account/Settings/ExportAccount.php
djaiss: The order doesn't matter as I've disabled the foreign constraints at the beginning of the file. If I remove it, the import is simply impossible due to how we've initially structured the database.

app/Services/Account/Settings/ExportAccount.php
djaiss: ohhh right yes you are right.

djaiss:
> This is being done (if I understand correctly) to alleviate #2937. In addition, as a follow up, I'd like to have an option to not be constrained by the schema of the database, and have a more "massageable" format to work with. I was thinking JSON/CSV - see the discussion on #3104.

Yes exactly. I also agree that we should allow other formats, but in that case, I’m not sure we should try to create a service that does both exports, as each one will be very specific. The drawback of having two services, though, is that we need to duplicate the work in that case.

app/Services/Account/Settings/ExportAccount.php
djaiss: When I export a sample account (using `artisan setup` command), the SQL can go up to 3.5Mb of data. It won't be good in production on large accounts.

tomgs:
> > This is being done (if I understand correctly) to alleviate #2937. In addition, as a follow up, I'd like to have an option to not be constrained by the schema of the database, and have a more "massageable" format to work with. I was thinking JSON/CSV - see the discussion on #3104.
> 
> Yes exactly. I also agree that we should allow other formats, but in that case, I’m not sure we should try to create a service that does both exports, as each one will be very specific. The drawback of having two services, though, is that we need to duplicate the work in that case.

I'm totally in favor of two different services. Let's make it three actually - one for JSON, one for CSV and the one you have for SQL. It will give me a nice intro to the project + it really does make sense to separate concerns in that regard.

djaiss:
> one for JSON

Why JSON though?

djaiss:
@asbiin now the service writes to a temp file directly as the script runs. Locally it’s really fast, I wonder how it will run in production. Can you check the PR again?

app/Services/Account/Settings/ExportAccount.php
```diff
@@ -0,0 +1,1418 @@
...
+        }
+
+        foreach ($contacts as $contact) {
+            $photos = DB::table('contact_photo')
+                ->where('contact_id', $contact->id)
+                ->get();
+
+            foreach ($photos as $photo) {
+                $sql = 'INSERT INTO contact_photo (contact_id, photo_id, created_at, updated_at) values (';
+                $sql .= $photo->contact_id.','.$photo->photo_id.",'".$photo->created_at."','".$photo->updated_at."');".PHP_EOL;
```
asbiin: Just need to finish this instructions. Why not using `$this->buildInsertSQLQuery` here?

app/Services/Account/Settings/ExportAccount.php
```diff
@@ -188,7 +189,7 @@ private function buildInsertSQLQuery(string $tableName, string $foreignKey, arra
     private function writeToTempFile(string $sql)
     {
         Storage::disk('local')
-            ->append($this->tempFileName, $sql);
+            ->append($this->tempFileName, $sql, '');
```
djaiss: Why do you keep adding this? The `append` method only takes two arguments!

djaiss: @asbiin my comment above

asbiin: The 3rd argument is a prepend string to insert between the existing string in the file and the appended content.
See https://github.com/laravel/framework/blob/5.8/src/Illuminate/Filesystem/FilesystemAdapter.php#L305

PR #3126, opened 2019-10-23
feat: export data as json

tomgs:
Relevant issue: https://github.com/monicahq/monica/issues/3104


asbiin:
I think this one is not relevant ;)

Issue #2835, opened 2019-07-14
Include User Story Tests as Part of Cyprus Tests


Mohammed90:
**Is your feature request related to a problem? Please describe.**

It's not as much a feature for the end-user besides stability. It's an addition to the code-base that'll bring peace of mind and increases the robustness of the application.

The 2 issues I filed recently (#2832 & #2833) were found as I was trying to setup my instance. I simply followed the standard process in setting up any new account on any service (signup, validate my settings, enable 2FA, etc.). They could have been caught before I found them.

**Describe the solution you'd like**

The current Cyprus tests are limited to testing a single use-case (signup, login). Adding more comprehensive tests that cover how a normal user would use various features of the application (e.g. signup, edit a few settings, enable 2FA, download recovery codes, navigate to dashboard, write a journal entry, then logout) will help uncover any failures that will affect the onboarding and the day-to-day usage experience.

**Describe alternatives you've considered (optional)**

🤷‍♂ 

**Additional context**

This is just a suggestion. I like this application as a concept, and just now have I got around to set one up for myself. I've played with it a bit before to figure out how I'd like to configure my instance before actually deploying it 😃

PR #3152, opened 2019-10-31
test: setup cypress

asbiin:
Fix cypress run and add it to the pipeline.

### List elements
When we deal with list elements, cypress cannot now the id of each item
- when a list is displayed
- when an object is added to the list

The solution is to add the id of every items in a `v-cy-items` attribute:
```html
<div v-cy-name="activities-body" v-cy-items="activities.map(c => c.id)">
  <div v-for="activity in activities" :key="activity.id" class="ba br2 b--black-10 br--top w-100 mb2" v-cy-name="'activity-body-'+activity.id">
...
  <a href="" class="pointer" v-cy-name="'edit-activity-button-'+activity.id" @click.prevent="$set(activity, 'edit', true)">{{ $t('app.edit') }}</a>

```

The test looks like:
```js
cy.get('[cy-name=activities-body]').should('be.visible')
  .invoke('attr', 'cy-items').then(function (item) {
    cy.get('[cy-name=edit-activity-button-'+item+']').click();
   ...
  })
```
the `item` var will contains the value of the `cy-items` attribute (here there is only 1 item)

Note that the `v-cy-items` and `v-cy-name` predicates does not do anything in production. On other environments, they add an attribute with the name of the predicate.


Mohammed90:
I guess this addresses #2835, right?

asbiin:
@Mohammed90 well, it's related to, but it does not solve this issue.
Here I would like to run cypress tests in the tests pipeline, and that's all.

resources/js/components/people/Tasks.vue
```diff
@@ -31,7 +31,7 @@
       </div>
 
       <!-- LIST OF IN PROGRESS TASKS -->
-      <ul>
+      <ul cy-name="tasks-body" :cy-items="inProgress(tasks).map(t => t.id)">
```
djaiss: I don't understand what's this does.

tests/Commands/SetupFrontEndTestUserTest.php
```diff
@@ -0,0 +1,26 @@
...
+use Tests\TestCase;
+use App\Models\User\User;
+use App\Models\Account\Account;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+
+class SetupFrontEndTestUserTest extends TestCase
+{
+    use DatabaseTransactions;
+
+    public function test_it_create_a_test_user()
```
djaiss:
```suggestion
    /** @test */
    public function it_creates_a_test_user()
```

tests/cypress/integration/auth/signup_spec.js
```diff
@@ -25,30 +23,21 @@ describe('Signup', function () {
   //   cy.contains('Login to your account')
   // })
 
-  it('should block registration if policy is not accepted', function () {
-    cy.visit('/register')
+  //it('should block registration if policy is not accepted', function () {
+  //  cy.register(faker.name.firstName(), faker.name.lastName(), faker.internet.password(), faker.internet.email(), false);
+  //  cy.get('.alert').should('exist');
```
djaiss: Is it commented on purpose?

resources/js/components/people/Tasks.vue
asbiin: this allow cypress to know the id of the items inside the iteration (it unlikely '1').
I just though it's kind a problem to output this in production ...

resources/js/components/journal/JournalList.vue
```diff
@@ -13,12 +13,14 @@
       <journal-rate-day @hasRated="hasRated" />
 
       <!-- Logs -->
-      <div v-for="journalEntry in journalEntries.data" :key="journalEntry.id" class="cf" :cy-name="'entry-body-' + journalEntry.id">
-        <journal-content-rate v-if="journalEntry.journalable_type == 'App\\Models\\Journal\\Day'" :journal-entry="journalEntry" @deleteJournalEntry="deleteJournalEntry" />
+      <div v-if="journalEntries.data" cy-name="journal-entries-body" :cy-items="journalEntries.data.map(j => j.id)" :cy-object-items="journalEntries.data.map(j => j.object.id)">
```
djaiss: I really don't understand the need of both cy-items and cy-object-item, and I don't understand the map thing.

djaiss: Alright I saw your comment about the map thing below. But why still need to cy data attribute here?

.azure/job-test-cypress.yml
```diff
@@ -0,0 +1,51 @@
...
+      echo "REQUIRES_SUBSCRIPTION=true" >> .env
+  - template: step-composer-install.yml
+  - template: step-prepare-db.yml
+  - template: step-seed-test-db.yml
+
+  - template: step-cache-yarn.yml
+  - script: yarn inst
+    displayName: yarn install
+
+  - script: npm run cy:verify
```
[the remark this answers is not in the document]
asbiin: cypress needs an installation before running.
See https://docs.cypress.io/guides/guides/command-line.html#cypress-verify

resources/js/components/journal/JournalList.vue
asbiin: I've added a description in the PR to explain this

tests/cypress/integration/auth/signup_spec.js
asbiin: I think it's because it's not working in cypress

PR #3158, opened 2019-11-03
feat: enhance user invitation mail

asbiin:
Remove the old invitation mail template


djaiss:
@asbiin why did you change the default admin password?

asbiin:
@djaiss because the password policy is: "min:6" (and we should change that too)

PR #3173, opened 2019-11-08
docs: debian instructions updated for buster

RussellAult:
Just finished a Debian Buster installation and thought I'd update the installation docs based on my experience:
- Removed installing unused utilities
- Changed all php packages to php* (as opposed to php7.3*) to aid with future system upgrades
- Removed php packages that are dependencies of other packages
- Composer installation is now through apt
- use crontab instead of adding a file to cron.d (which, technically, should be left for packages)
- Removed everything related to php-fpm (unless I'm really missing something there's no reason not to use the default libapache2-mod-php7.3)


asbiin:
Thank you @RussellAult !

RussellAult:
You're welcome!

PR #3203, opened 2019-11-14
fix: fix reminders API GET requests

### LINKED ISSUE #2557: Bug in reminders API endpoint
Getting reminders via. API has an error in date formatting. 
API endpoint responds as below,

```json
{
    "message": "Unexpected data found.\nUnexpected data found.\nUnexpected data found.\nData missing"
}
```

##### Important
- Initially, I thought, the error is in my hosted version but tested on https://app.monicahq.com and it occurs there too.

##### Steps to reproduce
- Create `PersoalAccessToken` from https://app.monicahq.com/settings/api (or make sure you are able to access the API with other methods)
- I am using **PersonalAccessToken**
- Visit `api/reminders` endpoint with your accessToken.
- You should get above error(If you have at least 1 reminder in future) I have just 1 reminder.
- Try to access other resources e.g. `api/contacts`, that works perfectly.
- A stack trace is attached to this issue. 

 - Hosting version 2.12.1 (app.monicahq.com)
 - My server-hosted version 2.12.0

**Additional context**
[monica-api-stacktrace.log](https://github.com/monicahq/monica/files/3028901/monica-api-stacktrace.log)

mechanarchy:
This PR will fix #2557 . As identified by @vrajroham, the `delible` field is of `bool` type but it was being decoded as a datetime in the API. I've changed the code so it uses the same type of boolean conversion as other API types.

I've also tweaked the `test_reminders_get_all()` function in `tests/Api/ApiReminderControllerTest.php` to ensure that `delible=False` is tested - I hope this is acceptable. Before fixing the controller, the test failed (as expected, given the bug reports); after fixing the controller, the test passes and I am also able to get results from the API from an external script.


asbiin:
@mechanarchy Thank you! Great catch!

PR #3212, opened 2019-11-14
feat: console command to send test email

### LINKED ISSUE #3128: Add a console command which sends a test email
I propose adding a new console command which would send a basic email to a specified user account. The purpose of this feature is to aid in debugging email/SMTP configurations in the self-hosted application. The command should cause an email to be sent regardless of any app state (eg: [Time of the day reminders should be sent](https://github.com/monicahq/monica/issues/1020#issuecomment-374817879)).

Perhaps it should accept the user's ID as an argument as a means of obtaining the `to` email address.

### LINKED ISSUE #3890: command to test emails
as a sysadmin I need a command to test that monica can send emails. please add an artisan command to send a test email y debug possible problems

mechanarchy:
This PR will close #3128, it will also close #3890. It adds the `monica:test-email` command, to test whether email delivery is working. Prior to this, I had to repeatedly attempt to send password reset emails for testing (as did user @theexplosivegroup).

The PR contains hard-coded strings: personally, I'm not sure there is value in translating them because they are strictly for testing purposes, and the content is largely irrelevant (you only care if you get the email).

Use as follows:
```
php artisan monica:test-email --email=my_email@provider.com
```
If no email is provided as an argument, the command will prompt you for one, but it does not verify that the email belongs to any registered user.


mechanarchy:
This fixes #3890

mechanarchy:
This PR will close #3890

asbiin:
@mechanarchy is it possible you write a test for the command?

mechanarchy:
Sorry it took me a while - I haven't done a lot of PHP. But there are some basic tests now! Thanks for the approval :)

asbiin:
Really great job @mechanarchy  ! Thanks a lot

app/Console/Commands/SendTestEmail.php
```diff
@@ -0,0 +1,62 @@
...
+        if (! $email) {
+            $email = $this->ask('What email address should I send the test email to?');
+        }
+
+        // only perform basic validation, the mail server can handle the rest
+        if (strpos($email, '@') === false) {
+            $this->error('Invalid email address - missing "@" symbol!');
+
+            return -1;
+        }
```
asbiin, line 45: You could use `filter_var` - see https://www.php.net/manual/en/filter.examples.validation.php
```suggestion
        if (filter_var($email, FILTER_VALIDATE_EMAIL) === false) {
            $this->error("Invalid email address: '$email'.");

            return -1;
        }
```

app/Console/Commands/SendTestEmail.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $this->info("Preparing and sending email to '".$email."'");
```
asbiin, line 47: No need for concatenate here:
```suggestion
        $this->info("Preparing and sending email to '$email'");
```

app/Console/Commands/SendTestEmail.php
mechanarchy: I've done this, but it fails Psalm for `PossiblyInvalidCast` `non-empty-array<array-key, mixed> cannot be cast to string (see https://psalm.dev/190)` in the string interpolation.

If I insert a deliberate cast for `$email = (string) $this->option('email');` instead, it fails on that line with `array<array-key, mixed> cannot be cast to string (see https://psalm.dev/190)`.

Any guidance? I've tried a few options and looked online but there was nothing obvious.

mechanarchy:
Hi asbiin, I tried converting the string-variable concatenation into an interpolation like you suggested, but Psalm tests fails due to `PossiblyInvalidCast` `non-empty-array<array-key, mixed> cannot be cast to string (see https://psalm.dev/190)`.

I looked at another command with a similar signature, and it is using string concatenation as well:
https://github.com/monicahq/monica/blob/5bf26262dc1d52596f5ceecec10429c6f8eb3ff7/app/Console/Commands/SetupProduction.php#L63-L69

I couldn't find a way around this after some searching and trial and error, so I think it may have to stay as concatenation unless you know a solution.

PR #3242, opened 2019-11-25
fix: update setup console command admin password

mechanarchy:
I've updated the console command output, as the admin account password is now `admin0`.


asbiin:
Nice catch @mechanarchy thanks !

PR #3243, opened 2019-11-25
fix: order contact tag list alphabetically

mechanarchy:
The list of applied contact tags on the Contacts list is currently unsorted:
image

This can make it quite difficult to locate and align related tags (eg `Work - Job1`, `Work - Job2`). This PR will sort the list of tags alphabetically:
image


app/Models/Contact/Tag.php
```diff
@@ -51,6 +51,7 @@ public static function contactsCount()
                     ->join('tags', 'tags.id', '=', 'contact_tag.tag_id')
                     ->where('tags.account_id', auth()->user()->account_id)
                     ->groupBy('tag_id')
+                    ->orderBy('name')
```
asbiin: Actually the `orderBy` sql instruction is really bad when it comes to non-ascii characters.
You'd better order the list after the get, with the `->sortByCollator` `Collection` macro.

mechanarchy: Thanks for the info! Resolved in https://github.com/monicahq/monica/blob/72ebeb5eb4dd1aea1a33f02b2b328263060f35fe/app/Models/Contact/Tag.php#L55

asbiin:
Thank you @mechanarchy

PR #3289, opened 2019-12-07
feat: enable api web navigation

asbiin:
This enables the api navigation into a browser, using actual credentials.
Authorization with token is still available.

This is really helpful, and is used on others apis like
* https://api.github.com/
* https://dev.azure.com/monicahq/monica/_apis

I also added an url in some resources, to be able to navigate, for instance from `/api/contacts` to `/api/contact/{id}` easily.

The view of `/api/`:
image


asbiin:
Problem: the 'web' auth does not check multi-factor authentication...

routes/api.php
```diff
@@ -2,129 +2,134 @@
...
+        Route::apiResource('contacts', 'ApiContactController')
+      ->names(['index' => 'contacts', 'show' => 'contact']);
+        Route::put('/contacts/{contact}/setMe', 'ApiContactController@setMe');
+
+        // Genders
+        Route::apiResource('genders', 'Account\\ApiGenderController');
+
+        // Relationships
+        Route::apiResource('relationships', 'ApiRelationshipController', ['except' => ['index']])
+      ->name('show', 'relationship');
```
djaiss: Indentation?

asbiin: This is due to StyleCI which has broken my indentation! I'll fix it

asbiin:
@djaiss 

> Just point to the URL in your browser and get the JSON back?

Yes that's it. Just log in the standard web interface, then open the `/api` url in your browser!

PR #3303, opened 2019-12-14
refactor: use request input

asbiin:
Refactor: use `$request->input` (Laravel) instead of `$request->get` (Symfony)

### DIFF
--- a/app/Http/Controllers/Api/ApiActivityController.php
+++ b/app/Http/Controllers/Api/ApiActivityController.php
@@ -85,7 +85,7 @@ class ApiActivityController extends ApiController
         JournalEntry::add($activity);
 
         // Now we associate the activity with each one of the attendees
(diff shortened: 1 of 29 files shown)


app/Http/Controllers/Api/ApiActivityController.php
```diff
@@ -140,7 +141,8 @@ public function update(Request $request, $activityId)
         JournalEntry::add($activity);
 
         // Get the attendees
-        $attendeesID = $request->get('contacts');
+        /** @var array */
```
djaiss: Why are you using this?

asbiin: Because it pleases psalm!

asbiin: (but I don't know why it does not detect it automatically)

PR #3310, opened 2019-12-17
test: use multiple small factories instead of a big one

djaiss:
This PR moves all the factories that were in the big `ModelFactory` file into smaller factories, grouped by types of objects – they actually follow the folder convention we have in place for the models now.

Having smaller files is easier to manage than a giant factory. I think it'll be easier to add new models that way.


djaiss:
@asbiin do you know why we are stuck here at the `monica ci` check that is not triggered? Can I merge without this?

djaiss:
Nevermind, it’s working now.

asbiin:
@djaiss did you try to turn it off and on again?

PR #3311, opened 2019-12-17
chore: add foreign keys on all tables

djaiss:
This finishes the job of #2280 and will close #2280 too.

This adds foreign keys to all tables in Monica.

I also have written two new services:
- ResetAccount
- DestroyAccount

which makes the Settings controller a little bit lighter and cleaner. We don't need the old way of resetting and deleting accounts with the foreign keys everywhere.


database/migrations/2019_12_17_024553_add_foreign_keys.php
```diff
@@ -0,0 +1,623 @@
...
+        $rows = DB::table('accounts')->select('id')->get();
+        $this->existingAccounts = [];
+        foreach ($rows as $row) {
+            $this->existingAccounts[$row->id] = 1;
+        }
+    }
+
+    private function contactExistOrFail(int $id)
+    {
+        if (isset($this->existingContacts[$id])) {
```
djaiss: I use `isset` for performance reasons. This is the fastest way to iterate through an array.

djaiss:
Ready for review @asbiin

database/migrations/2019_12_17_024553_add_foreign_keys.php
```diff
@@ -0,0 +1,623 @@
...
+     * users, as most data point to them somehow. For those 3 models, when we
+     * launch the script, we put all those ids in arrays. Then, for each
+     * foreign key, we check in the arrays if the key exists, instead of querying
+     * the database again and again. This is much faster.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::disableForeignKeyConstraints();
```
asbiin: great, but when are you calling `Schema::enableForeignKeyConstraints()`?

app/Http/Controllers/SettingsController.php
```diff
@@ -186,7 +173,7 @@ public function delete(Request $request)
             }
         }
 
-        DB::table('accounts')->where('id', $account->id)->delete();
+        $account->delete();
```
asbiin: Perfect!
This should be done on the `reset` function too?

database/migrations/2019_12_17_024553_add_foreign_keys.php
asbiin: 👍

asbiin:
I'm a little concerned about the performance here. The migration took ~6 sec on the [build pipeline](https://dev.azure.com/monicahq/monica/_build/results?buildId=2736&view=logs&j=dc476ecd-1df8-5762-5984-cfda40bf1aa7&t=c0083ff2-591c-5b64-8845-ca2b93e9b8a5&l=35), on an empty database.

djaiss:
> I'm a little concerned about the performance here. The migration took ~6 sec on the [build pipeline](https://dev.azure.com/monicahq/monica/_build/results?buildId=2736&view=logs&j=dc476ecd-1df8-5762-5984-cfda40bf1aa7&t=c0083ff2-591c-5b64-8845-ca2b93e9b8a5&l=35), on an empty database.

It took 4 minutes on my machine with a copy of the production database.

app/Http/Controllers/SettingsController.php
djaiss: It’s not possible, because in the case of a reset, we don't delete the account. So I can't evoke this and make it happen. I don't see how we can do it simply without iterating on every table like we do now.

database/migrations/2019_12_17_024553_add_foreign_keys.php
djaiss: Good point!

app/Http/Controllers/SettingsController.php
asbiin: We could delete all contacts for this account, and this should clean everything?

djaiss: You are right. I’m going to write a service for this.

djaiss:
Things that remain to be done
- [ ] Service to reset an account
- [ ] Add foreign keys to `contact` table

djaiss:
Ready for a new review @asbiin

asbiin:
On account reset, could we consider delete the default data populated with `$account->populateDefaultFields()`?

Also, I think you could add a foreign key on `avatar_photo_id`?

djaiss:
> On account reset, could we consider delete the default data populated with `$account->populateDefaultFields()`?

Why? We don't need to touch the default populated fields during the reset account process.

> Also, I think you could add a foreign key on `avatar_photo_id`?

Excellent point!

djaiss:
@asbiin another review ;-)

asbiin:
OK for me, go for it!

PR #3338, opened 2019-12-21
fix: update .platform.app.yaml - add app/temp to writable mounts

JGrubb:
Updates .platform.app.yaml - adding app/temp as writable directory under `mounts`.


asbiin:
Thank you @JGrubb

PR #3354, opened 2019-12-23
ADD yarn as devDependency

m0rtadelo:
`npm install` requires `yarn`:
imatge


asbiin:
Well, thank you @m0rtadelo but as we use yarn as our node dependency installer, we don't need to specify yarn as a dependency.
We don't use npm.

djaiss:
As @asbiin said, yarn is not a dependancy, it’s our package manager!

Issue #2196, opened 2018-12-24
Adding Households


tbirrell:
One of the things I've noticed is that handling events, addresses, notes, pets, and a few other things is very inelegant when it comes to households. 

For example, I have a friend who just got married. His wife still lived with her parents until their wedding. So upon marriage, I had to do several things. Update both of their addresses, add a life event for each. Remove a family pet from her profile and add his dog instead. Shortly thereafter they moved, which meant I had to go and update both addresses again, and add two more life events (one each), and leave an identical note on both accounts about the new place.

The only thing I did that was easy was changing their Love Relationship status from dating to married. Everything else had to done individually for each, which also means any change in the future that affects them as a couple, has to be synchronized to both. Or, for larger households, like with her old family pet, synchronized to a whole family (if I have multiple members in my DB) or group of roommates

So my suggestion is to implement an object called Households. It would be similar to tags, but if you added something to a person's account, you could click a checkbox that said "add for Household" or something, and it would be automatically duplicated for every other current member of the household. The reason it needs to be duplicated rather than assigned to the household and retrieved via a JOIN is for historical reasons. If someone leaves a household, I would still want to see all the life events and notes and other things that I had added.

This would be a large change, probably on par with Photos or Places. But I think it would be a good change, and definitely in line with the purpose of Monica. If this is something you are interested in pursuing, I have a few specific ideas that I can also share (though they may be more suited for a v2, after we get the initial idea up and running)

djaiss:
I agree. See #1810. Definitely interesting to add.

mccollam:
This would be very useful!  I think it's worth considering the idea that someone could potentially be a member of more than one household.  (An example would be a child with divorced parents.  The child spends 50% of their time at each parent's house, so they are a member of each parent's household.)

I also added a list of various "family" types to consider to #1810 which might have some overlap.

Issue #1810, opened 2018-09-10
Concept of family


djaiss:
I have two types of friends: 
- friends who are single
- friends who are in a couple

When I refer to a friend in a couple, most of the time I refer to the friend as a couple, not as a single entity.

Like: Let's go to a weekend with Rachel and Mike, even though I hang out with Mike.

- When two persons are in a relationship, we could suggest that those two are flagged "as a family".
- A family is a new type of entity which would contain all data about those two persons. If they have kids for instance (even if it's a recomposed family), they would all be listed there.
- All this would need a change of the UI. Viewing a single contact should not the same as viewing a family.
- The tricky part is to deal with the two people splitting. We would need to have a screen where we would ask the user to decide what to do with the data: where do the children go (in case of a recomposed family), what to do with the notes, etc...

djaiss:
@asbiin what do you think? It could be one of the next major new feature we implement. I think Monica should mimic real-life situations as much as possible.

mccollam:
If it's possible to generalize this beyond two people and still have it make sense, that would be great.  (I do have friends with families of >2 adults.)  A simple way to handle this might be to allow a "family" object to be a 1:many relationship, so any number of adults, children, pets, etc. can be part of the family.

This has the advantage of also allowing for things like spouse relationships where each member of the couple has a separate address.  (Which also happens, probably more frequently than triads or larger groups as families.  E.g. one person is in the military and deployed remotely while their spouse remains home.)

mccollam:
Just to expand on this some, here are some use cases (real ones in my life) that I would love to be able to capture as families:

* Two married people living together
* Two married people who live in separate locations
* Two married people living together with their children
* Two unmarried people living together
* Three or more adults living together, two of whom are married legally but are all three one family
* Two married people plus one of their parents living with them
* Two married people plus one of their parents plus their adult children (and their spouse(s) and their children)
* Two separate couples who each have part-time custody of the same child (i.e. the child is actually a member of two families that are otherwise unrelated)

So the notion of "family" is a bit complex :)

It might be that two people can be married but might not be a "family" for Monica (i.e. separate houses/pets), multiple adults could be a "family" (a triad or more, or a couple with adult parents or children), or one person could be a member of multiple "families" (a child with divorced parents).

Sorry if this is too much detail -- I just want to be sure that if a concept of family is introduced that it is as inclusive and useful as possible!

virtadpt:
[...]

PR #3377, opened 2019-12-27
feat: add concept of family

djaiss:
This PR adds the concept of family, as described in #2196 and #1810  

A family is a group of contacts. In this first iteration:
* you can create a family by giving a name
* you can associate contacts with it. Note that when you do associate a contact, this contact's relationships are not attached automatically.
* you can remove contacts from it.
* APIs for these features have been written.

## Checklist

- [ ] Front-end tests have been written with Cypress.

#### Backend/models changes
- [ ] The API has been updated.
- [ ] API's documentation has been added by submitting a pull request in the [marketing website repository](https://github.com/monicahq/marketing_site/pulls).
- [ ] Tests have been added for the new code.
- [ ] If you change a model, make sure the FakeContentTableSeeder is updated. We need seeders to develop locally and generate fake data.

- [ ] Make sure exporting account data as SQL is still working.
- [ ] Make sure your changes do not break importing data with `vCard` and `.csv` files.
- [ ] Make sure account reset and deletion still work.

#### Other tasks
- [ ] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG.md) entry added, if necessary, under `UNRELEASED`.
- [ ] If it's relevant and worth mentioning, create a changelog entry for this change. The changelog entry will appear inside the UI for all users to see. To know if your change is worth the creation of a changelog entry, [read the documentation](https://github.com/monicahq/monica/blob/master/docs/administrators/tips.md#when-is-it-relevant-to-create-a-changelog-entry).


asbiin:
That's very interesting! I will review this later, but I just wanted to put here an idea: why limit this to "families" if we could extend this to any kind of "groups"?
* you can create a group by giving a name
* you can associate contacts with a group.
* you can remove contacts from it.
A family is a group with some special features, so let's create the notion of "group" first.

djaiss:
> A family is a group with some special features, so let's create the notion of "group" first.

You are actually right. A family is a subset of a group, but having a notion of group would be more generic. I'll update the code to use group instead.

Don't review the PR now, a lot needs to be done before it’s ready for review.

djaiss:
@asbiin I’m very confused by this.

A group can be so much more than just a family. 
A group can 
* contain any number of contacts
* should **not** appear in the list of contacts. 
* Also, people in a group are not necessarily related to each other (with our concept of relationships).

However a family 
* has people that are all related to each other
* should appear in the contact list.

To me, family and groups are not the same thing at all. A group is a group of individuals. A family are people related by blood. A group can contain contacts AND families. But a family can't contain a group, only contacts.

I imagine groups being used in a business context, or a religious one, or in a community.

That means we would have this "hierarchy" of concepts:
contact <  family <  group.

I think we still need the Family object in Monica, as well as the Group object. I also recognize that these two are similar.

djaiss:
Also, I've started to completely redesign and simplify the contact list, but it’s too big for this PR, so I'll do it elsewhere. Here is the gist used for this: https://gist.github.com/djaiss/256af24e4ca26969ab6f6861a7e1263a and https://gist.github.com/djaiss/cef7e730182a8b787c20a4a1c3f56f14

djaiss:
@asbiin it’s ready for review.

There is no front-end for this yet. I’ll do the front-end in the new version of Monica.

The API doc is here: https://www.monicahq.com/api/groups and live.

asbiin:
This PR has been on hold for a long time.
I'll close it for now, don't hesitate to reopen if needed.

PR #3422, opened 2020-01-01
feat: enable activities edit

asbiin:
Close #3415

PR #3431, opened 2020-01-04
fix: api routes

paddockFerrari:
It looks like something has been deleted but the routes for it were still there. I experienced it when I wanted to call `php artisan route:list`.


asbiin:
They will be used is #3422 again!

paddockFerrari:
Ok, didn't take a closer look.

PR #3446, opened 2020-01-07
feat: add new name order pattern nickname (first name last name) 

### LINKED ISSUE #3139: New name pattern
I use the 'nick name' field to enter people's 'calling name' (roepnaam, in Dutch) - how we call ppl in daily life. That's also the name by which I'd like to display & sort contacts.

Therefore, I'd like to suggest a new name pattern: `<Nickname> (<First name> <Last name>)`, which would resort to `<First name> <Last name>` if there is no nickname.

maantje:
As requested in #3139, this PR introduces a new name pattern: \<Nickname> (\<First name> \<Last name>), which would resort to \<First name> \<Last name> if there is no nickname. 

Close #3139


maantje:
Yeah, that would make it more complete. 

I'll try to update the merge request this week.

PR #3460, opened 2020-01-11
chore: use class names in routes

asbiin:


asbiin:
Apparently it's not  ok :D

The gain is to have a direct code reference between the class and the route file, adding the ability to clik'n go to the class in your IDE like VSCode.

PR #3487, opened 2020-01-20
chore(docker): fix docker exclude list

asbiin:


nwalke:
This is causing #3743

PR #3503, opened 2020-01-24
fix: activity seeder to use new schema

mechanarchy:
The activity seeder got broken in commit 3e007af34e7b08772f424080384f8210591a0287 recently:
https://github.com/monicahq/monica/blob/3e007af34e7b08772f424080384f8210591a0287/app/Services/Account/Activity/Activity/CreateActivity.php#L26-L28
vs
https://github.com/monicahq/monica/blob/3e007af34e7b08772f424080384f8210591a0287/database/seeds/FakeContentTableSeeder.php#L283-L289

This PR fixes it.


asbiin:
thank you @mechanarchy

Issue #3390, opened 2019-12-29
List of Contacts shows contacts multiple times when ordering by activity date


xorgnz:
**Describe the bug**

Repro:

- Go to the list of contacts.
- Choose one of the two "Sort By" options from the drop down.
- Observe that contacts with multiple activities appear multiple times.

Only occurs if you have multiple activities logged for at least some contacts of the contacts listed. Occurs even if you have a search term that limits it down to a single contact.

Looking at code, it appears to be missing groupBy clauses in app/Models/Contact.php -- scopeSortedBy().

**Screenshots**
Running on MonicaHQ. Here's an example with a single contact 
![image](https://user-images.githubusercontent.com/931641/71551347-6a71a280-2a4a-11ea-938c-2d2fd69690b9.png)

djaiss:
This has been fixed in #3390 and will be live soon.

PR #3512, opened 2020-01-25
fix: contact sorting by last activity date (#3390)

anegoda1995:
I fixed it.
Before current fix:
image
image

With current fix:
image
image


djaiss:
Thanks so much @anegoda1995

PR #3536, opened 2020-02-01
feat: add a link to archive all contacts during downgrade process

djaiss:
This PR adds a link to archive all the contacts in your account during the downgrade process.

When we downgrade, we must have less than the authorized limit (as set in the .env file). In our instance, this limit is 10 contacts.

But for people who have uploaded a CSV, with say, 300 contacts, they can't easily archive contacts to facilitate the downgrade process.

This PR adds a way to achieve this.


djaiss:
@asbiin do you think we can remove the SonarCloud code analysis for the controllers? We don't really test controllers in our code anyway.

asbiin:
> @asbiin do you think we can remove the SonarCloud code analysis for the controllers? We don't really test controllers in our code anyway.

Well, I don't fully agree with that. We should test controllers, and we are doing it at 64.7% right now.
Solution is: you can exclude a function from code coverage by adding `@codeCoverageIgnore` to the function's phpdoc.

resources/views/settings/subscriptions/archive.blade.php
```diff
@@ -0,0 +1,51 @@
...
+          @include('partials.errors')
+
+          <h2>{{ trans('settings.archive_title') }}</h2>
+
+          <p>{{ trans('settings.archive_desc') }}</p>
+
+          <form method="POST" action="{{ route('settings.subscriptions.archive') }}">
+            @csrf
+
+            <p class="mb4"><button type="submit" href="" class="btn btn-primary">{{ trans('settings.archive_cta') }}</button></p>
```
asbiin: `href=""` is useless here

resources/lang/en/settings.php
```diff
@@ -44,7 +44,8 @@
     'layout_big' => 'Full width of the browser',
     'save' => 'Update preferences',
     'delete_title' => 'Delete your account',
-    'delete_desc' => 'Do you wish to delete your account? Warning: deletion is permanent and all your data will be erased permanently.',
+    'delete_desc' => 'Do you wish to delete your account? Warning: deletion is permanent and all your data will be erased permanently. Your subscription (if you have any) will also be immediately cancelled.',
+    'delete_other_desc' => 'Just to be clear: your data in the main database will be deleted immediately. However, as described in our privacy policy, we do daily backups of the database in case of failure and this backup is kept for 30 days – then it’s completely deleted. It’s unrealistic to imagine that we can go in all the backups to delete your specific data. By the way, this data is encrypted on very secure Amazon servers and no one has the encryption key except us. Therefore, your data will completely disappear in 30 days from all the backups, and no one will know this data ever existed in the first place.',
```
asbiin: Wow, that's clear!

app/Http/Controllers/Settings/SubscriptionsController.php
```diff
@@ -122,13 +131,41 @@ public function downgradeSuccess(Request $request)
...
     }
 
+    /**
+     * Display the archive all your contacts page.
+     *
+     * @return View|Factory|RedirectResponse
+     */
+    public function archive()
+    {
+        $contacts = auth()->user()->account->contacts();
```
asbiin: This is not used and should be removed

resources/views/settings/subscriptions/archive.blade.php
```diff
@@ -0,0 +1,51 @@
...
+
+          </form>
+
```
asbiin: It would have a better see adding 
```
          <div class="br3 ba b--gray-monica bg-white mb4">
            <div class="pa3 bb b--gray-monica">
```

and if the text and button are text-align center:
```
              <p class="tc">{{ trans('settings.archive_desc') }}</p>
```
```
                <p class="tc mb4">
                  <button type="submit" class="btn btn-primary">{{ trans('settings.archive_cta') }}</button>
                </p>
```

resources/views/settings/subscriptions/archive.blade.php
djaiss: Indeed, my bad.

resources/lang/en/settings.php
djaiss: Yes I want it to be extremely open and transparent.

resources/views/settings/subscriptions/archive.blade.php
djaiss: Yeah I agree 😀 The design is really crap but as it’s a feature that would be used very few times... it’s not that big of a deal. i’ll make the changes though.

djaiss:
@asbiin comments applied, feel free to recheck.

djaiss:
Oops I forgot the code coverage thing.

That being said, how do we test controllers, really?

asbiin:
> Oops I forgot the code coverage thing.
> 
> That being said, how do we test controllers, really?

All these tests are in the tests/Feature directory (a lot of them made by yourself I think ;) )

djaiss:
@asbiin yeah but the problem is that when you add a new method in a controller that existed before and was never tested against code coverage rate, you end up having to test the ENTIRE controller and all its methods (and not just the one you added in the first place) 😅

asbiin:
> you end up having to test the ENTIRE controller and all its methods

I don't see why?

PR #3537, opened 2020-02-01
fix: fix markdown not being applied for a phone call description

### LINKED ISSUE #3388: The calling texts doesn't show <ul> lists
**Describe the bug**

When you fill a note in a contact with a list like this:
`- Blabla`
`- Blabla`

The dashes are converted to a html list `<ul>`:
- Blabla
- Blabla

But the issue is in the calling texts, although I put a list, it isn't converted to a html list `<ul>`.

 - A server you maintain yourself with version 2.15.2

djaiss:
Close #3388

Result:

image


djaiss:
> Wow, I didn't know this could be so easy to fix!

Yeah I added a long time ago a CSS class that deals with every markdown related.

PR #3541, opened 2020-02-01
fix: fix display of the toggle buttons in the Settings page

### LINKED ISSUE #3496: Toggles don't change visually
When clicking these toggles, a message in the right bottom appears

image

image

But toggle state doesn't change visually. If I refresh the page they adopt their correct value

djaiss:
Fix #3496 

Toggles buttons were broken. They are not anymore.

They didn't change their state upon selection.


djaiss:
@asbiin again, failure on controllers coverage. We should disable this verification on controllers.

PR #3564, opened 2020-02-07
chore: refactor user model

djaiss:
The User model was huge, and contained a lot of logic that shouldn't belong to the model itself.

I've removed almost all of the methods that didn't belong there and created new services or new helpers for them.

The main goals are:
- models should have the least amount of logic possible. They should mainly be used to have setters/getters. If there is a logic, it should be placed in a service instead.
- a service should be used when the object is changed (create/edit/delete)
- models should be as small as possible.
- helpers should be used whenever there is a logic that we should know about but that doesn't change the object itself.


app/Services/User/UpdateViewPreference.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public function execute(array $data): User
+    {
+        $this->validate($data);
+
+        try {
+            $user = User::where('account_id', $data['account_id'])
+                ->findOrFail($data['user_id']);
+        } catch (ModelNotFoundException $e) {
+            throw new ModelNotFoundException(trans('app.error_user_account'));
```
asbiin: why rethrow a new ModelNotFoundException here ? removing the try/catch should be safe

app/Http/Controllers/Api/Account/ApiUserController.php
```diff
@@ -30,18 +36,35 @@ public function show(Request $request): UserResource
...
 
-        if (! $userCompliance) {
+        if ($termUser) {
+            $data = [
+                'signed' => true,
+                'signed_date' => DateHelper::getTimestamp($termUser->created_at),
+                'ip_address' => $termUser->ip_address,
+                'user' => new UserResource(auth()->user()),
+                'term' => new ComplianceResource($term),
+            ];
```
asbiin: Maybe this data could be transformed in a Resource? It's used 3 times in this file

app/Services/User/UpdateViewPreference.php
djaiss: So I can override the exception by adding a custom message to it, to give more context to the classes who use this service.

app/Http/Controllers/Api/Account/ApiUserController.php
djaiss: Interesting, thanks!

PR #3570, opened 2020-02-08
chore: refactor account model

djaiss:
This PR refactors the Account model.

The main goals are:
- models should have the least amount of logic possible. They should mainly be used to have setters/getters. If there is a logic, it should be placed in a service instead.
- a service should be used when the object is changed (create/edit/delete)
- models should be as small as possible.
- helpers should be used whenever there is a logic that we should know about but that doesn't change the object itself.

Here are all the changes:

* move the canDowngrade method to a new helper (Account Helper)
* remove the timezone() method that was not used anywhere
* move the getStorageSize() method to a new StorageHelper helper
* move hasReachedAccountStorageLimit() method to the StorageHelper helper
* move yearly statistics methods to AccountHelper helper
* move the hasLimitations() method to AccountHelper
* move accountHasLimitations method to AccountHelper
* move getRemindersForMonth() to AccountHelper
* move replaceGender to GenderHelper
* move defaultGender() method to AccountHelper
* move hasAny() method to InstanceHelper helper


djaiss:
@asbiin it was a really bad idea from me to submit a big PR like that. There are too many changes. It’s the last time.

resources/views/settings/subscriptions/downgrade-checklist.blade.php
```diff
@@ -49,11 +49,11 @@
               <span class="rule-to-succeed">{!! trans_choice('settings.subscriptions_downgrade_rule_invitations_constraint', $numberOfPendingInvitations, ['url' => route('settings.users.index'), 'count' => $numberOfPendingInvitations]) !!}</span>
             </li>
 
-            <li class="{{ (auth()->user()->account->hasReachedContactLimit() == true)?'fail':'success' }}">
+            <li class="{{ ($hasReachedContactLimit == true)?'fail':'success' }}">
```
asbiin:
```suggestion
            <li class="{{ $hasReachedContactLimit ? 'fail' : 'success' }}">
```

resources/views/settings/subscriptions/downgrade-checklist.blade.php
```diff
@@ -49,11 +49,11 @@
...
               <span class="icon"></span>
               <span class="rule-title">{{ trans('settings.subscriptions_downgrade_rule_contacts', ['number' => config('monica.number_of_allowed_contacts_free_account')]) }}</span>
               <span class="rule-to-succeed">{!! trans_choice('settings.subscriptions_downgrade_rule_contacts_constraint', $numberOfActiveContacts, ['url' => '/people', 'count' => $numberOfActiveContacts]) !!}</span>
-              @if ((auth()->user()->account->hasReachedContactLimit() == true))
+              @if (($hasReachedContactLimit == true))
```
asbiin:
```suggestion
              @if ($hasReachedContactLimit)
```

app/Helpers/InstanceHelper.php
```diff
@@ -60,4 +62,14 @@ public static function getChangelogEntries($limit = null)
...
+
+    /**
+     * Check if the instance has at least one account.
+     *
+     * @return bool
+     */
+    public static function hasAtLeastOneAccount(): bool
+    {
+        return DB::table('accounts')->count() > 0;
+    }
```
asbiin: We could create this function as `hasOnlyOneAccount`, as almost all call to this function are like `! hasAtLeastOneAccount()`

djaiss: I completely agree BUT it’s one more change in an already complex PR and I would prefer make this change in a separate PR.
I still need to fine tune the Account model to remove everything about the population of data anyway.

djaiss:
Changes made. Final check!

PR #3573, opened 2020-02-12
feat: services to set and clear personal description

djaiss:
This PR adds two new services:
* one to set the contact's personal description
* the other one to clear the personal description.

Both services will be needed for the new UI.


djaiss:
@asbiin I need this PR to be implemented tonight so I can move forward with what I want to do for the new UI. These are two really simple services so I think it’s safe that I merge them right away without code review.

PR #3585, opened 2020-02-14
feat: add ability to indicate the number of contacts in seeders

djaiss:
This PR adds the ability to specify the number of contacts that are generated using the seeder for tests account. Before, we only had between 60 and 100 contacts in a test account. Now, with this change, you can have any number of contacts.

Because we need to ask for the number of contacts, we should run the FakeContentTableSeeder in a command.

This is why all the code from FakeContentTableSeeder has been moved to the SetupTest command.

@asbiin this is an important change, but not a complex one.


asbiin:
Oh great, it's been a long time I wanted to do this 😊

PR #3586, opened 2020-02-15
chore: last consulted at field update is now in a job

djaiss:
The `contact->updateConsulted()` method is now done in a job, instead of being a method called on the Contact object.

First of all, it’s better to put this kind of operations in a job (instead of blocking the main thread for this), and also, I’ll need this in a job for the new UI 😀

cc @asbiin


asbiin:
OK

PR #3588, opened 2020-02-16
chore: add audit log for updating work information

djaiss:
This PR

- adds audit log when updating work information,
- rename this from `UpdateContactWork` to `UpdateWorkInformation`.

It also adds a unit test for this service which didn't existed.


djaiss:
@asbiin FYI 😀

PR #3592, opened 2020-02-17
chore: add audit logs for company creation

djaiss:
This PR adds audit log when a company is created.

Also, it moves all the strings for audit logs from the Account file to a new i18n file called `logs`.


djaiss:
@asbiin FYI.

asbiin:
> @asbiin FYI.

Noticed.
(Well I don't think it was just a _chore_)

PR #3597, opened 2020-02-18
fix: fix resend email verification feature not sending email

djaiss:
This PR fixes the Resend verification email not working.

The route needs a POST and not a GET, therefore going to error.

We probably should modify the test as well.


resources/views/auth/login.blade.php
```diff
@@ -57,15 +57,12 @@
...
-              @if (isset($errors))
-                @if (count($errors) > 0)
-                  <div class="form-group links">
-                    <ul>
-                      <li>{{ trans('auth.password_forget') }}&nbsp;<a href="{{ route('password.request') }}">{{ trans('auth.password_reset') }}</a></li>
-                    </ul>
-                  </div>
-                @endif
-              @endif
+              <div class="form-group links">
```
djaiss: This is just a nice to have. I don't see why we should hide this link unless we make an error.

resources/views/auth/verify.blade.php
```diff
@@ -22,7 +22,11 @@
               @endif
 
               {{ trans('auth.confirmation_check') }}<br/>
-              {!! trans('auth.confirmation_request_another', ['url' => route('verification.resend')]) !!}<br/>
+              {!! trans('auth.confirmation_request_another') !!}<br/>
+              <form action="{{ route('verification.resend') }}" method="post" class="di">
```
djaiss: This is the fix here. It needs to call a form that triggers a POST request, therefore the use of the <form> as well as the submit button.

resources/lang/en/auth.php
```diff
@@ -73,7 +73,8 @@
     'confirmation_title' => 'Verify Your Email Address',
     'confirmation_fresh' => 'A fresh verification link has been sent to your email address.',
     'confirmation_check' => 'Before proceeding, please check your email for a verification link.',
-    'confirmation_request_another' => 'If you did not receive the email <a href=":url">click here to request another</a>.',
+    'confirmation_request_another' => 'If you did not receive the email',
```
asbiin: I would rather write
```suggestion
    'confirmation_request_another' => 'If you did not receive the email <button type="submit" class=":class">click here to request another</button>.',
```

and use this single text in the form

resources/views/auth/login.blade.php
asbiin: OK why not

asbiin:
@djaiss I've submitted an other solution.
The big green button on the middle of the page was really not nice.
This is the UI with this solution:
image

(yes there is no difference with the actual one)

djaiss:
@asbiin awesome solution. let's merge this.

PR #3606, opened 2020-02-19
feat: add contact field labels

asbiin:
First step to add contact field labels.

A label is here to know the scope of a contact field (email, phone, etc) or address: work, home, main, personnal, etc.
It's already handled on many contact client (phone contact app, outlook, thunderbird, ...), so we just need to have it and provide it with vcards export.

- [x] add tables and models
- [x] improve vcard export
- [x] Api
- [x] Tests
- [X] UI will not be done in this PR

ContactFieldLabel type has label_i18n and label attribute:
- `label_i18n` is a standard label. The internal name (english) will be used on vcard export. On UI the label displayed will be get from `trans` method
- `label` is a personnal label. This label will be used on both vcard and UI because the user will write it theyself.

The behiavor is:
- Let the user choose for some standard label: home, work, cell, fax, pager, main, other ; using translations (the list will depends on the type of contact field currently editing, as every label are not standard on every types - this will be done in another PR)
- The label 'personal' is an extra label, is the user choose it, it let the user write a label itself. So 'personal' label is not intended to be used as a standard label.
- use the standard label, or the personal one as labels on the contact field: standard labels have a translation, and personal one will be displayed as is.


app/Interfaces/LabelProvider.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace App\Interfaces;
+
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+
+interface LabelProvider
```
djaiss: "Provider" for an interface? This is confusing. Why not "LabelInterface"?

asbiin: Good point

asbiin:
@djaiss ok I think it' ready for a review

PR #3638, opened 2020-02-29
feat: add a weekly job to update gravatars

### LINKED ISSUE #3206: Avatars: query gravatar for all emails
Avatar update only display the first email, which does not have necessary an associated gravatar. We could iterate over each email to check if there is a gravatar with it.

### LINKED ISSUE #3207: Schedule a gravatar update regularly
Gravatar linked to an email contact could change, but today we only change it on email add or edit.
We could query gravatar for an update of every email regularly.
For example it could be a good start to query an email every 3 months.

After that we have to notify the user about the change.

asbiin:
Add a weekly job that update the gravatars for all contact which has a gravatar url.
This close #3207

Also update gravatar based on all emails. The first email with a gravatar will be taken as gravatar url.
This close #3206


asbiin:
@djaiss tests have been done ;)

PR #3639, opened 2020-03-01
chore: remove access of eager load object if not needed

asbiin:
because we can


app/Console/Commands/SetupTest.php
```diff
@@ -269,7 +269,7 @@ public function populateBirthday()
             $birthdate = $this->faker->dateTimeThisCentury();
 
             app(UpdateBirthdayInformation::class)->execute([
-                'account_id' => $this->contact->account_id,
+                'account_id' => $this->contact->account->id,
```
djaiss: Nope!

```suggestion
                'account_id' => $this->contact->account_id,
```

app/Console/Commands/SetupTest.php
```diff
@@ -607,7 +607,7 @@ public function populateCalls()
     {
         if (rand(1, 3) == 1) {
             $this->contact->calls()->create([
-                'account_id' => $this->contact->account_id,
+                'account_id' => $this->contact->account->id,
```
djaiss: All this file is wrong :-D

djaiss:
Just changes in one file and it's good to go!

app/Console/Commands/SetupTest.php
asbiin: That's a bad revert indeed

PR #3661, opened 2020-03-05
fix: Sort relationship by relationship types

### LINKED ISSUE #2395: Sort Relationships
I believe the profile sidebar would look cleaner if the related contacts were better sorted by relation type. For instance, instead of showing a "grand child" tag next to every grand child, there should be a Heading "Grand Child" with each of them listed below.

Quick Mock-up:
monica-family_relationships

ashawe:
FIX #2395 
Sort the  array by relationshipType
if relationshipType occours for first time, save it and display it
else continue displaying other details

when relationshipType changes, it won't be equal to curr_relationship and hence will display it

**UI CHANGES:** replace the silver badge with <h3> @djaiss 

NOTE: test are not written in cypress yet

### DIFF
--- a/resources/views/people/relationship/_relationship.blade.php
+++ b/resources/views/people/relationship/_relationship.blade.php
@@ -1,7 +1,10 @@
 @php
-    function my_cmpr($a, $b)
+    if(!function_exists('my_cmpr'))
     {
+      function my_cmpr($a, $b)
+      {
         return strcmp($a->relationshipType->getLocalizedName(null, false, $relationship->ofContact->gender ? $relationship->ofContact->gender->type : null), $b->relationshipType->getLocalizedName(null, false, $relationship->ofContact->gender ? $relationship->ofContact->gender->type : null));
+      }
     }
     usort($relationships, "my_cmpr");
     $curr_relationship = NULL;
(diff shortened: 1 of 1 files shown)


resources/views/people/relationship/_relationship.blade.php
```diff
@@ -1,9 +1,25 @@
+@php
```
asbiin: I really don't like using `@php` in Laravel.
In *every* case, it could be done in the Controller instead, and that's the place it has to be moved.

ashawe: can you point me in the right direction and tell me in which file should I place the sort function and where does the ```$relationships``` array comes from, So I can sort them there itself and do the displaying here accordingly

asbiin: The relationships array comes from `resources/views/people/relationship/index.blade.php` file, and the `ContactsController@show` method. Each `@include` use another list of relationships (which is not really optimized code btw...)

You can call the `getLocalizedName` and store it in the relationship object using a `map` method, and reorder the collection with `sortByCollator` macro.

asbiin:
closing this after long non-activity time

PR #3675, opened 2020-03-09
docs: Update heroku.md to reflect switch to JawsDB

buenalabs:
Edit text of heroku.md to reflect switch from ClearDB to JawsDB as Monica's default MySQL storage provider.


asbiin:
Many thanks @buenalabs

PR #3681, opened 2020-03-11
chore: add webpack-shell-plugin

asbiin:


djaiss:
What is this for?

asbiin:
It allows you to run `php artisan lang:generate` every time you build the assets.
I'm not sure yet we should have it...

PR #3699, opened 2020-03-17
docs: Upload new screenshot to reflect switch to JawsDB

buenalabs:
Upload a new screenshot that will replace the second screenshot in heroku.md. The old screenshot is outdated, still showing IgniteDB (instead of JawsDB) as the default database storage provider

**Old screenshot**
outdated screenshot

**New screenshot**
updated screenshot


asbiin:
Thanks @buenalabs !

PR #3718, opened 2020-03-22
docs: Update heroku docs to use updated screenshot

buenalabs:
Update heroku.md to use an updated second screenshot to reflect switch to JawsDB (from ClearDB) as Monica's default database provider in Heroku.


asbiin:
Great ! thank you @buenalabs

Issue #2738, opened 2019-05-30
CardDAV Client Support


Wdavery:
This was mentioned near the top of #42
> Don't forget that some already have carddav servers setup. So having Monica updating those remote servers would be a good feature. That way we'd have a single source of truth for contact information.

Forgive me if this is part of what is currently being implemented, but from what I've read in the multiple open issues about CardDAV, it seems to only be implement a CardDAV server, not implementing a way to interface with existing CardDAV Servers.

I don't want to change where I'm hosting my contacts, so it'd be great if Monica could interface with my current setup and update contacts accordingly.

asbiin:
For now, this has not been prioritized.

Monica is aim to be a prinicipal contact (carddav) server, and not the other way.
But I understand the need. 

I'm not aware of any contact management system with carddav server, that implements a carddav client too... Maybe you have an example? But Monica could be the first?

Wdavery:
I totally understand the priority of server first, as the aim of the project is to be a complete management system itself.

I’m not sure I’ve seen a server/client example myself, Monica may be the first. 
I think that it exemplifies the open source idea, giving the power to the users, giving choice in how to use the software. 

I currently use Nextcloud as my CardDav server and my entire family uses it as well. However not everyone is interested in using Monica and I wouldn’t want to give up the sharing and grouping functionalities we have now by splitting my user base across two servers. 

Monica has come quite far in the time I’ve been watching it and I’m excited to continue seeing it grow!

lucker999:
From a user point of view, I would rather have a Nextcloud-ish server with a growing integrated apps base then use standalone Monica and dozen others services which need additional integration. For instance, I love the idea of personal CRM and have been searching for it quite a while. I think Monica is the surest way such personal / family CRM should be. But the core of a CRM are the contacts. And if no integration with existing, wide-spread and growing contact store/sync/manage solution provided, the user base will be limited to a very narrow lot of dedicated followers who tolerate inconveniences (or can create a solution by themselves). 
Since the project provide no way to sync contacts with existing solutions (android/iOS phones, PIM software such Thunderbird, Outlook, etc., cloud contact management like Gmail, Outlook365, etc) the usage case quite limited, tedious and cumbersome. But I would be a first to install, praise and donate to a day Monica provide sync solution for the contacts.

dosch:
Couldn't agree more. 
I find myself checking back on the progress of this kind of integration for well over a year now. Waiting for it to be complete so I can finally move over....

PR #3749, opened 2020-03-26
feat: add notion of addressbooks

asbiin:
This adds the notion of Address Book
It's the first step for  #2738

This let an account have multiple address book.
The default one is the current solution (with only 1 address book).
The notion of address books is already integrated on carddav.

This PR only address the ability to *have* multiple address books.
Other PR will gave the opportunity to
- add other distant address books, and add subscription to them
- connect to carddav clients and update them regularly
- display address books content on dashboard

All API functions only address the default address book.

So basically it does not change a lot of things immediatly.


database/migrations/2020_03_25_055551_add_address_book.php
```diff
@@ -0,0 +1,40 @@
...
+            $table->foreign('user_id')->references('id')->on('users')->onDelete('cascade');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
```
djaiss: We don't need down() method at all. We never rollback. I don't include them in my code ever.

djaiss:
I don't understand this feature. Can you explain more? Also, I have the feeling it makes the code much more complex, for instance when downgrading, now, as a developer, we will need to think about checking contacts in the address book. Why? How can we make it very simple for developers to not think about that?

database/migrations/2020_03_25_055551_add_address_book.php
asbiin: Well, I do.

asbiin:
We could add parameters used on the scope function to the `Account::contact` one, so it will be easy for developers, because `account->contact()` will be the default address book, and `account->contact($id, $abid)` will return a specific contact list.
The only limit is to get the ALL contacts of an account. Maybe if addressbookId is `-1`, it return every contacts.

tomperrine:
> I don't understand this feature. Can you explain more? Also, I have the feeling it makes the code much more complex, for instance when downgrading, now, as a developer, we will need to think about checking contacts in the address book. Why? How can we make it very simple for developers to not think about that?

For example, I have multiple address books - one from O365 at work, a personal gmail account, a consulting email account - I'd like to have all these in Monica, yet be able to manipulate them independently when needed.

app/Models/Contact/Contact.php
```diff
@@ -162,6 +163,26 @@ public function account()
...
+    }
+
+    /**
+     * Get the list of contacts from the same address book as this contact.
+     *
+     * @return HasMany|null
+     */
+    public function addressBookContacts()
+    {
+        return $this->account ? $this->account->addressBookContacts($this->addressBook ? $this->addressBook->name : null) : null;
```
djaiss, line 183: Way too complex in my opinion.
You are trying to be too smart 😀 Yes the code is more pure that way, but it comes with a price of having to decrypt the line to understand what it does.

Could you make it multi line so it’s easy to read in one glance?

app/Models/Contact/Contact.php
```diff
@@ -162,6 +163,26 @@ public function account()
...
+    {
+        return $this->belongsTo(AddressBook::class);
+    }
+
+    /**
+     * Get the list of contacts from the same address book as this contact.
+     *
+     * @return HasMany|null
+     */
+    public function addressBookContacts()
```
djaiss, line 181: I don't think this is a good name.

To reduce complexity we should have simple names. I think, in this case, what are you trying to achieve? To get the contacts from the same address book.
So it could be `similarContactsInAddressBook()` or `sameAddressBookContacts()`.

app/Models/Contact/Contact.php
```diff
@@ -546,6 +567,28 @@ public function scopeNotActive($query)
         return $query->where('is_active', 0);
     }
 
+    /**
+     * Scope a query to only include contacts from designated address book.
```
djaiss, line 571:
```suggestion
     * Scope a query to only include contacts from given address book.
```

app/Http/Controllers/DashboardController.php
```diff
@@ -28,19 +28,17 @@ public function index()
             )->with('debts.contact')
             ->first();
 
-        $numberOfContacts = $account->contacts()
-                                    ->real()
-                                    ->active()
-                                    ->count();
+        $numberOfContacts = $account->addressBookContacts()
```
djaiss, line 31: Here I don't like it 😀
Why do we have to think about querying this method instead of `contacts`?

I think address books should be transparent unless we really want to manage it in the code/controller because we need to manage different address books.

In this case, the right call should be `account->contacts()` and that's it. We should manage the notion of address books in the `contacts()` method, and that way everything would be transparent.

Or perhaps it should be something like 

```
$account->contacts()
     ->real()
     ->active()
     ->count();
```

and we could have another scope to specify the address book when we need to scope this to a specific address book?

database/migrations/2020_03_25_201407_add_contact_vcard_data.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('contacts', function (Blueprint $table) {
+            $table->string('vcard', 65536)->after('gravatar_url')->nullable();
```
djaiss, line 17: What does this do? It’s not clear.

app/Services/VCard/ImportVCard.php
```diff
@@ -414,7 +439,12 @@ private function existingContactWithEmail(VCard $entry): ?Contact
                 'contact_field_type_id' => $this->getContactFieldTypeId(ContactFieldType::EMAIL),
             ])->whereIn('data', iterator_to_array($entry->EMAIL))->first();
 
-            if ($contactField) {
+            if ($contactField &&
```
djaiss, line 442: We should add a comment above this `if` to give the reader an overview of what this complex `if` does.

app/Http/Controllers/DashboardController.php
asbiin: That was my first try.. and it does not work like this. contacts() return a HasMany, and I can add a filter in this function. I don't know where to insert the scope/Query without explicitly call it. If you have a better idea ok, but I have zero.

database/migrations/2020_03_25_201407_add_contact_vcard_data.php
asbiin: The best explanation is here:
https://sabre.io/dav/building-a-carddav-client/#retain-full-vcards%21

asbiin: The idea is to record the generated vcard.
It will only be used for external address books. Internal default address book will not save the vcard.

app/Http/Controllers/DashboardController.php
asbiin: There is not "less complex code" where we code magically guess the address book currently seeing.

I guess I could just revert the changes and go back to something like
```
$account->contacts()
     ->addressbook($addressBookName)
     ->real()
     ->active()
     ->count();
```
If $addressBookName is null it will scope to default address book (null id), if not it will scope to the address book.

asbiin:
Thanks for the review.
I don't think it adds so complex code .
However, I understand your point, and I tried to simplify it. I guess it's not working for you ...

PR #3755, opened 2020-03-27
docs: Replace screenshots in heroku.md with optimized ones

buenalabs:
- Upload Web-optimized screenshots that are 2/3 the size of the old ones and delete the later.
- Update heroku.md to reference newer files.


asbiin:
I don't see the difference between the images. The newer are even bigger than the old ones.

PR #3757, opened 2020-03-28
chore: change styleci conf

asbiin:
Use default laravel config for styleci


djaiss:
Wait wait. I like imports that are ordered by length and not by name! Why change that?

asbiin:
That is the default in Laravel now

PR #3779, opened 2020-03-31
fix: edit multiple notes at the same time

### LINKED ISSUE #1302: Editing multiple notes at the same time only edits one note
Actual behaviour:

When editing multiple notes at the same time (klicking "edit" on 
multiple notes), when saving one note all notes are set back to non-edit 
mode, and all changes in other notes but the one just saved are discarded.

Expected behaviour:

When editing multiple notes klicking on save only the saved note should 
return to non-edit note, the others should stay open.

jeremies:
Close #1302


asbiin:
Thank you @jeremies

PR #3851, opened 2020-04-16
feat: carddav client

### LINKED ISSUE #2738: CardDAV Client Support

asbiin:
This depends on #3749
It will close #2738

This adds the ability to add an external carddav source, and sync contacts in a new address book.
There are 2 steps
- **Add a new address book**: this is currently done with a new Command `monica:newaddressbooksubscription`. A dedicated UI will be added later in another PR.
- **Sync the contacts**: this is done by a new schedule command, based on a frequency

__Limitations__
- Second address book is not displayed in Monica today. This has to be done in another PR
- The external carddav server must respect standards.
    For instance, this is not the case of google api... Hopefully I managed how to handle this specific case, but we will not add further exceptions.
- When adding multiple address books (Monica comes with at least 1 default address book per account), there is no way to "merge" duplicates (= same contact in different address books). This has to be done in another PR
- There is no way to handle errors: if an address book sync has an error, it's logged in the app logs, but this it not accessible by the user, and new sync will be triggered again and again.


malinanroonie (approved):

malinanroonie (approved):
Jim

oliverjhyde:
Just commenting to say I'm happy to test this feature, this is all I'm waiting for before starting to use it in earnest

oliverjhyde:
Is it possible to get a status update on this?

asbiin:
@oliverjhyde I have to restart work on that, it's on hold for a while

app/Jobs/SynchronizeAddressBooks.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @var AddressBookSubscription
+     */
+    public $subscription;
+
+    /**
+     * Create a new job instance.
+     *
+     * @param  AddressBookSubscription  $addressBookSubscription
```
asbiin, line 25:
```suggestion
     * @param  AddressBookSubscription  $subscription
```

app/Console/Commands/DavClientsUpdate.php
```diff
@@ -0,0 +1,42 @@
...
+     * @return void
+     */
+    public function handle()
+    {
+        $subscriptions = AddressBookSubscription::active()->get();
+
+        $now = now();
+        $subscriptions->filter(function ($subscription) use ($now) {
+            return is_null($subscription->last_synchronized_at)
+                || $subscription->last_synchronized_at->addMinutes($subscription->frequency)->lessThan($now);
```
djaiss, line 37: This is not easily readable, even if it's "elegant". I have to think hard to get what the condition is.

app/Helpers/SearchHelper.php
```diff
@@ -44,7 +44,7 @@ public static function searchContacts(string $needle, string $orderByColumn, str
             });
 
             return $b->addressBook($accountId, $addressBookName)
```
djaiss, line 46: `$b` is not a great variable name (even though it wasn't written in this PR).

app/Jobs/SynchronizeAddressBooks.php
```diff
@@ -0,0 +1,54 @@
...
+     * @return void
+     */
+    public function handle()
+    {
+        try {
+            app(SynchronizeAddressBook::class)->execute([
+                'account_id' => $this->subscription->account_id,
+                'user_id' => $this->subscription->user_id,
+                'addressbook_subscription_id' => $this->subscription->id,
+                //'force' => true,
```
djaiss, line 46: Needed?

app/Services/DavClient/SynchronizeAddressBook.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @param  array  $data
+     * @return void
+     */
+    public function execute(array $data, GuzzleClient $httpClient = null)
+    {
+        $this->validate($data);
+
+        $account = Account::find($data['account_id']);
+        if (AccountHelper::hasReachedContactLimit($account)
```
djaiss, line 45: If `$account` doesn't exist above, I don’t think this will work.

Shouldn't we do `findOrFail` instead?

app/Services/DavClient/Utils/AddressBookContactsUpdater.php
```diff
@@ -0,0 +1,126 @@
...
+                        'href' => $href,
+                        'etag' => $contact[200]['{DAV:}getetag'],
+                        'vcard' => $contact[200]['{'.CardDAVPlugin::NS_CARDDAV.'}address-data'],
+                    ];
+                });
+        })->then(function ($contacts) {
+            foreach ($contacts as $contact) {
+                $this->syncLocalContact($contact['href'], $contact['etag'], $contact['vcard']);
+            }
+        });
```
djaiss, line 79: 🤯  Now that's a complex method.

djaiss (changes requested):
What a great PR.

So hard to understand, though. I don't understand carddav so I only understood the code itself, not what it does.

app/Services/DavClient/SynchronizeAddressBook.php
asbiin: actually, it's certain the account exist already, thanks to the `validate` 2 lines below.

app/Jobs/SynchronizeAddressBooks.php
asbiin: fixed

app/Helpers/SearchHelper.php
asbiin: I've fixed here

app/Console/Commands/DavClientsUpdate.php
asbiin: I've moved that in another function, I hope it's more readable now.

app/Services/DavClient/Utils/AddressBookContactsUpdater.php
asbiin: Well, I agree, what do you propose?

cstrat:
Do you know if this update will address #5150 ??

sitarane:
Hey, sorry to reanimate this thread, but I've been trying to figure out how to use this feature for a few days and I can't get my head around it.
If I understand well, this merged PR means that Monica can now get contacts from a third party CardDAV server, my usecase would be Nextcloud.
How do I point Monica to a CardDAV server?

daniele-athome:
I can't find any documentation about it, but a user tried: #5654

daniele-athome:
By the way is the synchronization one-way? Meaning CalDAV data is only read from the server, right?

cstrat:
I am finding strange behaviour where my Monica contacts on my iPhone over time simply duplicate themselves.
I need to disconnect Monica from my Contacts, let it purge my contact list, then re-enable it and sync again.

PR #3867, opened 2020-04-19
docs: add a note to the installation instructions stating that HTTPS will be forced when in production

montali:
I followed the tutorial, and lost hours trying to figure out why Monica kept fetching resources with HTTPS, even though the APP_URL was on HTTP. Just a line to save a bro out.


asbiin:
Hi @montali 
This is a problem of documentation, I will add this:

- HTTPS is enforce when you set `APP_ENV` to `production`.
  You can keep the http working by setting `APP_ENV` to `local`

- `APP_URL` is **only** used for the outgoing emails. It has **no** effect at all at web runtime.

montali:
Hi @asbiin , thanks for your fast response!
Even if `APP_URL` isn't changing anything, I think we should change the tutorial: using the provided Apache configuration (which listens on port 80) Monica doesn't work. Maybe it would be better to just insert an SSL-compatible configuration for Apache in the tutorial?

asbiin:
OK, I will rewrite the instructions a little bit, like this:
- if you're on a production server / a server on the internet: set APP_ENV=production, and set ssl on the web server, using a proxy apache/nginx server, letsencrypt, or any other way
- you're on a local server: set either APP_ENV=local or production, and you can use http or https.

I will try to find a generic way to set ssl on apache/nginx, but I don't think we should include too much server-config documentation.

montali:
That sounds perfect! Thanks mate.

PR #3901, opened 2020-04-24
chore: move currencies seeder as migration

asbiin:
Currencies seeder was moved from migrations to seeder in #1956
It was not a good move.

1. Seeding database are for test data
See https://laravel.com/docs/7.x/seeding : 
> Laravel includes a simple method of seeding your database with test data using seed classes. 

2. If you install a new instance, running `php artisan monica:update` should be enough.
    If you miss to run `db:seed`, the currencies will not be there.
3. You should not need to run `setup:production` once after first run.
    The command must be run after first install. It can be easily missed.


asbiin:
@djaiss what do you think of this?

asbiin:
@djaiss what do you think of this?

PR #3906, opened 2020-04-25
feat: support float amount values, and fix debts and gifts amount forms

### LINKED ISSUE #333: Allow float in debt amount field
When you enter a debt you have to enter it as an integer.
However since this represents money a float would make more sense.

### LINKED ISSUE #3761: Store amount (debt, gift cost) as int + currency
Accounts should be stored as integer, and with currency and MoneyHelper, it can be converted in string

asbiin:
This fix #333
This also fix #3761, storing amount in database as integer value, rather that decimal.
Every currencies don't use a 2 digits precision, and using decimal/float can lead to calculation errors.

This introduce the notions of monetary amount vs exchange value:
- exchange value is the amount displayed for the user, or entered by a user, like 12.34 (or 12,34)
- monetary amount is the amount stored in the database, so as an integer.


app/Helpers/MoneyHelper.php
```diff
@@ -40,14 +43,128 @@ public static function format($amount, Currency $currency = null)
...
+    }
+
+    /**
+     * Format a monetary value as exchange value.
+     *
+     * @param int|null $amount
+     * @param Currency|int|null $currency
+     * @return float
+     */
+    public static function exchangeValue($amount, $currency): float
```
djaiss, line 121: I don't understand this concept of exchange value.

app/Helpers/MoneyHelper.php
```diff
@@ -40,14 +43,128 @@ public static function format($amount, Currency $currency = null)
...
+        return (float) $moneyFormatter->format($money);
+    }
+
+    /**
+     * Get unit adjustement value for the currency.
+     *
+     * @param Currency|int|null $currency
+     * @return int
+     */
+    public static function unitAdjustment($currency): int
```
djaiss, line 138: I also don't understand this.

asbiin:
The reason for this PR is explain in this tweet: https://twitter.com/michaeldyrynda/status/1253516294571175938

In addition, as you may now, some currency have a 2 digit precision (like $ 12.34), but others could have 3 digits (dinar), 1 digit (Cambodian Riel), or 0.
So we really should store amount as an integer (like 1234 for $12.34), and that actually the format required by the \Money\Money class. It will reduce possible bugs, complexity, and improve accuracy.

app/Helpers/MoneyHelper.php
asbiin: This method allows to convert from display value to exchange value giving the adjustment value needed to convert.
So for € or $, it will return 100 (for 2 digits precision), for Dinar, it will return 1000 (3 digits ...)
So you can divide your amount in exchange value by this number to convert (or otherwise ...).

And if you see, this was already used previously mixed in the `format` function.

app/Helpers/MoneyHelper.php
asbiin: The exchange value is the amount stored as an integer, like 1234 for $12.34

PR #3942, opened 2020-05-02
feat: crop contact photos on upload

### LINKED ISSUE #1775: Positioning contact photos
After uploading a photo/avatar, I'd like to choose the area of the photo to show. 

photo-positioning-monica

jeremies:
Close #1775 

Now when you upload a contact photo you can crop it. The UI looks like this:

- You click **Choose file** and select a file from the your device:
image

- A modal opens that lets you crop the photo:
image

- When you click **Done** the cropped photo is shown below the button:
image


resources/js/app.js
```diff
@@ -28,6 +28,10 @@ import VueClipboard from 'vue-clipboard2';
 VueClipboard.config.autoSetContainer = true;
 Vue.use(VueClipboard);
 
+// Dependency of vuejs-clipper
+import VueRx from 'vue-rx';
+Vue.use(VueRx);
+
```
asbiin, line 34: I don't see where the plugin is used? If it's not used globally, is it good enough  to add it in the SetAvatar component?

asbiin:
Unfortunately, we use purgecss in Monica to limit size of the css, and it removes the css from vuejs-clipper.
So this will not work without removing purgecss.

asbiin:
I've added all needed classes to the exclude field, and it's working now.

asbiin:
OK I approve this PR

PR #3945, opened 2020-05-03
docs: rename index.md to readme.md

asbiin:
it's better on github


timjrobinson:
This is causing the link on the homepage to 404 now. Fixed in https://github.com/monicahq/marketing_site/pull/174

PR #3984, opened 2020-05-09
feat: add a queue restart command in monica:update command

asbiin:


asbiin:
In fact this is not a good idea. `php artisan queue:restart` must be run, but after the deploy is finished. Indeed, the queue:restart should not be run before the app has finished to be deploy, or else the restart will be run on the old source code, which has no interest.

The normal process is:
- update source code
- run upgrade
- deploy
- run queue:restart

So it's the deploy script that should run the restart, and it depends on how the app is deployed, and cannot be automatised in the upgrade process.

PR #4069, opened 2020-05-26
Fix mistake in the dates in CHANGELOG.md: 2020, not 2019

Flimm:
Small fix that is easier to show than describe


asbiin:
Thanks @Flimm

PR #4124, opened 2020-06-09
docs: add info that only mysql is working atm

duracell:
Because I thought pgsql is working and looking at the issues other people have the same problem, at least put this in here to warn people not to change the DB_CONNECTION setting to something else (e.g. pgsql, sqlite or other).


asbiin:
Thanks @duracell

PR #4131, opened 2020-06-11
fix: enable mod_rewrite in dockerfile.dev

chrootlogin:
Minimal change to enable mod_rewrite in Dockerfile.dev as in the normal Dockerfile


asbiin:
Thanks @chrootlogin

PR #4133, opened 2020-06-11
fix: add message.js and intellij *.iml files to .gitignore

chrootlogin:
small fix for .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ monicadump.sql
 .phpunit.result.cache
 .composer
 .idea
+*.iml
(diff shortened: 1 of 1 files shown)


asbiin:
I don't think that relevant.
- `/public/messages.js`: this file does not exist
- `*.iml`: use a global gitignore instead: https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

chrootlogin:
The /public/messages.js is automatically generated, if you use Docker as development tool.

PR #4134, opened 2020-06-12
fix: Generic, Debian, Ubuntu installation cron job instructions

RussellAult:
- the existing crontab instructions don't correctly deal with the stdout output of the schedule:run command, resulting in a new mail message ever minute for whichever user is running the webserver; this will fix that according to https://laravel.com/docs/7.x/scheduling
- the existing cron job instructions for Ubuntu installed the job into a location that is technically reserved for apt packages as opposed to in the www-data user's crontab


asbiin:
Thank you @RussellAult !

RussellAult:
You're welcome!

PR #4138, opened 2020-06-14
feat: Set max length to fake real text

P-Shakibafar:
because in migration set length for task title


asbiin:
we can reopen this when you are ready ...

PR #4246, opened 2020-07-20
fix: allow decimals in GPS coordinates

Zegnat:
Minimal change to enable adding GPS coordinates to addresses from the browser.

Several browsers, including latest Firefox stable, mark fields of `type="number"` as invalid when they include decimals. GPS coordinates are basically always going to include decimals. The simplest fix is to set a `step` value.

The `step` value set in this PR is based on how accurate the GPS goes in Google Maps URLs. At max zoom (21) Google uses 7 decimals.

A more complete solution may be to drop `type="number"` entirely. See for example [What to Use Instead of Number Inputs](https://css-tricks.com/what-to-use-instead-of-number-inputs/). But I leave that as an exercise to someone else :wink: I just needed to be able to enter GPS coordinates for my contacts and thought I would submit my fix upstream.


asbiin:
Nice work, thank you @Zegnat

PR #4283, opened 2020-08-02
docs: Bump recommended git checkout tag to latest

l1n:
this is a four character change not sure if this warrants adding to CONTRIBUTORS :P


asbiin:
Thank you @l1n 
I'm not sure we will update the version for every release, but why not!

PR #4310, opened 2020-08-10
chore: correct dav connection credential help text

nebulade:
Some of our users were confused that the password didn't work despite being mentioned within the app. The devs have confirmed that only api tokens are supposed to work. This likely needs more fixes for other languages as well.


asbiin:
Thank you @nebulade !

PR #4354, opened 2020-08-24
fix: fix tags not being stored when pressing enter

djaiss:
This PR does two things:

* it removes the blank line that appears below the input text when there are no results
* it saves entries when we press Enter.


asbiin (approved):
Nice !

PR #4416, opened 2020-09-07
feat: [3.x] add account and contact models

djaiss:


yarn.lock
```diff
@@ -0,0 +1,7031 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
asbiin, line 1: We don't need yarn anymore, we can use npm directly, and the file `package-lock.json` is here to ensure package versions.

asbiin (changes requested):
See my comment

yarn.lock
djaiss: We don’t use Yarn anymore?

asbiin: With the 3.X branch, we could swith to npm yes

Issue #4308, opened 2020-08-09
Add a confirmation to delete a journal entry


asbiin:
_Feedback from a user_

The "Delete" action for a journal entry is very close to the "Edit" link and you can easily delete a journal entry and lose all of the info as there's no confirmation of the delete action.

We should at least add a confirmation on the delete action.

TeddyBear06:
Hi @asbiin ,

Do you think that a confirm box like this one https://imgur.com/a/IhJl7Fi can do the job ?

I already edited the _trash()_ method of _JournalContentEntry.vue_ to do so.

asbiin:
@TeddyBear06 Thank you. That would do the job yes ! :+1:

PR #4514, opened 2020-10-03
feat: Add a confirmation to delete a journal entry #4308

simonardejr:
Implementing the native confirmation dialog when click on `delete` on journal entries list. This was requested on #4308


asbiin (changes requested):
Thank you @simonardejr 
However, we don't use `confirm` function anymore. See usage of the `confirm` component (`<confirm>`) in some resource blade files.

simonardejr:
Ah, ok! give me some minutes and I'll update this :grin:

simonardejr:
Don't know if I've done everything right (fetching, merge, and stuff), but let me know if I need to do anything else :grin:

asbiin:
Perfect @simonardejr :+1: 
And thanks also @TeddyBear06  ;)

Issue #4541, opened 2020-10-12
Invalid Validation Messages


kavanpancholi:
On any validation message, it shows something like the following message "validation.vue.required" instead of proper validation message.

Screenshot:
![image](https://user-images.githubusercontent.com/5137072/95723031-4fe37f80-0c92-11eb-88cc-23c4d6e5cc52.png)

Root Cause:
In the code, validation message uses vue-i18n plugin to show all the locale messages.
And there is no "vue" key defined in the json file (`public/js/langs/en.json`) and because of that it shows plain text instead of message from locale file.

Daibushi:
Can i fix It?

kavanpancholi:
@Daibushi I have already tried to fix it. I've created this PR https://github.com/monicahq/monica/pull/4542.
Waiting for the maintainer to approve it.

Daibushi:
Ok.

asbiin:
fixed

PR #4542, opened 2020-10-12
fix: #4541 change en locale lang variable syntax to support vue-i18n

kavanpancholi:
Solves Issue: https://github.com/monicahq/monica/issues/4541

There are 2 issues in regards to validation messages.
1. https://github.com/monicahq/monica/blob/18fadb77ce7fb4b46fd71ae8205127b9e8c9581d/resources/js/components/partials/form/Select.vue#L127 In the code, validation messages has vue key specified but there is no vue key defined in the localization files. https://github.com/monicahq/monica/blob/18fadb77ce7fb4b46fd71ae8205127b9e8c9581d/resources/lang/en/validation.php#L105

2. Laravel Locale File takes variable in `:attribute` syntax, but vue-i18n plugin takes it in `{attribute}` syntax.
That is the reason, the attribute couldn't take dynamic value.

This PR maps attribute in line with laravel validation attributes and changes variable syntax to support vue-i18n arguments.

**To-Do before Deployment**:
- Please execute php artisan lang:generate command to generate multi-lang json file.

**Note**: For now, I've made changes in english language files only just to make the small changes in the PR.
Once this changes gets approved, I can do those changes in other languages as well.


kavanpancholi:
Hey, @asbiin Can you please review this PR?
And if you feel like merging it, can you please add "hacktoberfest-accepted" label to it?
Thank you. :+1: 
Cheers.

kavanpancholi:
Hi @asbiin Can you please review this PR?
And if that is not possible, can you add "hacktoberfest-accepted" label to it?
Thank you :)

asbiin:
Thank you @kavanpancholi 
I really appreciate the effort, this is a good work!
But to be honest, I don't think we can do such changes.

The validation messages comes from Laravel package. They are somtimes updated upstream.
The messages can be used in php, where the ":attribute" format is required.
In addition, we use `laravel-lang/lang` (formerly `caouecs/laravel-lang`) to update the translation automatically, and add new languages. In this package, the ":attribute" format is a default.

It would be so much easier is both php/laravel and vue-i18n could use the same syntax. I nether though about it, but maybe we could use the same extension we've done in [`resources/js/pluralization.js`](https://github.com/monicahq/monica/blob/master/resources/js/pluralization.js) file ...

What do you think about this?

asbiin:
@kavanpancholi anyways, I've still added the label

kavanpancholi:
@asbiin Noted! Thank you for adding the label.
I understand that we can't replace :attribute

Issue #4644, opened 2020-12-03
Installation script doesn't check for ImageMagick extension


smaffulli:
The Installation process went smoothly following the instructions but at first login I got a `Whoops` message

```You need to install the imagick extension to use this back end```

A simple check at install time for imagick extension would make the experience a lot more pleasant. Or at the very list make it an explicit [requirement](https://github.com/smaffulli/monica/tree/master/docs/installation#requirements).

arisAlexis:
I just submitted a pull request #4646 to solve this

PR #4646, opened 2020-12-04
docs: update ubuntu.md to add php-imagick dependency

arisAlexis:
default install with dashboard doesn't run without php-imagick and shows an error. Solves #4644


asbiin:
Thank you @arisAlexis

Issue #4673, opened 2020-12-08
“Gift offered” should be “Gift given”


mgrhm:

mgrhm:
Closed by merge of #4742.

PR #4681, opened 2020-12-10
fix: Wording "Gift offered" to "Gift given" (#4673)

mgrhm:
Fixes the wording issue raised in issue #4673.


asbiin (changes requested):
Thank you for your submission.
Unfortunately, you made too much changes.
The only locale file to modify is the `resources/lang/en/people.php` - then regenerate the assets like you did. The other locale file will be updated after localization process.

mgrhm:
I've updated just the `/en/` and `/en-GB/` `people.php` files now and submitted a new commit.

resources/lang/en-GB/people.php
```diff
@@ -308,7 +308,7 @@
     'gifts_delete_cta' => 'Delete',
     'gifts_add_title' => 'Gift management for :name',
     'gifts_add_gift_idea' => 'Gift idea',
-    'gifts_add_gift_already_offered' => 'Gift offered',
+    'gifts_add_gift_already_offered' => 'Gift given',
```
asbiin, line 311: Sorry but `en-GB` is also a translation, and must not be edited (see the header of the file).

asbiin (changes requested):

resources/lang/en-GB/people.php
mgrhm: Okay. I’ve made a commit that _only_ changes the `en` file, and pushed it.

PR #4736, opened 2020-12-23
docs: add APP_URL to Heroku template

quentinms:
This should make it easier for people using Heroku for their emails to link to their app instead of the default URL ([http://localhost](https://github.com/monicahq/monica/blob/10f7f197309112836f39279de27e82ae5270ed4f/config/app.php#L66)).


asbiin:
OK good. However, the value of the url can only be known *after* deployment.

asbiin:
@quentinms Thank you!

PR #4741, opened 2020-12-27
style: minor rewording of dashboard.php for consistency.

mgrhm:


resources/lang/en/dashboard.php
```diff
@@ -1,26 +1,26 @@
 <?php
 
 /**
- * ⚠️ Edition not allowed except for 'en' language.
+ * ⚠️ Editing not allowed except for 'en' language.
```
asbiin, line 4: This could also be changed in all other lang files

asbiin (approved):
Thank you for this style fix!

PR #4742, opened 2020-12-27
style: change "offered" to "given" for gifts

mgrhm:


mgrhm:
This fixes two other instances of ‘offered’, which the previous pull request missed.

Issue #4564, opened 2020-10-22
Implement apple web application


maik-s:
Since Monica does not have a dedicated smart phone application, it would be nice to configure it as a web application, [as mentioned in the apple developer guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html) 
By this, a bookmark to the homescreen could have the original monica logo, instead of a random screenshot of the page.

Furthermore, it would feel more like a native application, as it is contained into a dedicated webview instead of the every-day browser.

johan456789:
I guess this should be closed?

PR #4743, opened 2020-12-27
feat: add Apple icons for use when bookmarking.

mgrhm:
Adds icons for when Monica is added as a bookmark to the home screen on an iOS device.  Currently, it defaults to a miniaturised screenshot.


asbiin (changes requested):
Some remarks:
- Do we need to add these 11 image files? All of them are not used in the skeleton blade file.
- I don't agree with the names of the files, which don't apply to apple only.
- It would be better to use the logo image (you used it in `apple-icon.png` file), without the "Monica" name

mgrhm:
Hi Alexis,

Thanks for the feedback.  I’ve updated the code and made a new commit to address your comments.

To respond to them directly:

> Do we need to add these 11 image files? All of them are not used in the skeleton blade file.

No. I've slimmed this down to four images now —- most of the smaller images were for older devices that are unlikely to be in wide use now.

> I don't agree with the names of the files, which don't apply to apple only.

These _are_ explicitly used for Apple devices, hence the (now amended) filenames and the very specific icon sizes.  I've worked on this revised commit based on the explanation in [this Apple guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html).

I'm researching how to do this for Android devices too, and will submit this as a new pull when I've worked it out.

> It would be better to use the logo image (you used it in apple-icon.png file), without the "Monica" name.

It looks like the icons I intended to include in the third commit didn't get included.  This is fixed, and they're all the same: white background with just the logo on it.

I've also added another line so that it defaults to the name 'Monica' when added to the home screen as a web app (line XXX).

Hope this is satisfactory.

mgrhm:
I’ve just seen issue #4564. This pull request would resolve that issue.

asbiin (approved):
That's great, thank you @mgrhm 
However, wouldn't it be better to name the 4 files with a transparent name, like "monica-64.png", "monica-72.png", etc? Maybe they will be reused some day for another usage.

mgrhm:
When naming the files, I followed the [Apple documentation’s recommendations](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html).  I’m not sure what effect it would have giving them different filenames.

asbiin:
Great! Thank you @mgrhm

PR #4757, opened 2020-12-31
style: change language files warning message

mgrhm:
As [requested by @asbiin](https://github.com/monicahq/monica/pull/4741#discussion_r550456862).

### DIFF
--- a/resources/lang/ar/app.php
+++ b/resources/lang/ar/app.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * ⚠️ Edition not allowed except for 'en' language.
+ * ⚠️ Editing not allowed except for 'en' language.
  *
  * @see https://github.com/monicahq/monica/blob/master/docs/contribute/translate.md for translations.
  */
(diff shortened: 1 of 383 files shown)


asbiin:
Thank you @mgrhm - but as explained in each of the files, the modification must be done *only* in the "en" files. All others language files must be kept untouched!

mgrhm:
The only thing changed is the opening comment “Edition is not allowed” – which I’ve corrected to “Editing is not allowed”.

I’d taken your request to mean that you wanted this fixed globally.  The comment itself is in English and “editing” is the correct word here.

Do you want this left uncorrected as “Edition” in the non-`en` files?  I can revert and fix only the `en` files if that’s the case.

mgrhm:
I've reverted the commit and applied the correction to _only_ the `en` files.

I’m still of the opinion that the comment, which is English language metadata on the non-`en` files, should be corrected everywhere.  If you agree, I'll submit a separate pull that makes the correction globally.

asbiin:
That's better!
The change will be made in the "en" files, then the translation process, with crowdin, will update all the translated files automatically, in a separate PR (see #4687 )

mgrhm:
Thanks, @asbiin – I wasn’t aware that was how Crowding worked.

PR #4779, opened 2021-01-09
feat: export data as json format

### LINKED ISSUE #32: Add export function in JSON

### LINKED ISSUE #355: Create UUIDs for data types to be shared via API / Webhooks
With UUIDs Monica and third-party applications can exchange information about the same user/contact/event/etc.

asbiin:
This resolves #32
It's also related to #3104
It will also close #355

The idea is to export datas with json format, with a flat, simple structure
There are 1 root data "account" that contains "users", "contacts", and other datas for config: activityTypes, activityTypesCategories, etc.
All datas related to contact are detailed in the contact itself.

Photos and documents are serialized directly in the json file (which is not the case with the sql export)


config/monica.php
```diff
@@ -285,4 +285,14 @@
     */
     'adorable_api' => env('ADORABLE_API', 'https://api.hello-avatar.com/adorables/'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Export limit
+    |--------------------------------------------------------------------------
+    |
+    | Maximum number of exports.
```
djaiss, line 293: This is not clear. Max number of exports... total? at the same time?

Also, you don't provide an env variable for this setting. Is this on purpose?

djaiss (changes requested):
This is an amazing job. Now that the code review is done, I'll test it.

Also: very good idea to create ExportResources objects.

app/Http/Controllers/SettingsController.php
```diff
@@ -183,15 +192,85 @@ public function export()
...
+     * @param  string  $type
+     * @return ExportJob
+     */
+    private function newExport(string $type): ExportJob
+    {
+        $exports = ExportJob::where([
+            'account_id' => auth()->user()->account_id,
+            'user_id' => auth()->user()->id,
+        ])
+            ->orderBy('created_at');
```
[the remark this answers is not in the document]
asbiin: Not required, but it's better indeed, to avoid multiple queries. Thanks

app/Http/Controllers/SettingsController.php
```diff
@@ -183,15 +192,85 @@ public function export()
...
-        return response()
-            ->download($adapter->getPathPrefix().$path, "monica-export.$exportdate.sql")
-            ->deleteFileAfterSend(true);
+    /**
+     * Create a new ExportJob.
+     *
+     * @param  string  $type
+     * @return ExportJob
+     */
+    private function newExport(string $type): ExportJob
```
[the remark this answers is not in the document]
asbiin: Yeah good idea!

config/monica.php
asbiin: I've updated the comment

djaiss:
@asbiin In an empty account, I added one contact, then exported my data in Json. It failed, and I have no idea why.

asbiin:
> @asbiin In an empty account, I added one contact, then exported my data in Json. It failed, and I have no idea why.

I cannot reproduce it, it works fine with me. Don't you have any exception in your log file?

public/changelog.json
```diff
@@ -1,5 +1,10 @@
 {
     "entries": [
+        {
+            "date": "Jan 11, 2022",
+            "title": "Preview feature: export data as json",
+            "description": "We are experimenting with exporting data as json. [Try it now](/settings/export). This feature is in preview mode, please give us feedback on [these GitHub discussion](https://github.com/monicahq/monica/discussions/5824)."
```
djaiss, line 6:
```suggestion
            "description": "We are experimenting with exporting data as json. [Try it now](/settings/export). This feature is in preview mode, please give us feedback on [this GitHub discussion](https://github.com/monicahq/monica/discussions/5824)."
```

djaiss (changes requested):

Issue #1192, opened 2018-04-18
Add dark mode


djaiss:
Twitter does it well.
Zulip also.

Check:

![image](https://user-images.githubusercontent.com/61099/38937270-d32fbc0a-42f0-11e8-9708-0dfac70ede6b.png)

pqhf5kd:
I imagine this is a low priority, but if it does get implented, can it use the new `prefers-color-scheme` query? More details [here](https://blog.iconfactory.com/2018/10/dark-mode-and-css/)

mgrhm:
@pqhf5kd: Yeah, it can be implemented with `@media (prefers-color-scheme: dark)` in the CSS. It’s actually pretty easy to do it – but I’ve only ever done it on very small static sites with small, simple CSS files.  It could very easily be much more complicated to implement in Monica.

vergissberlin:
@mgrhm I have some experiences with it. The idea is to use custom properties in combination withe the mentioned media query. For colours, this pallet could be an inspiration: https://color.adobe.com/de/Dark-UI-color-theme-7388196/

I am interested to help more.

mgrhm:
Hi @vergissberlin: I’d be interested in working on this with you. I’ve only ever done it for really small sites (like [my own site](https://mgrah.am/)) which have much simpler CSS than Monica appears to.

I like the palette you’ve suggested, but feel there needs to be a higher contrast pale colour for the body text – `#A1B4C4` may be too dark against a `#202529` background at the size/weight of text Monica uses.  (With all that said, it does get a [pass on WebAIM’s contrast checker](https://webaim.org/resources/contrastchecker/?fcolor=A1B4C4&bcolor=202529).)

vergissberlin:
@mgrhm Great. When we use custom properties, the colours are changeable very easily. So, I don't care too much about it. I just want to get rid of the brightness in my eyes.  😎  

## Current status
I've seen, that this project uses _tachyons_ and parts of bootstrap. Both frameworks using custom properties already ([tachyons](http://tachyons.io/docs/table-of-properties/), [bootstrap](https://getbootstrap.com/docs/4.0/getting-started/theming/#css-variables)) which is great. I never heard of _tachyons_ before, but I've found a repository with modified properties for a dark mode: https://github.com/ocd-sg/ui-tachyons-dark/blob/master/src/index.css

## Starting point
After a short look into the code, it seems that Monica currently don't uses custom properties at all.
To replace all colour values with custom properties, is a good first starting point.

[...]

PR #4797, opened 2021-01-13
style: Colour simplification

mgrhm:
This reduces the colour count from over 80 to 21.

This is the first half of solving #1192 -- once merged, we'll start work on implementing a dark mode based on this simplified colour scheme.


resources/sass/_variables.scss
```diff
@@ -1,17 +1,17 @@
...
+// $blue:
+// $indigo:
+// $purple:
+// $pink:
+// $red:
+// $orange:
+// $yellow:
+// $green:
+// $teal:
+// $cyan:
```
asbiin, line 11: Shouldn't we define the used colors as variables here, in replacement of these ones, so future use are consistent?

asbiin (changes requested):
I've tested it, it looks fine to me!

resources/sass/_variables.scss
mgrhm: Would we be able to do that _and_ integrate the dark-mode variables [as suggested by @vergissberling](https://github.com/monicahq/monica/issues/1192#issuecomment-754507231)?  I'd actually planned to move over to using `--var`-style names in the dark-mode conversion pull using that technique.

I don't really understand SASS, so it's possible this can be done.  I just don't know enough to say either way.

vladiiancu: Actually, CSS custom properties (or variables as they are commonly called) are available in vanilla CSS. No preprocessor like SASS needed. 

With the notable exception of Internet Explorer which lacks support, everything else is compatible. 

And, yes, they would be quite helpful for an easier dark-mode implementation. Good inspiration from [this article](https://mwichary.medium.com/dark-theme-in-a-day-3518dde2955a).

mgrhm: I much prefer CSS variables to SASS's equivalent -- if only because I understand them and have experience of them.  An earlier comment [by @vergissberlin](https://github.com/monicahq/monica/issues/1192#issuecomment-754507231) shows how to use them really efficiently for dark mode.

Naming them may be a little tricky, since the colours switch between light and dark modes, but it's not impossible.

Once we have the simplified colours in master, I’ll start a new branch to do dark-mode.  My grasp of git is too wobbly to try and do it on a fork that isn't forked from master.

asbiin: OK let's merge this one

asbiin (approved):

PR #4798, opened 2021-01-14
feat: add Android icon for use when bookmarking

mgrhm:
Adds icons for when Monica is added as a bookmark to the home screen on an Android device. This commit is an Android equivalent of #4743, and also contains a fix for broken URLs for that pull.


asbiin:
Thank you @mgrhm !

PR #4890, opened 2021-02-19
style: bring back bg color and change the placement of the delete button

djaiss:
Two things in this PR:

- we bring back the background color that was removed in the previous PR. It was too ugly.
- we fix the placement of the Delete button that fucked up the layout

image


asbiin (approved):
good!
It's like the title is truncated, right?

Issue #4801, opened 2021-01-16
Ability to lock Conversations once saved


mgrhm:
As it stands, Conversations are too easy to overwrite when reading them later.  (My cat nearly did it by walking across the laptop a few minutes ago.)

This could be fixed by either the option to 'lock' a conversation’s contents once saved, or a modal 'Edit' option to change from reading to editing on an old Conversation.  The flow for adding new Conversations wouldn't be changed.

mgrhm:
Ideally, this would work just like phone-calls do now.

sktlmarat:
Hi @asbiin @djaiss. 
I started working on this issue and my suggestion is to use the disable feature of the HTML inputs. In the video below I show how I implemented this functionality. Should I make a pull request?

https://user-images.githubusercontent.com/40805789/109667981-8d538700-7b25-11eb-98ac-bd15f4ba14cf.mov

mgrhm:
Love it. The only suggestion I’d make is that the Edit/Save link be moved to the bottom right below the text field for consistency. Otherwise, it’s perfect.

sktlmarat:
Thanks for your feedback! I moved the edit/save button under the form in the new commit.

PR #4932, opened 2021-03-03
feat: 4801 lock conversations

sktlmarat:
this feat for #4801 

- used readonly feature to lock <textarea> tag
- added edit button to unlock messages


asbiin (commented):
I like the idea of locking conversation.
However, the whole workflow is not satisfying.

I would have prefer, when clicking on the discussion title to enter in a "view" only mode, then have the ability to Edit it, which will unlock the entire conversation and other properties, and save the whole changes.
What do you think about that?

sktlmarat:
@asbiin, thank you for your feedback!
I like your suggestion. Your idea seems more convenient for users than mine.
I thought about individual messages and didn't think about conversation as a whole.
I think I will close this PR, and start with a new one. Is it OK?

asbiin:
@sktlmarat Thank you, you can close it, or reuse this one, as you wish.

PR #4937, opened 2021-03-06
docs: fix php7.4 imagick version

asbiin:


gtback:
Thanks for this! I ran into the same problem a week or so when trying to install Monica locally. I have some other notes from my experience that I'll try to make a PR for soon.

PR #4968, opened 2021-03-16
fix: 4912 remove the need for verification when disabling 2fa

### LINKED ISSUE #4912: Can't Disable 2FA if Lost Phone
(Note: you don't need to follow this template, nor to keep headlines or bold sentences - they are just there to guide you. Feel free to delete everything. We review every issue even if we don't immediately respond.)

Thanks for filing an issue and for your interest in the project.

**Describe the bug**
Currently, if a user has 2FA enabled and loses the security device, they can still log in with a recovery code. However they cannot disable or reset 2FA without an existing 2FA code. In this particular case, if a user no longer possesses the associated security device they will not be able to log into Monica without regenerating new recovery codes.

**Additional context**
I believe a 2FA code should *not* be required to disable 2FA for this use case since to log into the dashboard in the first place, the user already needed to present either a 2FA code or recovery code. Therefore, the disable 2FA button should just work. 

P.S. Are there any workarounds for this currently?

SimonVanacco:
Hi !
This removes the need to enter a valid code before disabling 2FA and closes #4912 
As mentioned in the ticket, the current behavior makes it impossible to disable 2FA if you lose your way of generating codes.


asbiin:
That's a security issue.
Disabling 2FA **must** require you to enter a 2FA key, or some security pattern.
If you left an open connection (your browser will remember your connection with monica), then someone could - from your computer - deactivate 2FA.

What we should do is request either a 2FA code, or a recovery code. That way it's secure.

SimonVanacco:
> What we should do is request either a 2FA code, or a recovery code. That way it's secure.

Ok i'll try to do it that. Closing this to start from scratch :)

PR #4970, opened 2021-03-16
feat: allow recovery codes when disabling 2FA

### LINKED ISSUE #4912: Can't Disable 2FA if Lost Phone

SimonVanacco:
This PR allows the user to enter recovery codes to verify his identity when disabling 2FA in the settings.  
This closes #4912  

I've added a new translation message for the field. 
I could have created a new form and a new route but decided that re-using the existing ones posed no security risks, and didn't made the code too bad... tell me if that's not ok for you !


asbiin:
@SimonVanacco I've made some changes to factorize code. Thank you for your contribution!

PR #4976, opened 2021-03-18
style: Minor fix to style and spacing of headers on Dashboard.

mgrhm:
Minor fix so that both "Events" and "Changes" header are bold and evenly spaced.  Added divider between title of "Product changes" and the list for consistency.


asbiin (approved):
@mgrhm Thank you!

PR #4977, opened 2021-03-18
style: minor fix for spacing on subheadings between emoji and text

mgrhm:
As with #4976, this pull fixes inconsistent spacing between emojis and subheadings.


asbiin (approved):
Nice job!

PR #4985, opened 2021-03-19
fix: sort and group relationships by relationship type

### LINKED ISSUE #2395: Sort Relationships

dragosh635:
FIX #2395
Sort and group the relationship by relationshipType
Group the relationships and then show them all based on the group

@djaiss
UI CHANGES: removed the silver badge with the relationship type name localized
[Print Screen here](https://prnt.sc/10pvop5)

Please let me know if I need to change something so that my PR gets merged. 
I'm kind of a begginer in the opern source world :)

Thanks!
Dragos


dragosh635:
Hello, I'm trying to make this PR to pass the Build Assets check and I cannot go by this error [print screen here](https://prnt.sc/10q0caa )
Do you guys have any ideea why is this happening ? 
I'ved already run yarn lint:all and yarn run prod but with no effect.

Thank you

asbiin (approved):
That's great! Thank you @dragosh635

asbiin:
> Hello, I'm trying to make this PR to pass the Build Assets check and I cannot go by this error [print screen here](https://prnt.sc/10q0caa)
> Do you guys have any ideea why is this happening ?
> I'ved already run yarn lint:all and yarn run prod but with no effect.
> 
> Thank you

This build process is sometimes a little bit tricky - we'll get rid of it soon.
I've rebuild the assets myself.

PR #5000, opened 2021-03-25
style: Minor fixes to wording on people.php.

mgrhm:


resources/lang/en/people.php
```diff
@@ -10,28 +10,28 @@
 
     //index
     'people_not_found' => 'Contact not found',
-    'people_list_number_kids' => '1 kid|:count kids',
+    'people_list_number_kids' => 'One child|:count children',
```
asbiin, line 13: I don't see any reason not to keep the number even for "1"
```suggestion
    'people_list_number_kids' => ':count child|:count children',
```

resources/lang/en/people.php
```diff
@@ -10,28 +10,28 @@
...
     'people_list_last_updated' => 'Last consulted:',
-    'people_list_number_reminders' => '1 reminder|:count reminders',
+    'people_list_number_reminders' => 'One reminder|:count reminders',
```
asbiin, line 15: I don't see any reason not to keep the number even for "1"
```suggestion
    'people_list_number_reminders' => '1 reminder|:count reminders',
```

resources/lang/en/people.php
```diff
@@ -10,28 +10,28 @@
...
     'people_list_blank_title' => 'You don’t have anyone in your account yet',
     'people_list_blank_cta' => 'Add someone',
     'people_list_sort' => 'Sort',
-    'people_list_stats' => '1 contact|:count contacts',
+    'people_list_stats' => 'One contact|:count contacts',
```
asbiin, line 19: I don't see any reason not to keep the number even for "1"
```suggestion
    'people_list_stats' => '1 contact|:count contacts',
```

resources/lang/en/people.php
```diff
@@ -10,28 +10,28 @@
...
     'people_list_lastnameAZ' => 'Sort by last name A → Z',
     'people_list_lastnameZA' => 'Sort by last name Z → A',
-    'people_list_lastactivitydateNewtoOld' => 'Sort by last activity date newest to oldest',
-    'people_list_lastactivitydateOldtoNew' => 'Sort by last activity date oldest to newest',
+    'people_list_lastactivitydateNewtoOld' => 'Sort by last activity date, newest to oldest',
+    'people_list_lastactivitydateOldtoNew' => 'Sort by last activity date, oldest to newest',
     'people_list_filter_tag' => 'Showing all the contacts tagged with',
     'people_list_clear_filter' => 'Clear filter',
-    'people_list_contacts_per_tags' => '1 contact|:count contacts',
+    'people_list_contacts_per_tags' => 'One contact|:count contacts',
```
asbiin, line 28: I don't see any reason not to keep the number even for "1"
```suggestion
    'people_list_contacts_per_tags' => '1 contact|:count contacts',
```

asbiin (changes requested):
Thank you @mgrhm 
However, I have some suggestion changes

resources/lang/en/people.php
mgrhm: Most English style guides recommend using words for numbers below ten, and especially when they start a sentence.  I don’t know how to implement that for two through nine, so I split the difference and did it with ‘one’.

resources/lang/en/people.php
```diff
@@ -112,7 +112,7 @@
     'call_emotions' => 'Emotions:',
 
     // Conversation
-    'conversation_blank' => 'Record conversations you have with :name on social media, SMS, ...',
+    'conversation_blank' => 'Record conversations you have with :name on social media, SMS…',
```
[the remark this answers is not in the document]
mgrhm: It reads badly to have a comma followed by an ellipsis, and I don’t think that’s correct form in English.

What about _“…social media, SMS, and other media”_?

resources/lang/en/people.php
asbiin: In this case, it's not really a "sentence", but just displayed information.
There are 2 reasons not to change that:
- consistency in the display: if there is 1 or 2 items, I don't want to display it differently. The important part is the number of children. It's easier to see the number of children in '1 child' than in 'One child'.
- translation rules: It's hard to make the translation, and so the contributors in translation, know all the programmatic rules. All languages does not follow english here ('1' for singular version, '0' and '2' or more for plural version), see this [laravel function](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Translation/MessageSelector.php#L110), and I'm afraid all of them will translate that following "One child" as a text, and ":count children" as another text, which is not the case.

resources/lang/en/people.php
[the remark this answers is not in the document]
asbiin: Lets keep your original proposal so, it's fine with me.
`Record conversations you have with :name on social media, SMS…`

asbiin (approved):
Good job @mgrhm Thank you!

asbiin (approved):
There are some tests to be fixed...

mgrhm:
Thanks for fixing this.

PR #5020, opened 2021-03-29
style: Made green button's :hover state more legible.

mgrhm:
The button previously used the standard :hover colour of blue, which was hard to read against the green background.  I've reused the medium yellow used elsewhere to provide visible differentiation and contrast when hovered.


asbiin:
I like it.
image

asbiin (approved):

asbiin:
Thank you @mgrhm

Issue #5025, opened 2021-04-04
remove call to https://fonts.googleapis.com


krswin:
**Is your feature request related to a problem? Please describe.**

A clear and concise description of what the problem is.
* For privacy reasons its important not to connect internet for any type of requests.
* Monica gets CSS from internet API to get CSS
* https://fonts.googleapis.com/css?family=Open Sans:400,600

**Describe the solution you'd like**
* Please include CSS in the code so no need to connect to internet for any type of requests. 

Thank you for your help.

mgrhm:
I've opened pull #5027 to remove two instances where Google Fonts are called.  These are the only two 'Monica native' occurrences of Google Fonts.

I can see several instances of Google Fonts being called in Laravel and other vendor packages, but I haven't removed those.  I think that they'd need to be removed upstream on the package's master branch then merged into Monica from there.

krswin:
@mgrhm Thanks.

asbiin:
Fixed by #5027

PR #5027, opened 2021-04-05
style: Remove google fonts (addresses #5025).

mgrhm:


iloveitaly:
I'm curious why we need to remove these google fonts. Security / load speed?

mgrhm:
In their (original request)[https://github.com/monicahq/monica/issues/5025], @krswin said it's important that their Monica instance doesn't connect to the internet for any types of requests.  I'd assume that they're hosting it locally and don't want Monica making HTTP calls outside their local network.

From my perspective, I can't see a good reason why these are included.  They don't improve the site's function (they're barely used), they add to load time, and they're making requests to Google's servers.  I don't see a benefit to any of those -- and I'd say removing Google from an application I self-host is a plus.

In this pull, I'm removing Google Fonts from an error page.  The page works just fine, and will load faster, without that additional overhead.

Self-hosting webfonts is a bit fiddly to set up, but not that difficult really.  If it's something that Monica wants/needs, we should add the fonts and code so that they're hosted and served locally.  As it stands, though, Monica's font-stack is pretty thorough and doesn't rely on non-system fonts.

iloveitaly:
Makes sense! Thanks for explaining. Definitely in favor of this!

resources/views/errors/layout.blade.php
```diff
@@ -15,10 +13,10 @@
                 margin: 0;
                 padding: 0;
                 width: 100%;
-                color: #B0BEC5;
+                color: #999999;
```
asbiin, line 16: Why changing the color here?

resources/views/errors/layout.blade.php
```diff
@@ -15,10 +13,10 @@
...
                 display: table;
                 font-weight: 100;
-                font-family: 'Lato';
+                font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
```
asbiin, line 19: I suggest adding 'Lato Light' at first so it's quite the same as before on systems where the font exist.
```suggestion
                font-family: "Lato Light", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
```

asbiin (changes requested):
Good work, thank you @mgrhm

resources/views/errors/layout.blade.php
mgrhm: The colour #B0BEC5 is only used in that one instance – so I replaced it with its nearest neighbour as part of the standardisation process.

asbiin:
Thank you @mgrhm

PR #5093, opened 2021-04-23
refactor: remove reminder_sent table

asbiin:
We don't really use it


database/migrations/2021_04_23_190837_remove_reminder_sent.php
```diff
@@ -3,7 +3,7 @@
 use Illuminate\Support\Facades\Schema;
 use Illuminate\Database\Migrations\Migration;
 
-class DeleteReminderSentTable extends Migration
```
asbiin, line 6: @djaiss ;)

djaiss: Awesome

PR #5102, opened 2021-04-24
feat: order introductions contact list by first and last name

iloveitaly:
It's feels much more natural to me to order the introductions list by first and last name.

#### Before submitting the PR

### General checks
- [x] The name of the PR should follow the [conventional commits guideline](https://github.com/monicahq/monica/blob/master/docs/contribute/readme.md#conventional-commits) that the project follows.

#### Other tasks
- [x] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG.md) entry added, if necessary, under `UNRELEASED`.
- [x] If it's relevant and worth mentioning, create a changelog entry for this change. The changelog entry will appear inside the UI for all users to see. To know if your change is worth the creation of a changelog entry, [read the documentation](https://github.com/monicahq/monica/blob/master/docs/administrators/tips.md#when-is-it-relevant-to-create-a-changelog-entry).
- [x] Don't forget to [ask for a free account](mailto:regis@monicahq.com) on https://monicahq.com as anyone who contributes can request a free account.


djaiss:
Hi @iloveitaly and thanks for the PR.

To you it might feel more natural, but in other cultures, it might not.

This is why we introduced the column called `name_order` in the `users` table that define how people want their contact to be sorted.

I think we should use this in this case, what do you think @asbiin ?

iloveitaly:
Ah, great, didn't realize there was an option for that. It sounds like we should add a [switch similar to this one](https://github.com/monicahq/monica/blob/708eb6090ebe8ef71546e99621718c5786712dee/app/Helpers/FormHelper.php#L16) to add 2-3 `orderBy`s to the query. I'm not that familiar with the codebase—where would be the best place for this helper?

djaiss:
> Ah, great, didn't realize there was an option for that. It sounds like we should add a [switch similar to this one](https://github.com/monicahq/monica/blob/708eb6090ebe8ef71546e99621718c5786712dee/app/Helpers/FormHelper.php#L16) to add 2-3 `orderBy`s to the query. I'm not that familiar with the codebase—where would be the best place for this helper?

Good question.
We might need to use this in other places too. So I would perhaps create a `ContactHelper` with the right method name in it, like `getNameInUserOrder` or something like this (yes, naming is hard).

app/Helpers/ContactHelper.php
```diff
@@ -0,0 +1,55 @@
...
+use App\Models\User\User;
+
+class ContactHelper
+{
+    /**
+     * Get the name order that will be used when rendered the Add/Edit forms
+     * about contacts.
+     *
+     * @param Builder $query
+     * @param User $user
```
asbiin, line 14:
```suggestion
     * @param User $user
     * @param \Illuminate\Database\Eloquent\Builder $query
```

app/Helpers/ContactHelper.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace App\Helpers;
+
+use App\Models\User\User;
```
asbiin, line 5:
```suggestion
use App\Models\User\User;
use Illuminate\Database\Eloquent\Builder;
```

app/Helpers/ContactHelper.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    /**
+     * Get the name order that will be used when rendered the Add/Edit forms
+     * about contacts.
+     *
+     * @param Builder $query
+     * @param User $user
+     * @return string
+     */
+    public static function orderContactQueryByUserPreference(User $user, $query)
```
asbiin, line 17:
```suggestion
     * @return Builder
     */
    public static function orderContactQueryByUserPreference(User $user, $query): Builder
```

iloveitaly: I couldn't get the builder return type to work for me. Looks like it's failing the build. Let me know if you want to me to make any other changes here to get this merged. Feel free to modify this branch as much as you'd like!

iloveitaly:
@asbiin Any thoughts on how to fix the typing error? `Builder` as a param + return type seems to return an error since `HasMany` is passed in.

app/Helpers/ContactHelper.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * Get the name order that will be used when rendered the Add/Edit forms
+     * about contacts.
+     *
+     * @param User $user
+     * @param Builder $query
+     * @return Builder
+     */
+    public static function orderContactQueryByUserPreference(User $user, $query): Builder
```
asbiin, line 18: In fact if you want to use Builder here, it's better to use a [scope](https://laravel.com/docs/8.x/eloquent#local-scopes)
See https://github.com/monicahq/monica/blob/master/app/Traits/Searchable.php

asbiin (changes requested):
phpstan is complaining, and I think it would be better to use a scope

PR #5136, opened 2021-04-30
feat: Datestamp filename of exported SQL file.

mgrhm:


app/Http/Controllers/SettingsController.php
```diff
@@ -194,8 +194,10 @@ public function exportToSql()
 
         $adapter = disk_adapter(ExportAccountAsSQL::STORAGE);
 
+        $exportdate = date('Y-m-d');
```
asbiin, line 197: Like phpstan report mentioned, you need to add `use function Safe\date;` at the beginning of the file

asbiin (changes requested):
Good idea!

app/Http/Controllers/SettingsController.php
mgrhm: I've given it a go in the newest commit.  I don't actually know any PHP, so this was done by combining Googled results with trial and error.  It tested successfully on my Monica instance without `Safe\date;`, but I've added it.  Hope this works!

asbiin: 'Safe' is a package we use - see https://github.com/thecodingmachine/safe

asbiin (approved):
Good job!
You just have to fix `Tests\Feature\SettingsTest::test_user_can_export_account` test - and merge/rebase with master to fix a flaky test I've fixed.

mgrhm:
> You just have to fix `Tests\Feature\SettingsTest::test_user_can_export_account` test - and merge/rebase with master to fix a flaky test I've fixed.

I've given it a go.  I don't know any PHP, but I googled to find the PHP regex and I _think_ this should work.

If it works as intended, it should match any year between 2000 and 2099, and it should not pass any months or days that couldn't exist naturally.

mgrhm:
Okay, that hasn't worked.  I don't know what else to try.  Not knowing PHP, it's hard to know if a regex is even the right thing to use here.  Sorry!

app/Http/Controllers/SettingsController.php
```diff
@@ -203,8 +204,10 @@ public function exportToSql()
...
```
asbiin, line 207: It will be better like this:
```suggestion
        $exportdate = Carbon::now(DateHelper::getTimezone())->format('Y-m-d');
```

app/Http/Controllers/SettingsController.php
```diff
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers;
 
+use function Safe\date;
```
asbiin, line 5: See suggestion below
```suggestion
use Illuminate\Support\Carbon;
```

tests/Feature/SettingsTest.php
```diff
@@ -51,7 +51,7 @@ public function test_user_can_export_account()
         $response = $this->get(route('settings.sql'));
 
         $response->assertStatus(200);
-        $this->assertTrue($response->headers->get('content-disposition') == 'attachment; filename=monica.sql');
+        $this->assertTrue($response->headers->get('content-disposition') == 'attachment; filename=monica-export.20[0-9][0-9]-[0-1][0-9]-[0-3][0-9].sql');
```
asbiin, line 54:
```suggestion
        $this->assertTrue($response->headers->get('content-disposition') == 'attachment; filename=monica-export.2021-11-01.sql');
```

tests/Feature/SettingsTest.php
```diff
@@ -51,7 +51,7 @@ public function test_user_can_export_account()
         $response = $this->get(route('settings.sql'));
```
asbiin, line 51: With Carbon you can force the date for the tests:
```suggestion
        Carbon::setTestNow(Carbon::create(2021, 11, 1, 7, 0, 0));

        $response = $this->get(route('settings.sql'));
```

asbiin (changes requested):
Using Carbon to define the date is easier with the test context.

tests/Feature/SettingsTest.php
mgrhm: That all went over my head, but you know what you're doing and I don't!  Thanks very much for fixing this.

asbiin:
Many thanks @mgrhm !

PR #5146, opened 2021-05-02
docs: Revised minimum hardware requirements due to build process

mgrhm:
Prior versions worked because yarn assets were pre-built and included.  With the removal of these, I don't think Monica can be successfully updated on a machine with 768MB or less of RAM.  I *think* yarn's memory requirements are too high.

Monica had been working fine on a Debian VPS with 256MB RAM.  It was a little slow but fully functional.  While trying to upgrade to 3.0, I could not successfully run the `yarn run production` command.  Despite being left for 6+ hours on several attempts, it simply freezes and does not go any further during the "fetching packages" stage.

I tried to build 3.0 on another Debian VPS with 768MB of RAM.  It got further in the process but repeatedly failed with this error:

    error Command failed with exit code 134

Which, as far as I can tell from Google, is because it runs out of memory.

Given that, I'm no longer comfortable recommending Monica run on such a constrained machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ If you want to host Monica yourself, you will need a server with:
 - Composer
 - MySQL
 
-Monica has been successfully self-hosted on a VPS with a single CPU core and 256&thinsp;MB of RAM. Monica was fully functional on this server, although suffered occasional slowdowns due to constrained resources. A machine with more RAM and CPU cores will give faster performance.
+To successfully build and host Monica, we recommend a system with at least 1.5&thinsp;GB for RAM.  Monica can run on systems with significantly less memory, but due to the high memory requirements of the build process during updates, you may encounter issues and failed builds.
 
 ### Update your instance
 
--- a/README.md
+++ b/README.md
@@ -134,6 +134,8 @@ If you want to host Monica yourself, you will need a server with:
 - Composer
 - MySQL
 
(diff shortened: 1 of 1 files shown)


asbiin:
Thank you @mgrhm 
We have to distinguish the hardware requirement to run Monica, and the hardware requirements to build it.
You can comfortably either download the build assets from the releases files, or build it on your own computer, and then push the files in the web server.
What do you think?

mgrhm:
> We have to distinguish the hardware requirement to run Monica, and the
hardware requirements to build it.

I hadn't considered those as two separate sets of requirements, but now I realise that _building_ Monica is far more resource intensive than _running_ Monica.

I think this leaves us with two options with regards to documentation:

1. We determine a sensible minimum spec for building Monica and declare that the recommended minimum for hosting your own instance
2. We delineate two sets of specs, hosting and building, and explain the differences in the documentation

Personally, I prefer option 1 because it's simpler, and seems less likely to lead people into trouble.

What do you think?

asbiin:
@mgrhm I'm OK with your proposition 1 yes!

mgrhm:
Brilliant.  Do you have any idea of what the minimum recommended spec should be?  I've tried searching for minimum memory for `yarn`, but haven't found anything definitive.

From my own testing, anything less than 1GB probably isn't enough.  I've successfully run `yarn run production` on two machines -- but neither has less than 4GB of RAM, which is almost certainly excessive and sets too high a barrier of entry to Monica.

We could be a bit cautious and chose 1.5GB to give a margin of error?  That's surely enough?

mgrhm:
I've rewritten the appropriate section of the README.  I've tried to make it accurate and to guide users without being overly prescriptive.  Please let me know what you think.

asbiin:
Thank you @mgrhm !

Issue #5083, opened 2021-04-21
Quick add contact from name + email string


iloveitaly:
A common workflow I have is copying an email address from gmail into monica that looks like:

```
Mike Bianco <mike@mikebian.co>
```

This currently takes a lot of steps:

1. Copy/paste first & last name into separate fields
2. Create the contact
3. Add new contact information
4. Select email
5. Paste the email address
6. Press "done"

It would be great if this was a two-click operation:

1. Click "Quick add new contact"
2. Paste `Mike Bianco <mike@mikebian.co>`
3. Click "add" (monica parses first, last, and email from the string)

I'm happy to work on a PR for this if you could point me in the right direction in the codebase. Thanks!

cstrat:
This would be great.

Monica already does this with first/last name if you enter it into the search box and click add user.
Adding email with that format shouldn't be too hard - although the form doesn't have an email field so it would be a bit more of a change than just parsing the string.

PR #5182, opened 2021-05-12
feat: quick add with email

iloveitaly:
https://github.com/monicahq/monica/issues/5083

Not sure if it's worth writing a test for this case, feel free to point me to an existing one for the quick add parsing and I can update it.

#### Before submitting the PR

### General checks
- [x] The name of the PR should follow the [conventional commits guideline](https://github.com/monicahq/monica/blob/master/docs/contribute/readme.md#conventional-commits) that the project follows.

### Front-end changes
- [x] If you change the UI, make sure to ask repositories administrators first about your changes by pinging djaiss or asbiin in this PR.
- [x] Front-end tests have been written with Cypress.


iloveitaly:
@asbiin this one is ready to go! Want to take a look?

asbiin:
@iloveitaly I think that's a good idea, however we need to be sure it does downgrade search performances. Is there a index on email on the contacts table?

iloveitaly:
@asbiin Doesn't look like it

image

Any issue with adding an index to this table?

asbiin:
@iloveitaly no, we should add it, that's the idea.

asbiin:
@iloveitaly don't hesitate if you need help about adding the index

asbiin:
Thank you @iloveitaly

iloveitaly:
@asbiin no problem!! So glad this got merged. One improvement that would be great here is to detect an email-only search entry and use that value for the email field instead of the first name field.

PR #5183, opened 2021-05-13
Fixing local docker development

iloveitaly:
The individual copy commands missed some key files that are required to build the image. I don't see any downside to copying the entire source over.

### DIFF
--- a/docs/contribute/docker.md
+++ b/docs/contribute/docker.md
@@ -21,21 +21,28 @@ docker-compose up
 Use this process if you're a developer and want complete control over
 your Monica container.
 
-Edit `.env` again to set the `DB_*` variables to match your database. Then run:
+If you aren't using docker-compose, edit `.env` again to set the `DB_*` variables to match your database. Then run:
 
 ```sh
-docker build -t monicahq/monicahq .
+# assets are copied from the host machine, make sure they are built
+npm install --production && npm run preproduction && npm run production
+
+docker build -t monicahq/monicahq -f scripts/docker/Dockerfile .
+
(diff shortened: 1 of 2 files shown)


docs/contribute/docker.md
```diff
@@ -21,21 +21,28 @@ docker-compose up
 Use this process if you're a developer and want complete control over
 your Monica container.
 
-Edit `.env` again to set the `DB_*` variables to match your database. Then run:
+If you aren't using docker-compose, edit `.env` again to set the `DB_*` variables to match your database. Then run:
 
 ```sh
-docker build -t monicahq/monicahq .
+# assets are copied from the host machine, make sure they are built
+npm install --production && npm run preproduction && npm run production
```
asbiin, line 28: The commands to run are:
```suggestion
yarn install && yarn run production
```

iloveitaly: I don't think this works: without `preproduction` the build won't complete, unless yarn magically runs that command.

iloveitaly:
> Another solution would be to build the assets inside the Dockerfile. It will increase the time to build the image, but might be more interesting...

@asbiin That would be neat! The image used in this docker file doesn't have npm installed and I don't have the time to dedicate to tinkering with this further since it's at least working for me now.

docs/contribute/docker.md
asbiin: Exactly, that's the purpose of this commands: the "pre{pattern}" commands are run before the "{pattern}" command itself
See https://docs.npmjs.com/cli/v7/using-npm/scripts#pre--post-scripts

iloveitaly: Ah, neat, wasn't aware of that npm feature!

PR #5192, opened 2021-05-15
feat: download and get storage files as private

### LINKED ISSUE #987: Profile Pictures Are Public (accesssable to anyone without login) In Docker Version
I've migrated my setup from a manual install over to the docker version. The docker version is great except that any file in "www.myserver.com/storage/avatars/" can be accessed by anyone if they know the image file name. In my previous manual install I had a setup in my virtual host in Apache that blocked access to these files. It was as follows:
```
<Directory /var/www/monica>
   Options FollowSymLinks
   Options -Indexes
   AllowOverride All
   Require all denied
 </Directory>
 <Directory /var/www/monica/public>
   Require all granted
 </Directory>
```
However, to my knowledge it isn't possible to do this with docker because you can't use \<Directory> for a reverse proxy. I tried to use \<Location> and \<LocationMatch> but I was unable to make those work. 

Is having the images public intentional or is this a bug with the docker version of Monica? 

This is my current reverse proxy virtual host for the Docker version: 
```
<IfModule mod_ssl.c>
<VirtualHost *:443>
    ServerAdmin webmaster@myserver.com
    ServerName monica.myserver.com
    ErrorLog ${APACHE_LOG_DIR}/monica-error.log
    ProxyPreserveHost On
    ProxyRequests Off
    ProxyPass / http://localhost:8080/
    ProxyPassReverse / http://localhost:8080/
SSLCertificateFile /etc/letsencrypt/live/monica.myserver.com/fullchain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/monica.myserver.com/privkey.pem
Include /etc/letsencrypt/options-ssl-apache.conf
</VirtualHost>
</IfModule>
```

asbiin:
This close #987

It allows to store photos, avatars, and document privately in s3, and limit access to the file to the only user who owns it.
In addition, we can leverage the browser cache, which will increase speed view of Monica.

You can opt-out to the feature by setting `FILESYSTEM_DEFAULT_VISIBILITY` to 'public' - it will be 'private' by default.
The cache behavior can be change by defining `DEFAULT_CACHE_CONTROL`.


app/Helpers/StorageHelper.php
```diff
@@ -57,4 +57,14 @@ public static function hasReachedAccountStorageLimit(Account $account): bool
...
     }
+
+    /**
+     * Get visibility to put files in storage.
+     *
+     * @return string
+     */
+    public static function visibility(): string
+    {
+        return config('filesystems.secure_files') ? 'private' : 'public';
```
djaiss, line 68: Why don't you simply return the config variable, and define a sensible default for the variable on the config file instead of making this comparison here?

routes/web.php
```diff
@@ -46,6 +46,8 @@
         Route::post('/dashboard/setTab', 'DashboardController@setTab');
     });
 
+    Route::get('/store/{file}', 'StorageController@download')->where('file', '.*')->name('download');
```
djaiss, line 49: It's weird to have a `GET` route that is actually called `store` and the method in the controller is called `download`.

djaiss (commented):
A monster of a PR! Awesome job. A few comments.

routes/web.php
asbiin: `storage` is not usable. What would you use? `files`?
The function can be called `get` indeed.

app/Helpers/StorageHelper.php
asbiin: you mean define a config to be `private` or `public` (other values would probably not work ...)?
That's possible

PR #5209, opened 2021-05-19
style: Gender neutral defaults

mgrhm:
These changes are all non-breaking as far as I can tell.  I've added a `_male` equivalent for each one where it was appropriate, then changed the default to be gender neutral, with the exception of 'uncle', which I couldn't cleanly make neutral.  The best I could think of is 'relative', which is too broad, or 'parent's sibling', which is clumsy.

Where there was any doubt, I've checked spellings and usage in the *[Shorter Oxford English Dictionary](https://en.wikipedia.org/wiki/Shorter_Oxford_English_Dictionary)*.  I've noted any of the more obvious examples in my commit messages.

I don't know enough PHP to make the changes required to use the new `_male` alternates yet, so will have to do a lot of reading before I'm ready to do that.  In the meantime, this should render Monica gender-neutral by default.


jacobm001:
Most of the changes look good to me, but I would note that _boyfriend/girlfriend_ are not really equivalent to _partner_. Generally speaking, I interpret the latter as being much more serious than the former. 

In my experience, the connotation here is that _boyfriend/girlfriend_ is used relatively early on in a relationship. _Partner_ comes once the relationship has stabilized and has become a long term relationship (from the couple's perspective). Also worth noting is the fact that _husband/wife_ has religious connotations for many, so non-hetero couples may not like the association causing them to use the term _partner_ over spouse. 

This may be coming less of a concern now that gay marriage is legal in the US and such connotations are going away, but I think it's still relevant in how people around me describe themselves.

mgrhm:
> Generally speaking, I interpret [partner] as being much more serious than [boyfriend/girlfriend].

Agreed – ‘partner’ would generally lead me to believe the relationship had been established for some time, and/or the people in it were older.  I chose it as a gender-neutral compromise because I couldn’t find a direct equivalent of boy/girlfriend.

What would you suggest as a replacement for ‘partner’?  I’d personally prefer ‘partner’ as a default over ‘boyfriend’, even thought it’s maybe more serious.

> Also worth noting is the fact that husband/wife has religious connotations for many, so non-hetero couples may not like the association causing them to use the term partner over spouse.

Which they’re free to do.  I don’t see how making ‘spouse’ default for a married relationship impacts people who choose not to be married, or to reject traditional husband/wife labels.  They can choose whichever title/label for their relationship that best captures it.

Conversely, many non-heterosexual couples, in the Anglophone and beyond, actively and enthusiastically choose to use husband/wife – many having fought for a long time for marriage equality.

In both cases, I’d argue that making the default gender neutral and letting people choose for themselves, is a better compromise.  Trading a little accuracy of language in some cases is a better compromise than _not_ making it gender neutral and a lesbian finding her married partner defaulting to ‘husband’.

mgrhm:
I’d also be interested if you have a gender-neutral suggestion for ‘uncle’.  I’m still unhappy with that, and I still don’t have a better one.

jacobm001:
> Which they’re free to do. I don’t see how making ‘spouse’ default for a married relationship impacts people who choose not to be married, or to reject traditional husband/wife labels. They can choose whichever title/label for their relationship that best captures it.

My only thought in mentioning it here was that I wasn't sure if using _partner_ as the default for boyfriend/girlfriend prevented the same label from being used by a couple in a different status. I do know a few who are legally married (though not religiously) and use the term _partner_. I think _significant other_ is a better, and more standard (at least in my dialects) gender neutral form here.

There is still no standard gender neutral for aunt/uncle nor nephew/niece as far as I'm aware. 

[Nibling](https://www.merriam-webster.com/words-at-play/words-were-watching-nibling) as a gender neutral of the latter seems to be approaching mainstream, but that may be very regional. I'm from a fairly liberal area (Willamette Valley, Oregon, US) and I've still never heard someone use it in conversation.

Aunt/Uncle is much harder. I heard someone try to use _parsib_ as a portmantua of _parents sibling_, but it felt really forced.

asbiin:
@mgrhm @jacobm001 what about `ex-lover` for the gender neutral version?

jacobm001:
@asbiin gender neutral as what?

Personally, I'd never use the term lover/ex-lover as they have some pretty negative connotations here, but it's certainly a valid term people wouldn't have trouble understanding.

asbiin:
> @asbiin gender neutral as what?

The default version is considered gender neutral by this PullRequest.
As you disliked `boyfriend/girlfriend`, and `partner`, I just made another proposition. I understand none is perfect, but we have to make a choice. Or maybe you have another solution in mind?

mgrhm:
Hi @asbiin, hi @jacobm001,

I think we’re in very serious danger of [bike-shedding here](https://en.wikipedia.org/wiki/Law_of_triviality) and getting nowhere, which would be worse than making some decisions and accepting that this will never be perfect.

tl;dr: “partner” is the best/least worst gender-neutral, non-specific title for a variety of romantic partnerships.  It’s broad and sometimes non-specific, but captures within it most combinations.  We’ll just have to live with the others for now.

> My only thought in mentioning it here was that I wasn't sure if using partner as the default for boyfriend/girlfriend prevented the same label from being used by a couple in a different status.

I don’t think it does: these don’t need to be used once-and-only-once.  The aim here would be to set a gender neutral default and then let people choose.  They’ll be selecting from the drop-down list, so they’ll see all the options and can choose the best one for them.

> I do know a few who are legally married (though not religiously) and use the term partner. I think significant other is a better, and more standard (at least in my dialects) gender neutral form here.

“Significant other” is common in American English but far less so in other varieties.  It’s only really present in them because of American media, but I’ve never heard it used naturally by speaker of another variety of English.

Arguably, Monica’s default is U.S. English, so we _could_ go with that, but most English speakers will leave it on the default so I’d argue we should choose non-regional words wherever possible.  Monica’s true default English should be treated as “Standard Worldwide English”.

> Nibling as a gender neutral of the latter seems to be approaching mainstream, but that may be very regional. I'm from a fairly liberal area (Willamette Valley, Oregon, US) and I've still never heard someone use it in conversation.

If you haven’t heard it used, I’d say it’s not approaching mainstream.  Words are approaching mainstream when regular people use them naturally; anything else is astroturfed, which is what I’d say Merriam-Webster is doing here.  This, like “Uncle” is one where we should just live without a gender-neutral option for now.

> Aunt/Uncle is much harder. I heard someone try to use parsib as a portmantua of parents sibling, but it felt really forced.

What a horrible, forced neologism. I don’t think we have anything except to leave it as-is for now and accept that English doesn’t have a gender-neutral word for “parent’s sibling”.

> I'd never use the term lover/ex-lover as they have some pretty negative connotations here[.]

I’d have to second this: “lover” has implications of an illicit relationship in English which it probably doesn’t have in other languages.  Let’s stick with “partner” and “ex-partner” where there’s no clear option, and also as the generic defaults.

jacobm001:
@mgrhm My apologies, it wasn't my intent to derail the PR, just trying to offer another perspective since it looked inactive. I have no strong objections here, except the lover thing.

resources/lang/en/app.php
```diff
@@ -168,45 +170,61 @@
...
+    'relationship_type_child_male' => 'son',
+    'relationship_type_child_with_name' => ':name’s child',
     'relationship_type_child_female_with_name' => ':name’s daughter',
+    'relationship_type_child_male_with_name' => ':name’s son',
 
-    'relationship_type_stepparent' => 'stepfather',
+    'relationship_type_stepparent' => 'step-parent',
     'relationship_type_stepparent_female' => 'stepmother',
-    'relationship_type_stepparent_with_name' => ':name’s stepfather',
+    'relationship_type_stepparent_father' => 'stepfather',
```
asbiin, line 196:
```suggestion
    'relationship_type_stepparent_male' => 'stepfather',
```

resources/lang/en/app.php
```diff
@@ -168,45 +170,61 @@
...
+    'relationship_type_stepparent_with_name' => ':name’s step-parent',
     'relationship_type_stepparent_female_with_name' => ':name’s stepmother',
+    'relationship_type_stepparent_male_with_name' => ':name’s stepfather',
 
-    'relationship_type_stepchild' => 'stepson',
+    'relationship_type_stepchild' => 'stepchild',
     'relationship_type_stepchild_female' => 'stepdaughter',
-    'relationship_type_stepchild_with_name' => ':name’s stepson',
+    'relationship_type_stepchild_female' => 'stepson',
```
asbiin, line 203:
```suggestion
    'relationship_type_stepchild_male' => 'stepson',
```

asbiin (changes requested):

asbiin (approved):

asbiin:
Thank you @mgrhm and @jacobm001 !

mgrhm:
@jacobm001: No need to apologise. These are hardest decisions to make because there’s such a degree of subjectivity.

PR #5232, opened 2021-05-24
fix: Show fields in correct order on "Add Contact" page

mgrhm:
The `if`-`else` statement here didn't appear to work correctly -- it always displayed the last-name field before the first-name field.  This fixes is so that it displays first-name ahead of last-name by default, and uses the `if`-`else` to flip it for languages that use last/first.

### DIFF
--- a/resources/views/people/create.blade.php
+++ b/resources/views/people/create.blade.php
@@ -30,7 +30,17 @@
 
       <div class="pa4-ns ph3 pv2 bb b--gray-monica">
         {{-- This check is for the cultures that are used to say the last name first --}}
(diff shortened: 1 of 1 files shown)


asbiin (changes requested):
Thank you @mgrhm , but I don't think this is the real problem.
`formNameOrder` is get from FormHelper::getNameOrderForForms which might miss some cases. I will fix it with another PR and add some checks.

mgrhm:
Thanks, Alexis. I'll close this now.

asbiin:
Superseded by #5255

PR #5271, opened 2021-06-14
docs: update debian instructions to support deb10

tresni:
1. curl is not installed by default on Debian 10 (not sure about previous version), so use `wget` instead
2. PHP 7.4 is not available on Debian <=10 (only available in testing/unstable), instead use deb.sury.org packages (from the maintainer of the Debian packages.)
3. Composer v1 is the only available version on Debian <=10, install from getcomposer.org to get v2

Doing this allowed me to get Monica running locally on a clean Debian 10 netinst.  The hardest part was getting php 7.4 installed, apt kept wanting to install php8 and I didn't know if that was supported.

### DIFF
--- a/docs/installation/providers/debian.md
+++ b/docs/installation/providers/debian.md
@@ -42,23 +42,32 @@ sudo apt install -y git
 
 **PHP:**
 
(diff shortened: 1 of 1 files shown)


JavaZauber:
Excuse me why is there no debian package like there's for e. g. MediaWiki or GitLab?
Both Ubuntu and Debian would benefit from a single package which sets the the software up

PR #5285, opened 2021-06-19
feat: make archived contact readonly

asbiin:


mgrhm:
This is a great idea!

PR #5304, opened 2021-06-23
style: Stop Activity tailoring links wrapping over line breaks

mgrhm:


mgrhm:
I’m unsure how to fix this error. The log says:

    Error: Assets are not up to date. Please rebuild with: 'yarn run lint:all' and 'yarn run prod'.
    Error: Process completed with exit code 255.

I’ve run both of those commands and committed the updated manifest, but it’s still failing.

Any ideas?

asbiin (changes requested):
These files must not be added in the sources anymore
- public/css/app-ltr.css
- public/css/app-rtl.css
- public/css/stripe.css
- public/mix-manifest.json

mgrhm:
Reverted these changes, so this should be okay now.

mgrhm (commented):
Compiled assets removed.

asbiin:
Thank you @mgrhm !

PR #5103, opened 2021-04-24
feat: search notes when searching through contacts

iloveitaly:
This adds support for searching through notes [as explained here](https://github.com/monicahq/monica/issues/5079). There are already tests for the contact search helper, but let me know if you feel like this functionality needs a dedicated test.

#### Before submitting the PR
- [x] Read the [CONTRIBUTING document](https://github.com/monicahq/monica/blob/master/CONTRIBUTING.md) before submitting your PR.
- [x] If the PR is related to an issue or fix one, don't forget to indicate it.
- [x] Create your PR as draft if it is not final yet. Mark it as ready... when it’s ready. Otherwise the PR will be considered complete and rejected if it's not working.

### General checks
- [x] Make sure that the change you propose is the smallest possible.
- [x] The name of the PR should follow the [conventional commits guideline](https://github.com/monicahq/monica/blob/master/docs/contribute/readme.md#conventional-commits) that the project follows.

#### Backend/models changes
- [x] The API has been updated.
- [x] API's documentation has been added by submitting a pull request in the [marketing website repository](https://github.com/monicahq/marketing_site/pulls).
- [x] Tests have been added for the new code.
- [x] If you change a model, make sure the SetupTest file is updated. We need seeders to develop locally and generate fake data.

#### Other tasks
- [x] [CHANGELOG](https://github.com/monicahq/monica/blob/master/CHANGELOG.md) entry added, if necessary, under `UNRELEASED`.
- [x] [CONTRIBUTORS](https://github.com/monicahq/monica/blob/master/CONTRIBUTORS) entry added, if necessary.
- [x] If it's relevant and worth mentioning, create a changelog entry for this change. The changelog entry will appear inside the UI for all users to see. To know if your change is worth the creation of a changelog entry, [read the documentation](https://github.com/monicahq/monica/blob/master/docs/administrators/tips.md#when-is-it-relevant-to-create-a-changelog-entry).
- [x] Don't forget to [ask for a free account](mailto:regis@monicahq.com) on https://monicahq.com as anyone who contributes can request a free account.

PR #5306, opened 2021-06-24
fix: fix search being extremely slow

djaiss:
This is a temporary fix for #5103 which caused a lot of performance issues in production, considering the size of the database.


asbiin (approved):
Approved, we have to look deeper to have a performing search

PR #5343, opened 2021-07-06
feat: add Vietnamese language

nhymxu:
Translate 95%. Some emotion string will translate soon
image

### DIFF
--- a/config/lang-detector.php
+++ b/config/lang-detector.php
@@ -41,6 +41,7 @@ return [
             'nl',
             'sv',
             'tr',
+            'vi',
             'zh',
             'zh-TW' => 'zh-TW',
         ])
--- a/resources/lang/en/settings.php
+++ b/resources/lang/en/settings.php
@@ -94,6 +94,7 @@ return [
     'locale_pt-BR' => 'Portuguese (Brazil)',
     'locale_ru' => 'Russian',
     'locale_sv' => 'Swedish',
+    'locale_vi' => 'Vietnamese',
     'locale_zh' => 'Chinese Simplified',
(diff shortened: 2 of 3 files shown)


asbiin:
Thank you @nhymxu I will review the translation ASAP as well.
I already see 1 glitch: I think in Vietnamese there is no plural form (all ":count contact|:count contacts" source text). In that case you don't have to add a plural form in the translated text, ex: ":count contact|:count contacts" => ":count liên hệ".
It's based on a laravel code, and you can also see [here](https://github.com/monicahq/monica/blob/139d53cef7e31a51b2c2f39999492fa9a2ed63e4/resources/js/pluralization.js#L50) it's included in Monica.

nhymxu:
@asbiin 
Thank you, I will update this on crowdin when see this source text

asbiin:
@nhymxu actually I have validated all other translations, you can see all not approved text in crowdin: https://crowdin.com/translate/monicahq/12/en-vi?filter=basic&value=5 (select others files too)

nhymxu:
@asbiin 
I updated some string. Except `every day|every :number days` will translate difference for singular `hàng ngày` and plural `mỗi :number ngày`.
Rule same with `month` and `year`

asbiin:
@nhymxu That's amasing, thank you! For `every day|every :number days` and alike, you'll still have to define a unique form, because only the 1st item will be rendered very time. For instance, this text will always be translated as `hàng ngày`, no matter the number of days ... which is not what we want. I agree it's not exactly the same text, well that's the problem with translations: some times we have to deal with it.

Issue #5412, opened 2021-08-06
The confirm contact deletion message template is slightly broken and the name doesn't fill in correctly.


cstrat:
**Describe the bug**
The confirm contact deletion message template is slightly broken and the name doesn't fill in correctly.
See screenshot below.

**Screenshots**
This person is not called 'name' 😄 
![image](https://user-images.githubusercontent.com/4209076/128456797-f7b67bc3-7b5a-4727-ad3b-c0eb68c271dc.png)

**Which version are you using:**
 - Self Hosted
 - v3.1.3

PR #5431, opened 2021-08-16
fix: Contact name population on delete confirmation

benfes:
Fix bug where the contacts name is not populating the delete confirmation

Resolves: https://github.com/monicahq/monica/issues/5412


asbiin (approved):
Good one! Thank you @benfes

PR #5545, opened 2021-10-01
docs: typofixes

nexxai:
Just cleaning up some typos in the docs


asbiin:
Thank you @nexxai ! Can't believe we left these typos for so long :D

PR #5546, opened 2021-10-01
docs: update build badge in README

particleflux:
The build badge in the README file was still pointing to an outdated Azure build run from February.
Seems like builds have been migrated to GitHub Actions afterwards.

I used the `tests` workflow, since that seems to be the correct one to replace a *whole* build flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 <div align="center">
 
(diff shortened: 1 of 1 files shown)


asbiin:
Thank you @particleflux

PR #5550, opened 2021-10-02
style: Added theme-color meta tag

mgrhm:
This makes changes the status bar and over scroll areas to Monica’s dark blue on recent iOS and Android devices.  There’s a good [write-up of it here](https://stuffandnonsense.co.uk/blog/time-to-update-your-theme-color-meta-tag-for-safari).

I’ve also grouped the `meta` tags together, to be more consistent with the other tags in the `head` area which were also grouped by type.


asbiin:
Thank you @mgrhm

PR #5569, opened 2021-10-05
fix: fix duplicate reminders on dashboard

### LINKED ISSUE #5381: Repeated task in home screen
Version 3.1.3
I created a reminder, then in home screen, I see events are repeated.
Please fix, thanks.
image

jsawo:
This PR addresses the issue with multiple reminders being shown on dashboard that was reported in #5381.
The problem occured when more than one user was assigned to a single account. The dashboard was displaying all reminders assigned to an account, which means one for every user. 

The proposed change restricts the reminders shown to a currently logged in user.

this should close #5381


asbiin:
Thank you @jsawo !

PR #5602, opened 2021-10-11
refactor: cleaning up constructors

krzysztofrewak:
In general, constructors of Laravel Commands and Controllers should not be setting any values or services. Every (!) constructor is ran before some commands like `artisan route:list`, `artisan env`, `artisan queue:work` and others, so putting things into constructors just making the application slower.

I just moved unnecessary invocations to handlers or endpoint methods. It shouldn't change or break anything.


asbiin:
Thank you @krzysztofrewak

PR #5607, opened 2021-10-12
fix: fix creating default gender

jsawo:
This PR fixes a bug in "Add gender type" dialog.

Screenshot_20211012_224514

When adding a new gender type, the user is presented with an option to set the new gender type as default (*"Select the default gender for a new contact"*). 
This however did not work. The default was not set and user was required to change default using a separate step - *"Change default gender"*.


asbiin:
Nice catch! Thank you @jsawo

PR #5661, opened 2021-10-28
fix: fix edit an activity with a category

### LINKED ISSUE #5392: When editing an activity the category displayed is not the current category
When editing an activity, the category dropdown shows the first category on the dropdown, not the one I selected.

This caused me some confusion, as well as an accidental saving of a different category.

If it matters, I'm hosting my own instance. Version is 2.20.0

ghost:
This fix #5392


asbiin:
Thank you @ribounette

PR #5662, opened 2021-10-28
fix: fix import table layout

### LINKED ISSUE #574: View Error - Imports Table
After importing a record, the table changes from 2 column to three, and the JS filling out that table is apparently choosing "last" instead of "2nd" when filling out the information. Or each row type needs it's own builder to properly handle the differences in content.
screenshot 2017-09-03 09 44 22

ghost:
Fix #574


asbiin:
Thank you @ribounette

PR #5826, opened 2022-01-02
feat: add contact soft delete and prunable

### LINKED ISSUE #5815: Contact not deleted with CardDAV on monica
Hi, I have a problem with Carddav and monica. 

When I delete a contact on my phone and start the synchronization in Davx5, the contact is never deleted in monica but does not appear anymore on my phone. (I am on Android)

Is this problem known or not? 

Thanks for your help

asbiin:
This will fix #5815
It also fix a major break: photos and document file deletion while the model is deleted.

Softdelete on contacts aims at having a list of deleted contacts on carddav to let other systems synchronize the deletion on their own.


djaiss (approved):
All looks good to me mate!

Co-7:
Hi, 

Does this feature fix the fact that when I delete a contact on my phone and sync it recreates it on my phone and does not delete it in monica as if the sync was only from monica to the phone?

Thank you for your work

PR #5989, opened 2022-02-12
feat: 3513 add Pronouns

### LINKED ISSUE #3513: Add Preferred Pronouns
As gender identify becomes more fluid, it might be nice to be able to track preferred pronouns for people who choose something non-default.  It could be tracked as notes, but it might also be nice to have a dedicated field for this, perhaps using the current "gender" tracking for the defaults for a specific pronoun.

Kotarosz727:
@djaiss 
UI Change as below. New pronouns input field is also added to people edit, relationship crate page.
image

#### Before submitting the PR
- [x] Read the [CONTRIBUTING document](https://github.com/monicahq/monica/blob/main/CONTRIBUTING.md) before submitting your PR.

### General checks
- [x] The name of the PR should follow the [conventional commits guideline](https://github.com/monicahq/monica/blob/main/docs/contribute/readme.md#conventional-commits) that the project follows.

### Front-end changes
- [x] If you change the UI, make sure to ask repositories administrators first about your changes by pinging djaiss or asbiin in this PR.
- [x] Front-end tests have been written with Cypress.

#### Backend/models changes

- [x] Make sure your changes do not break importing data with `vCard` and `.csv` files.

resolve #3513


alice-i-cecile:
Ideally this would be prepopulated based on the gender selected above (with default pronouns for each gender configurable when defining them). I'm not sure how feasible that is though :)

puppycodes:
> Ideally this would be prepopulated based on the gender selected above (with default pronouns for each gender configurable when defining them). I'm not sure how feasible that is though :)

I believe the less assumptions made in this section the better. Gender and pronouns do not necessarily have a relationship to one another. The analogy I would draw here would be trying to infer the gender from someones first name, which would be a bad idea because of the degree of variation and the fact that they are not necessarily connected. I believe the most inclusive tactic is to simply allow these both of these fields (gender and pronouns) to be treated like basic strings, allowing for any entry. This would be a future proof and maximally inclusive and also happens to simplify the dependencies.

pmlandwehr:
Would it be possible to default family relations to "sibling", "parent", etc. when gender is unknown?

puppycodes:
> Would it be possible to default family relations to "sibling", "parent", "spouse", etc. when gender is unknown?

+1 good call

asbiin:

PR #6014, opened 2022-02-21
fix: only include real contacts in carddav sync

### LINKED ISSUE #5150: CardDAV includes contacts which aren't 'real' contacts
This should be a simple fix by filtering, I am surprised it hasn't been raised before.

My iPhone sync's with Monica using CardDAV.
I am seeing names for people I have added - but not created contact cards for.
Example would be the names of children or siblings and parents of friends and associates.

Those names have no data attached, so they shouldn't be fetched by CardDAV for the Sync.

codemonkeysoftware:
fixes #5150

Adds the real scope to the carddav sync query so that non-real contacts are not included.


aarnold89:
This seems to have become a problem again. I'm getting all the spouses and kids back in my carddav sync.

PR #6039, opened 2022-03-03
5956 check male translation and fall back to generic

codemonkeysoftware:
Right now relationship type strings do not display a male version (eg. instead of "father" it displays parent). This is because the `RelationshipType`'s `getLocalizedName()` code gets the female translation for female relationship types, but the generic translation for male relationship types. To complicate matters, only 13/28 relationship types include a male entry in the translation file.

This change has `RelationshipType.getLocalizedName()` return the male translation for the male version and adds the missing male translations for English. Because I don't want this change to break if the other translations don't get updated with the new keys immediately, it also includes a fallback that recognizes Laravel's default return for missing keys (which is the requested key name) and instead returns the generic translations. This fallback can be removed as soon as the male translation entries get added.


app/Models/Relationship/RelationshipType.php
```diff
@@ -79,7 +79,10 @@ public function getLocalizedName(Contact $contact = null, bool $includeOpposite
         }
 
         $femaleVersion = trans('app.relationship_type_'.$this->name.'_female');
-        $maleVersion = trans('app.relationship_type_'.$this->name);
+        $maleVersion = trans('app.relationship_type_'.$this->name.'_male');
+        if ($maleVersion == 'app.relationship_type_'.$this->name.'_male') {
```
asbiin, line 83: It's better to use a `===` comparison here.
```suggestion
        if ($maleVersion === 'app.relationship_type_'.$this->name.'_male') {
```

app/Models/Relationship/RelationshipType.php
```diff
@@ -79,7 +79,10 @@ public function getLocalizedName(Contact $contact = null, bool $includeOpposite
...
+            $maleVersion = trans('app.relationship_type_'.$this->name);
+        }
 
         if (! is_null($contact)) {
             $maleVersionWithName = trans('app.relationship_type_'.$this->name.'_with_name', ['name' => $contact->name]);
```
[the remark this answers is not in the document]
codemonkeysoftware: You're right, I totally forgot about that.

Koellewe:
@asbiin @codemonkeysoftware What's the hold-up here? StyleCI has a small whitespace issue, but other than that this bugfix looks good to go

app/Models/Relationship/RelationshipType.php
```diff
@@ -79,10 +79,16 @@ public function getLocalizedName(Contact $contact = null, bool $includeOpposite
...
+        if ($maleVersion === 'app.relationship_type_'.$this->name.'_male') {
+            $maleVersion = trans('app.relationship_type_'.$this->name);
+        }
 
         if (! is_null($contact)) {
-            $maleVersionWithName = trans('app.relationship_type_'.$this->name.'_with_name', ['name' => $contact->name]);
+            $maleVersionWithName = trans('app.relationship_type_'.$this->name.'_male_with_name', ['name' => $contact->name]);
+            if ($maleVersionWithName === 'app.relationship_type_'.$this->name.'_male_with_name') {
+                $maleVersionWithName = trans('app.relationship_type_'.$this->name.'_with_name');
+            }
```
asbiin, line 91:
```suggestion
            }
```

asbiin (approved):

shaliru:
Sorry for stupid question, but how do I go about applying this fix to my installation? I'm on the latest version (v3.7.0) which was released in February - before this fix was was merged. Do I just drop the two changed files into my install directory?

Bodge-IT:
I'm 3.7.0 too, finding it difficult if I can't create proper relationships.  I can exec into the container but have no edit tools in there and no clue how to mod an image.

PR #6046, opened 2022-03-07
fix: life event creation with unknown month/day

### LINKED ISSUE #5936: Unable to add life event without specifying a day and month
**Describe the bug**
When adding life event without specified month and day of the event, the confirmation `Life event was added` pops ups, but the event is in fact not added.
No error shows up in either Monica or database logs.
No 'POST' event in Monica logs, which would normally show up upon successful creation.

**Screenshots**
This does not work:
image

This doesn't work either:
image

This works perfectly fine:
image

 - Self-hosted Docker Monica v3.6.1

particleflux:
This is a JS problem, current version sends this on create:

```
happened_at	"2022-0-0"
happened_at_day_unknown	false
happened_at_month_unknown	false
```

Whereas old version sent this:
```
happened_at	"2022-1-1"
happened_at_day_unknown	true
happened_at_month_unknown	true
```

This issue was introduced by 57b31f6526e3628cc98c8c271108bc34dd2d074e

The strict comparison added in the `updateDate` function in CreateLifeEvent.vue broke it.
The values for `this.selectedMonth` and `this.selectedDay` can be numbers/integers or strings, depending on where they come from.
Initial values and values set by JS are numbers, whereas values set by user input are strings.

Fixes #5936

I guess there might be a nicer approach to this in vue, or in general, though I don't have an idea :)

### DIFF
--- a/resources/js/components/people/lifeevent/CreateLifeEvent.vue
+++ b/resources/js/components/people/lifeevent/CreateLifeEvent.vue
@@ -261,10 +261,10 @@ export default {
      * Same for the month.
      */
     updateDate() {
-      if (this.selectedDay === 0) {
+      if (this.selectedDay === 0 || this.selectedDay === '0') {
         this.newLifeEvent.happened_at_day_unknown = true;
         this.newLifeEvent.happened_at = this.selectedYear + '-' + this.selectedMonth + '-01';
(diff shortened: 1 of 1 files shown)


asbiin:
Good catch, thank you @particleflux

Issue #6048, opened 2022-03-08
Error during installation - "node" incompatible 


SlidingHorn:
**Describe the bug**
Following the Debian installation instructions (using Bullseye), on the following step:

> Run `yarn install` to install frontend packages, then `yarn run production` to build the assets (js, css).

`sudo yarn install` gives the following error:

> Error @: The engine "node" is incompatible with this module. Expected version "16.x". Got "14.19.0"

This is after following the instructions word for word (with the sole exception being that I installed php7.4 directly from the Debian repositories, as it's available in Bullseye, thus a third party source is not required)

### Additional context

- **OS:** Debian 11 (fresh install on a virtual machine)
- **Server:** apache-2.4.52-1~deb11u2
- **PHP:** 7.4.28-1+deb11u1
- `yarn --version`: 1.22.17
- `node --version`: 14.19.0

binlabs:
@SlidingHorn Thanks for pointing this out. All of the Linux docs have now been updated to reflect that the minimum version of Node is 16.

binlabs:
This issue can be closed now that PR #6077 has been merged.

MaeiMaei:
This issue happens again due to the instructions not being bumped to 18, despite requiring Node 18.

PR #6077, opened 2022-03-27
docs: node 16 in debian and generic

binlabs:
A few weeks ago, @aaronpk updated the Ubuntu installation doc in PR #5980 to note that Node.js 16 is now the minimum required version, but forgot to update the Debian and Generic docs with the same change. This PR simply updates those two docs.

Issue #6048 was recently opened pointing out this issue on Debian, so this PR solves that issue.

### DIFF
--- a/docs/installation/providers/debian.md
+++ b/docs/installation/providers/debian.md
@@ -67,7 +67,7 @@ wget -q -O - https://getcomposer.org/installer | sudo php -- --install-dir=/usr/
 **Node.js:** Install node.js with package manager.
 
 ```sh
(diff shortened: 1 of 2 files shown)


djaiss:

Takingitstollonme (commented):
Who makes this call?

binlabs:
> Who makes this call?

I can't speak to this exact time, but it's generally due to dependencies: Either a new dependency added for functionality, or an existing dependency upgraded for security reasons, and that dependency only runs on a specific version of Node or higher.

PR #6087, opened 2022-04-06
fix: vcard bday export format with unknown year

### LINKED ISSUE #6079: vCard Export Format Birthdate wrong
For Contacts without a Year in their Birthday, the exported vCard contains a wrong formatted date

actually 
BDAY:--01-01

shoud be
BDAY:--0101

MarkusDick:
Fix #6079 

The vCard export is incorrect when the year is unknown.
It exported the birthday like:
`--mm-dd`

.. but it should be:
`--mmdd`

Referring to https://datatracker.ietf.org/doc/html/rfc6350/#section-6.2.5
Added test and fix.

#### Before submitting the PR
- [x] Read the [CONTRIBUTING document](https://github.com/monicahq/monica/blob/main/CONTRIBUTING.md) before submitting your PR.

### General checks
- [x] The name of the PR should follow the [conventional commits guideline](https://github.com/monicahq/monica/blob/main/docs/contribute/readme.md#conventional-commits) that the project follows.

#### Backend/models changes

### DIFF
--- a/app/Services/VCard/ExportVCard.php
+++ b/app/Services/VCard/ExportVCard.php
@@ -196,6 +196,8 @@ class ExportVCard extends BaseService
     /**
      * @param  Contact  $contact
      * @param  VCard  $vcard
+     *
+     * @see https://datatracker.ietf.org/doc/html/rfc6350#section-6.2.5
      */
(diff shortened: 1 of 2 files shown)


asbiin:
Thank you @MarkusDick

PR #6201, opened 2022-06-26
feat: add DB_TESTING_PORT in database config

aleksasaric:
fix: adding DB_TESTING_PORT in env.example and database config so testing connection can work on ports other than 3306


asbiin:
Thank you

PR #6236, opened 2022-07-20
fix: allow configuring port for test database

### LINKED ISSUE #6200: Connection for seeding testing db fails if DB PORT is not 3306
For developers forking and installing projects locally, if database connection port is not 3306, there is an Illuminate\Database\QueryException thrown on command php artisan migrate --database testing.

Screenshot 2022-06-26 at 21 04 11

particleflux:
The `testing` database used for the tests had no way to configure a db
server port. It always used some default fallback port (probably of the
pdo driver).

Added the `port` config setting with a new env variable `DB_TEST_PORT`.

Fixes #6200


djaiss:
Thanks for this!

PR #6250, opened 2022-07-31
feat!: switch to php 8.1+ dependency

asbiin:
The minimum php version for Monica is now php 8.1.

This is motivated for package dependency reasons, some package we'd like to upgrade to requires php 8.1.


djaiss (approved):
👍

PR #6252, opened 2022-08-01
fix: update php version in Dockerfile

credevator:
When I tried to run `docker-compose -f docker-compose.dev.yml up -d`  it failed with composer error.
It required PHP 8.1, I found that Dockerfile is using the PHP version 8 and composer is set with dependency on 8.1

I changed this in Dockerfile to make it work.


asbiin:
Thanks!

PR #6286, opened 2022-08-26
feat: add a name order with the last name outside of brackets

### LINKED ISSUE #6285: Nickname with bracketed first name
I really like that I can specify someone's "proper" name, but also what we call them in practice.  For many of my contacts, this is a diminutive, replacing fore-names but not the surname.  So I'd like a mechanism to indicate that.

For example, take "Alexandra Smith".  It can be important, and it's important to me, to capture her full name correctly.  But she's known in most circles as "Alex".  I'd like to be able to display her name as "Alex (Alexandra) Smith".

**Describe alternatives you've considered (optional)**

I'm currently using a bracketed full name, which gives "Alex (Alexandra Smith)", which isn't quite what I'm after -- if I refer to my friend by name, I'll probably call her "Alex Smith".

**Additional context**
I have a local branch which adds what I hope is a suitable implementation, which I'll create a PR for.

andrewaylett:
This fixes #6285.

Where nicknames are used as diminutives of full first names, it (hopefuly) makes sense to indicate as such in both the full name and short name fields.

### DIFF
--- a/app/Models/Contact/Contact.php
+++ b/app/Models/Contact/Contact.php
@@ -879,7 +879,7 @@ class Contact extends Model
                 break;
             case 'nickname_bracketed_firstname_lastname':
                 $completeName = $this->first_name;
-                
+
                 if (! is_null($this->middle_name)) {
                     $completeName = $completeName.' '.$this->middle_name;
                 }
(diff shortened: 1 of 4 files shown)


flyinprogrammer:
It would be great to have this feature ported to chandler, it was the first thing my significant other asked about 😂

asbiin (approved):

PR #6314, opened 2022-09-06
fix: Continuously pressing enter shows empty tags

### LINKED ISSUE #6235: Disallow to create empty tags
By going to the following page (the page of one person)

http://localhost:9090/people/h:9qyA7LJYWZ18mVavMj

When adding a tag (for example "a"), if you type enter multiple times, it creates as many empty tags as the number of keypresses.

Here is a screenshot of what it results in : 

Selection_002

This bugs seems to happen only after entering one tag that is not an empty string.

particleflux:
When on a contact page, and adding tags, after adding the first tag, continuously pressing enter adds "empty" tags to the UI.

Seems to be introduced by a fix of QA tool issues in 57b31f6526e3628cc98c8c271108bc34dd2d074e

Before this change `this.search != ''` evaluated to true if search was `null`, with a strict comparison that is no longer the case. The `this.search` is reset to `null` though in some cases.

I thought about simply extending the condition to also check for `!== null`, though imho the better solution would be to always reset it to empty string state, as is the default value. That also matches some click handler value, and is generally both the value on "non-changed" and on "changed but empty".

fixes #6235


asbiin:
Thank you @particleflux

PR #6532, opened 2023-01-23
chore: update psalm

lptn:
Update Psalm and laravel-psalm versions, plus update code based on Psalm's output


asbiin:
Great! Thank you @lptn

Issue #6469, opened 2022-12-11
Call to undefined function Illuminate\View\Compilers\token_get_all()


IngwiePhoenix:
**Describe the bug**
After building the public assets and alike, the application can't render the initial view. Instead, I get a rather surprising bit of output in my FPM log.

**Screenshots**
**Which version are you using:**
 - A server you maintain yourself (if so, please indicate your current version of Monica)

The log:
```
==> php8-fpm.log <==
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "NOTICE: PHP message: PHP Fatal error:  Uncaught Error: Call to undefined function Illuminate\View\Compilers\token_get_all() in /sdcard/srv/monica/vendor/laravel/framework/src/Illuminate/View/Compilers/BladeCompiler.php:250"
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "Stack trace:"
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "#0 /sdcard/srv/monica/vendor/laravel/framework/src/Illuminate/View/Compilers/BladeCompiler.php(159): Illuminate\View\Compilers\BladeCompiler->compileString('@extends('error...')"
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "#1 /sdcard/srv/monica/vendor/laravel/framework/src/Illuminate/View/Engines/CompilerEngine.php(55): Illuminate\View\Compilers\BladeCompiler->compile('/sdcard/srv/mon...')"
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "#2 /sdcard/srv/monica/vendor/facade/ignition/src/Views/Engines/CompilerEngine.php(37): Illuminate\View\Engines\CompilerEngine->get('/sdcard/srv/mon...', Array)"
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "#3 /sdcard/srv/monica/vendor/laravel/framework/src/Illuminate/View/View.php(139): Facade\Ignition\Views\Engines\CompilerEngine->get('/sdcard/srv/mon...', Array)"
[11-Dec-2022 17:26:13] WARNING: [pool www] child 7807 said into stderr: "#4 /sdcard/srv/monica/vendor/laravel/framework/src/Illuminate/View/View.php(122): Illuminate\View\View->getContents..."
[...]

PR #6577, opened 2023-02-14
docs: refers missing php tokenizer dependency

### LINKED ISSUE #6524: Missing dependency to php-tokenizer
**Describe the bug**
Monica fails with a blank page when and
`Call to undefined function Illuminate\View\Compilers\token_get_all` in the laravel log, when the tokenizer php extension is not installed. There should probably a check for it and a useful error message.

SimonVanacco:
This PR adds the php tokenizer extension as a dependency in composer.json, which will prevent user from suffering unexpected consequences such as the cryptic error message in #6469 

Composer will now warn on install  that the extension is needed. Also, the installations docs have been updated to document this dependency.

Closes #6524 and #6469

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
         "ext-gmp": "*",
         "ext-intl": "*",
         "ext-redis": "*",
-        "ext-tokenizer": "*",
         "asbiin/laravel-adorable": "^1.0",
         "asbiin/laravel-webauthn": "^3.0",
         "bacon/bacon-qr-code": "^2.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -15,6 +15,7 @@
         "ext-gmp": "*",
         "ext-intl": "*",
         "ext-redis": "*",
+        "ext-tokenizer": "*",
```
asbiin, line 18: the tokenizer extension is already required by laravel/framework, so I won't add it here
```suggestion
```

asbiin (approved):

PR #6579, opened 2023-02-16
Laravel 10 support

amirsadeghi1:


asbiin:
Thanks for your contribution.
However it's not as easy as upgrading the version number in composer.json.
There are multiple changes to do, I'll open another PR.

Issue #1814, opened 2018-09-11
Stay in touch should be affected by last activity


djaiss:
Feedback from user: 

I've noticed that currently Stay in touch ignores activities. Obviously, reminders should be only sent after a set time from the last "touch" (call, activity, conversation,...).

LukasRos:
Also related to #1970 and #1940.

PetraSophos:
I just got started using Monica. Thanks for all your work on it! This feature would be very useful.

nyalex:
I created a PR for this: https://github.com/monicahq/monica/pull/6586

If the maintainers could review, that would be great.

The logic is that whenever you log a call, conversion, or activity on a contact, the contact's notification date is set to the numbers of days from the current time.

PR #6586, opened 2023-02-23
fix: reset notification date when user adds call, activity, or conversation

nyalex:
Pushes the next notification date from "stay in touch" setting by the number of dates user selected. Intended to do so only when _adding_ calls, activities, or conversations to contact, not updating.

Fixes issue #1814 when adding calls, activities, or conversations to a contact.

### DIFF
--- a/app/Models/Contact/Contact.php
+++ b/app/Models/Contact/Contact.php
@@ -1523,6 +1523,16 @@ class Contact extends Model
     }
 
     /**
+     * Reset the notification date from current time
+     *
+     * @return void
+     */
+    public function resetStayInTouchTriggerDate()
+    {
+        $this->setStayInTouchTriggerDate($this->stay_in_touch_frequency ?? 0);
+    }
+
+    /**
      * Get the weather information for this contact, based on the first address
      * on the profile.
      *
--- a/app/Http/Controllers/Contacts/ConversationsController.php
+++ b/app/Http/Controllers/Contacts/ConversationsController.php
@@ -103,6 +103,9 @@ class ConversationsController extends Controller
                 ->withErrors($result);
         }
 
+        // Reset and update stay in touch trigger date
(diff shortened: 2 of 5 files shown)


nyalex:
Looks like styleci needs to be rerun? Not sure why it is listing that period in my comment...

app/Models/Contact/Contact.php
```diff
@@ -1522,6 +1522,16 @@ public function setStayInTouchTriggerDate($frequency, $triggerDate = null)
...
     }
 
+    /**
+     * Reset the notification date from current time
+     *
+     * @return void
+     */
+    public function resetStayInTouchTriggerDate()
+    {
+        $this->setStayInTouchTriggerDate($this->stay_in_touch_frequency ?? 0);
```
BobChubby, line 1532: Am I reading this right that the trigger gets reset from the time it was recorded? Should it instead be based off of when the event occurred? I'm thinking this could cause issues when retroactively filling out activities

BobChubby: If a user creates a new event, but sets the date of that event in the past, the creation time isn't exactly the same as when they last interacted with the person.

For example, let's say on day 0 I set up a reminder for person X every 7 days. Then on day 6, I retroactively record an event with person X that happened on day 3. I would think I should be reminded about person X on day 10 (7 days after I last interacted with person X on day 3), but instead I'm reminded on day 13 (7 days after I recorded the event on day 6).

Maybe this is preferred since the reminder is based off of when you last thought about the person. Just calling it out as it's not how I originally interpreted the issue.

Scronkfinkle: I agree with @BobChubby . My intuition would be that if I added an event that happened a few days before the entry date, Monica would move the reminder date relative to the date of the event, rather than the creation of the event.

asbiin:

PR #6588, opened 2023-02-24
fix: show todays reminders on the dashboard too

### LINKED ISSUE #6570: Events that are today do not show up in Dashboard
Events that are scheduled for today do not show up in the Dashaboard under Events, only Events starting tomorrow and on.

For example, today is Feb 10 and a contact has a birthday today. His birthday event has been showing up in my Dashbaord up until today. I double-checked that the birthday is still listed in the actual contact, and it is.

Please ensure that events due "today" show up in the Dashboard on the day. Thank you!

nyalex:
I am not sure why the `while()` and then the `if()` here: https://github.com/monicahq/monica/blob/main/app/Models/Contact/Reminder.php#L117-L121

Seems redundant. So, I reworked it to fix #6570 

Could use another pair of eyes to validate my approach, however.

### DIFF
--- a/app/Models/Contact/Reminder.php
+++ b/app/Models/Contact/Reminder.php
@@ -114,11 +114,8 @@ class Reminder extends Model
             $date = $this->initial_date;
         }
 
-        while ($date->isPast()) {
-            $date = DateHelper::addTimeAccordingToFrequencyType($date, $this->frequency_type, $this->frequency_number);
-        }
-
-        if ($date->isToday()) {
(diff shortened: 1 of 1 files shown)


wrufeger:
tested PR locally, but didn't fix problem on dashboard

asbiin:
Hello. Sorry but it does not fix https://github.com/monicahq/monica/issues/6570 which unfortunately will require more deep work.

Issue #6643, opened 2023-06-08
SQL error on people search


nothub:
On a fresh instance, after adding a single person, selecting "All" rows per page in the people search causes an SQL error.

In detail, selecting "All" rows per page in the people search triggers a GET request to url `http://127.0.0.1:8080/people/list?page=1&perPage=-1&search=`. The server responds with error code 500 and the following response message:
```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'offset 0' at line 1 (SQL: select `contacts`.`id`, `contacts`.`first_name`, `contacts`.`middle_name`, `contacts`.`last_name`, `contacts`.`nickname`, `contacts`.`description`, `contacts`.`gender_id`, `contacts`.`account_id`, `contacts`.`created_at`, `contacts`.`updated_at`, `contacts`.`is_partial`, `contacts`.`is_starred`, `contacts`.`avatar_source`, `contacts`.`avatar_adorable_uuid`, `contacts`.`avatar_gravatar_url`, `contacts`.`avatar_default_url`, `contacts`.`avatar_photo_id`, `contacts`.`default_avatar_color` from `contacts` where `contacts`.`account_id` = 1 and `contacts`.`account_id` is not null and `address_book_id` is null and `is_partial` = 0 and `is_active` = 1 and `is_dead` = 0 and `contacts`.`account_id` = 1 and ( `contacts`.`first_name` LIKE '%%' OR `contacts`.`middle_name` LIKE '%%' OR `contacts`.`last_name` LIKE '%%' OR `contacts`.`nickname` LIKE '%%' OR `contacts`.`description` LIKE '%%' OR `contacts`.`job` LIKE '%%' ) and `contacts`.`deleted_at` is null order by `is_starred` desc, `first_name` asc offset 0)
```

Monica Version: `4.0.0`
Docker Image: `monica:4` / `d2e81248d1a49b177a5b182f3a7968407f5f2b685b56bebca3d7cfaf6c0c0c68`

SnoopAir:
I can't confirm the SQL Error but using a fresh installation with importing 80 contacts with monica:4.1.2 I'm experiencing an error as well when selecting "All" rows per page.

Im my case It simple states "Loading..." and nothing else is happening. The container log does not show any error but only:

[...]

PR #6650, opened 2023-06-10
fix: sql error when selecting All pages

MohamAladham:
A fix for #6643


app/Http/Controllers/ContactsController.php
```diff
@@ -740,11 +740,18 @@ public function list(Request $request)
         $perPage = $request->has('perPage') ? $request->input('perPage') : config('monica.number_of_contacts_pagination');
 
         // search contacts
-        $contacts = $contacts->search($request->input('search') ?? '', $accountId, 'is_starred', 'desc', $sort)
-            ->paginate($perPage);
+        $contacts = $contacts->search($request->input('search') ?? '', $accountId, 'is_starred', 'desc', $sort);
+
+        if($perPage == -1){
```
kniziol, line 745: I recommend creating separate methods. It might look like this:
```suggestion
        $total = $this->getTotal($contacts, $perPage);
        $contacts = $this->getContacts($contacts, $perPage);
```
Example of the `getTotal()` implementation:
```php
private function getTotal($contacts, int $perPage): int
{
    if ($perPage === -1) {
        return $contacts->count();
    }

    return $contacts->paginate($perPage);
}
```
Please add a type hint for the `$contacts` argument. The `getContacts()` method might look similar to the above.

asbiin (changes requested):
See @kniziol comment

asbiin:

PR #6751, opened 2023-07-08
feat: implement DAV client subscriptions

asbiin:


app/Domains/Contact/DavClient/Jobs/DeleteVCard.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Send Delete contact.
+     */
+    public function handle(): void
+    {
+        if (! $this->batching()) {
+            return;
+        }
+
+        Log::info(__CLASS__.' '.$this->uri);
```
djaiss, line 36: Do we need to keep all these `Log::info`? Should we keep only the Log::debug or something? I'm afraid this will trigger quite a few alerts otherwise.

stanleyshly:
Can't wait to see this merged into the main chandler branch; DAV features were the last major feature I wanted to see before starting the switch to chandler.

stanleyshly:
How is the DAV features coming along? I can beta test if needed.

asbiin:
> How is the DAV features coming along? I can beta test if needed.

It's working well. Right now it's a little bit too big, so I've splited the work, but it's usable. However there is no UI yet to subscribe properly, this will come in time.

app/Domains/Contact/DavClient/Jobs/DeleteVCard.php
asbiin: I've removed the `Log::info` and kept only the `Log::debug`

vnghia:
Hello, I've just tried this feature and it is working really well. One thing missing is the avatar of the contact. Has the photo synching already been implemented or not ? Thank you very much.

Issue #6445, opened 2022-11-26
Edit older Activities


TheFox:
How can I edit older Activities. The Activities on the Person view are somehow limited to the latest 15 entries, right? How can I edit older Activities?

TheFox:
I see that the latest version on the main branch already has a `Load More` button/link on the right-hand site at the end of the activites list. What would be necessary to add the functionality to customize the Limit of the list? I have a lot of Activites and want to load 100 at once.

PR #6758, opened 2023-07-17
feat: Add filter to journal

khalilJayab:
Based on Issue #6445 , it appears that users are facing difficulties accessing older activities in the application.

In my pull request, I have proposed a solution by introducing a filter that allows users to navigate and access older activities in the journal. This enhancement would greatly improve the user experience by providing a means to conveniently access older activities that may have been missed or require modifications.

Untitled

Thanks.

PART OF GTC OPEN SOURCE INTIATIVE

### DIFF
--- a/app/Http/Controllers/JournalController.php
+++ b/app/Http/Controllers/JournalController.php
@@ -28,10 +28,26 @@ class JournalController extends Controller
      *
      * @return array
      */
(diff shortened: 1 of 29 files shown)


khalilJayab:
done

asbiin (approved):

Issue #6172, opened 2022-05-30
Possibility to change tags


Poldi1977:
In the settings it is possible to delete tags, but not to rename them.

PR #6760, opened 2023-07-19
feat: allow to edit tags in settings

khalilJayab:
Hi,
This pull request addresses issue #6172 , which was related to the inability to rename tags in the settings. With this fix, users can now easily edit the name of existing tags in the settings section.

Untitled

Thanks

PART OF GTC OPEN SOURCE INTIATIVE


app/Http/Controllers/SettingsController.php
```diff
@@ -363,6 +365,36 @@ public function deleteTag($tagId)
...
+    /**
+     * Method editTag
+     *
+     * @param Tag $tag 
+     * @param Request $request
+     *
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    
+    public function editTag(Tag $tag, Request $request)
```
kniziol, line 378:
```suggestion
    public function editTag(Tag $tag, Request $request): RedirectResponse
```
Shall we add the return type?

resources/views/settings/tags.blade.php
```diff
@@ -72,6 +76,10 @@
                     </ul>
                   </div>
                   <div class="table-cell actions">
+                    <a href="javascript:;" class="edit-icon-btn" onclick="toggleEditInput(this)">
```
kniziol, line 79:
```suggestion
                    <div class="edit-icon-btn" onclick="toggleEditInput(this)">
```
Will it work? In this way we can avoid `href="javascript:;"`.

resources/views/settings/tags.blade.php
```diff
@@ -94,3 +102,15 @@
 </div>
 
 @endsection
+<script>
+  function toggleEditInput(editIcon) {
+    var tableRow = editIcon.closest('.table-row');
```
kniziol, line 107:
```suggestion
    let tableRow = editIcon.closest('.table-row');
```
The `let` is recommended when declaring variables
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

kniziol (commented):
Added some suggestions

khalilJayab:
Done, thanks

resources/views/settings/tags.blade.php
```diff
@@ -94,3 +102,15 @@
 </div>
 
 @endsection
+<script>
+  function toggleEditInput(editIcon) {
+    let tableRow = editIcon.closest('.table-row');
+    var tagName = tableRow.querySelector('.tag-name');
```
kniziol, line 108: Would mind you replacing all the other `var`s with `let`s as well? Thanks.

khalilJayab: done

app/Http/Controllers/SettingsController.php
```diff
@@ -363,6 +365,36 @@ public function deleteTag($tagId)
...
+    
+    /**
+     * Method editTag
+     *
+     * @param Tag $tag 
+     * @param Request $request
+     *
+     * @return \Illuminate\Http\RedirectResponse
+     */
+
```
kniziol, line 377:
```suggestion
```
Just one, minor thing

asbiin (approved):

PR #6796, opened 2023-08-16
refactor: fixing test and other small changes

asbiin:
- replacing Eloquent's `Model::where(...)->first()` with `Model::firstWhere(...)`
- add a `group_must_belong_to_vault` validate rule
- refactor validation rules in `BaseService` to ensure the order and dependencies between the rules

### DIFF
--- a/app/Domains/Contact/ManageContact/Web/Controllers/ContactLabelController.php
+++ b/app/Domains/Contact/ManageContact/Web/Controllers/ContactLabelController.php
@@ -18,7 +18,7 @@ class ContactLabelController extends Controller
     {
         $vault = Vault::findOrFail($vaultId);
 
-        $contacts = Label::where('id', $labelId)->first()
+        $contacts = Label::find($labelId)
             ->contacts()
             ->where('vault_id', $request->route()->parameter('vault'))
             ->where('listed', true)
--- a/app/Services/BaseService.php
+++ b/app/Services/BaseService.php
@@ -5,8 +5,10 @@ namespace App\Services;
 use App\Exceptions\NotEnoughPermissionException;
 use App\Models\Account;
 use App\Models\Contact;
+use App\Models\Group;
 use App\Models\User;
 use App\Models\Vault;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Facades\Validator;
 
 abstract class BaseService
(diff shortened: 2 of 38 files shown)


app/Domains/Contact/ManageContact/Web/Controllers/ContactLabelController.php
```diff
@@ -18,7 +18,7 @@ public function index(Request $request, string $vaultId, int $labelId)
     {
         $vault = Vault::findOrFail($vaultId);
 
-        $contacts = Label::where('id', $labelId)->first()
+        $contacts = Label::find($labelId)
```
djaiss, line 21: Why `find` and not `first`?

app/Services/BaseService.php
```diff
@@ -24,7 +26,45 @@ abstract class BaseService
...
+     * The contact object.
+     */
+    public ?Group $group = null;
+
+    /**
+     * Dependencies between permissions.
+     *
+     * @var array<string,array<string>>
+     */
+    private static array $dependencies = [
```
djaiss, line 41: Would `permissionDependencies` be a better name for this? `dependencies` alone seems a bit off.

app/Services/BaseService.php
```diff
@@ -57,35 +97,60 @@ public function validateRules(array $data): bool
...
-            $this->validateAuthorIsAccountAdministrator();
-        }
+        $permissions = collect($this->permissions());
 
-        if (in_array('vault_must_belong_to_account', $this->permissions())) {
-            $this->validateVaultExists($data);
-        }
+        foreach (self::$dependencies as $key => $value) {
+            if ($permissions->contains($key)) {
+                collect($value)->each(function ($v) use ($permissions, $key) {
```
djaiss, line 104: Why `v` as a name?

djaiss (commented):
A few comments. Otherwise, it's great.

app/Domains/Contact/ManageContact/Web/Controllers/ContactLabelController.php
asbiin: `find(xx)` is actually the same as `where('id', xxx)->first()`

app/Services/BaseService.php
asbiin: done, thanks for the suggestion

app/Services/BaseService.php
asbiin: Correct, I've renamed it

PR #6798, opened 2023-08-18
refactor: import and export vcard improvements

asbiin:
Use generics
Add columns to contact and groups

### DIFF
--- a/app/Domains/Contact/ManageContact/Dav/ExportNames.php
+++ b/app/Domains/Contact/ManageContact/Dav/ExportNames.php
@@ -22,7 +22,7 @@ class ExportNames extends Exporter implements ExportVCardResource
     /**
      * @param  Contact  $resource
      */
-    public function export($resource, VCard $vcard): void
+    public function export(mixed $resource, VCard $vcard): void
     {
         $vcard->remove('FN');
         $vcard->remove('N');
--- a/app/Domains/Contact/ManageContact/Dav/ExportNames.php
+++ b/app/Domains/Contact/ManageContact/Dav/ExportNames.php
@@ -8,25 +8,36 @@ use App\Domains\Contact\Dav\Order;
 use App\Models\Contact;
 use Sabre\VObject\Component\VCard;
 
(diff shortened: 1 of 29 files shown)


app/Domains/Contact/ManageContact/Dav/ExportNames.php
```diff
@@ -8,25 +8,36 @@
...
-    public function export(Contact $contact, VCard $vcard): void
+    public function getType(): string
+    {
+        return Contact::class;
+    }
+
+    /**
+     * @param  Contact  $resource
+     */
+    public function export($resource, VCard $vcard): void
```
djaiss, line 25: Shouldn't `$resource` be typed?

app/Helpers/helpers.php
```diff
@@ -67,3 +69,32 @@ function htmldir()
...
+if (! function_exists('subclasses')) {
+    /**
+     * Get all subclass of the given class name.
+     *
+     * @template T of object
+     *
+     * @param  class-string<T>  $className
+     * @return \Generator<array-key,ReflectionClass<T>>
+     */
+    function subclasses(string $className): Generator
```
djaiss, line 82: If we want to "chipote", `subclasses` should be named `subClasses`

djaiss (commented):
My god this is complex.

app/Domains/Contact/ManageContact/Dav/ExportNames.php
asbiin: I would like to type it `T`, but php doesn't accept it yet. It could however been type `mixed`, but it's worthless ...

app/Helpers/helpers.php
asbiin: done

app/Domains/Contact/ManageContact/Dav/ExportNames.php
asbiin: I've typed it `mixed` after all.

PR #6799, opened 2023-08-20
feat: implement Dav for groups

asbiin:
This implements Carddav for groups.

The group type is now optional, carddav doesn't know about the group types and I think it's easier this way - and I think it's better to make it optional anyways.


app/Domains/Contact/ManageContact/Dav/ImportContact.php
```diff
@@ -114,22 +114,17 @@ private function getExistingContact(VCard $vcard): ?Contact
      */
     private function getContactData(?Contact $contact): array
```
djaiss, line 115: Why would we call this function without a contact?

djaiss (commented):
Good work!

app/Domains/Contact/ManageContact/Dav/ImportContact.php
asbiin: It happens when we import a new vcard for the 1st time

app/Domains/Contact/ManageGroups/Web/Controllers/ContactModuleGroupController.php
```diff
@@ -50,9 +51,10 @@ public function store(Request $request, string $vaultId, string $contactId): Jso
             'group_type_role_id' => $request->input('group_type_role_id') ?? null,
         ];
 
-        $group = (new AddContactToGroup())->execute($data);
+        AddContactToGroup::dispatchSync($data);
```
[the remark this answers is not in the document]
asbiin: The benefit is: consistency, by always calling `AddContactToGroup` with dispatch. But it's not a big change, true.

PR #6916, opened 2023-10-06
refactor: remove useless condition

thor-son:
When value is empty string, it's caught in empty check condition and return null.
So suggest removing useless condition.

```
$val = '';
var_dump(empty($val)); // true
var_dump($val == '');  // true
```


asbiin:
Thank you @thor-son

PR #6928, opened 2023-10-09
fix: fix incorrect connection between date label and input

connordoner:
This is a possible accessibility problem.


roman-1983 (commented):
✅ Looks good to me

Issue #6983, opened 2023-10-30
[Chandler] Storage Limit 0 is not unlimited


DunklerPhoenix:
## Bug description

The ENV  "DEFAULT_STORAGE_LIMIT=0" sets the limit in the account table to 0, but chandler doesn't see that as unlimited. It sees it as 0b

## How to reproduce

1. set ENV DEFAULT_STORAGE_LIMIT to 0
2. create account
3. look at /settings/storage and see 0b
4. try to upload a file and you get "no space left" (not the real text. I have it in german)

## Environment

self-hosted instance

DunklerPhoenix:
Fixed by https://github.com/monicahq/monica/pull/6989

Thank you

PR #6989, opened 2023-10-31
fix: storage_limit can be unlimited

SecurID:
#6983


asbiin:
Great fix @SecurID Thanks!
Can you also add a test in https://github.com/monicahq/monica/blob/main/tests/Unit/Helpers/StorageHelperTest.php ?

SecurID:
Sure :)

PR #7037, opened 2023-11-22
Use correct env variable for mail driver

flohoss:
MAIL_MAILER didnt work for my installation, found out after looking at the docker compose files


asbiin:
`MAIL_MAILER` is the right [environment variable name](https://github.com/monicahq/monica/blob/6ec0735a226ef45863b343b2270de9f3f840e429/config/mail.php#L16). `MAIL_DRIVER` is the old value.
So actually both variables should work.

PR #7099, opened 2024-01-02
feat: remove twitter socialite provider

asbiin:
Since Elon Musk's take over of Twitter, this social network became a place of arrogance, racism, and conspiracy, making it the heaven for extreme right thoughts.
I can no longer support this social network, and so I'm removing the login provider support for it.


djaiss (approved):
I completely agree. This social network has become a cancer for society.

PR #7138, opened 2024-01-30
feat: Increase storage

Bhuvanasundari:
How do I increase storage on Chandler beyond 50MB?

### DIFF
--- a/app/Domains/Settings/ManageStorage/Web/ViewHelpers/StorageIndexViewHelper.php
+++ b/app/Domains/Settings/ManageStorage/Web/ViewHelpers/StorageIndexViewHelper.php
@@ -38,7 +38,7 @@ class StorageIndexViewHelper
             ->where('type', File::TYPE_AVATAR)
             ->count();
 
-        $accountLimit = $account->storage_limit_in_mb * 1024 * 1024;
+        $accountLimit = 64 * 1024 * 1024;
 
         return [
             'statistics' => [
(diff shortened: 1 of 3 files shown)


Bhuvanasundari:
I increase the 50MB to 64MB changes occur in storage

app/Domains/Settings/ManageStorage/Web/ViewHelpers/StorageIndexViewHelper.php
```diff
@@ -38,7 +38,7 @@ public static function data(Account $account): array
             ->where('type', File::TYPE_AVATAR)
             ->count();
 
-        $accountLimit = $account->storage_limit_in_mb * 1024 * 1024;
+        $accountLimit = 64 * 1024 * 1024;
```
asbiin, line 41: We don't want to change this, the accountLimit comes from this calculation

app/Helpers/StorageHelper.php
```diff
@@ -25,7 +25,7 @@ public static function canUploadFile(Account $account): bool
 
         $totalSizeInBytes = File::whereIn('vault_id', $vaultIds)->sum('size');
 
-        $accountLimit = $account->storage_limit_in_mb * 1024 * 1024;
+        $accountLimit = 64 * 1024 * 1024;
```
asbiin, line 28: We don't want to change this, the accountLimit comes from this calculation

asbiin (changes requested):
Thank you, but I think this needs more work.

djaiss:
This is not something that we want.

If you need more storage, simply increase your ENV variable in your `.env` file.

The content of your env file should be 

```
DEFAULT_STORAGE_LIMIT=64
```

PR #7178, opened 2024-03-06
docs: update to "What Monica is" section

DeeprSoni:
Added more use cases for what Monica can be used for.

The current version is too short and fails to mention other important use-cases which may hinder usage of the software.

Please do review the changes made and do revert to me on any further revision required on this section.


djaiss:
@DeeprSoni thanks for this change.

This is much better, and also, it looks like it comes from a GPT prompt somehow :-)

Anyway, we'll merge it, and I'll make a few changes.
But it's still much better you are right.

djaiss:
After reading it again, it's too "chatGPT-y" for me. We don't need that much details, I think. Thanks anyway for the effort and the contribution 🫶

Issue #7171, opened 2024-02-27
"Unknown Age" for babies


slifty:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/monicahq/monica/issues?q=is%3Aopen+is%3Aissue+label%3Abug) _(I've searched it)_.
- [X] I agree to follow Monica's [Code of Conduct](https://github.com/monicahq/monica/blob/main/CODE_OF_CONDUCT.md).

### Bug description

If a contact is less than 1 year old their age is reported as "unknown

### Steps to reproduce

1. Create a new contact
2. Set their birthday to be less than 1 year ago (e.g. the previous month)
3. View the "activity in this vault" tab and they will have an unknown age.  If they are family members / in a group they will not have a parenthetical `(0)` like other contacts with birthdays would

### Expected behavior

I would expect the age to be rendered as `0` (or in terms of months)

### Environment

Your own self-hosted instance (v5 a.k.a chandler version or main branch)

### Version of Monica

v5-beta.3

### Installation method

Docker image

### Web server

Nginx

### Database engine version

PostgreSQL

Laurell876:
this issue is resolved 
![image](https://github.com/user-attachments/assets/fa30a4d9-fe2f-4beb-8b31-54016d6ab6cb)

PR #7184, opened 2024-03-09
fix: unknown age for babies

abdabdii:
"Unknown Age" for babies #7171 - fixed


asbiin (commented):
Thank you! See my comment :

PR #7193, opened 2024-03-15
chore: upgrade to Laravel 11

asbiin:


composer.json
```diff
@@ -9,32 +9,45 @@
     "social"
   ],
   "license": "AGPL-3.0-or-later",
+  "repositories": [
```
asbiin, line 12: We have to wait for these packages dependencies to be solved before merging

PR #7254, opened 2024-05-02
feat: Disable Signups for new users

### LINKED ISSUE #6687: How do disable Signups for new users?
On Chandler (selfhosted), how do I disable signups? That way I can host the service at monica.mydomain.com without having to worry about other people signing up and using storage space.

davpsh:
These changes allow to enable/disable new user registration based on ENV setting. The feature was implemented in version 4.x and was not present in chandler. Fixes #6687 

Changes:

- added new environment parameter `APP_DISABLE_SIGNUP`
- restricting registration access based on env setting value


app/Http/Middleware/EnsureSignupIsEnabled.php
```diff
@@ -0,0 +1,32 @@
...
+    public function __construct(SignupHelper $signupHelper, Translator $translator)
+    {
+        $this->signupHelper = $signupHelper;
+        $this->translator = $translator;
+    }
+
+    public function handle(Request $request, Closure $next): Response
+    {
+        if (!$this->signupHelper->isEnabled()) {
+            abort(Response::HTTP_FORBIDDEN, $this->translator->get('auth.signup_disabled'));
```
davpsh, line 27: Same behavior as in Monica 4.x.

phpunit.xml
```diff
@@ -8,6 +8,7 @@
     <testsuite name="Unit">
       <directory suffix="Test.php">./tests/Feature</directory>
       <directory suffix="Test.php">./tests/Unit</directory>
+      <directory suffix="Test.php">./tests/PHPUnit</directory>
```
asbiin, line 11: Please don't create another testsuite. Those tests should be placed under `Unit`

tests/Feature/Auth/RegistrationTest.php
```diff
@@ -1,47 +1,67 @@
...
+use Mockery;
 use Tests\TestCase;
 
 class RegistrationTest extends TestCase
 {
     use DatabaseTransactions;
 
-    #[Test]
-    public function registration_screen_can_be_rendered()
+    public function testAccessToRegistrationPage(): void
```
asbiin, line 18: Why removing `#[Test]` ?

app/Providers/FortifyServiceProvider.php
```diff
@@ -57,4 +63,19 @@ public function boot()
...
+        }
+
+        $router = $this->app->make(Router::class);
+        $routes = $router->getRoutes();
+        collect(['create', 'store'])->each(function ($method) use ($routes) {
+            if ($route = $routes->getByAction(RegisteredUserController::class . '@' . $method)) {
+                $route->middleware('monica.signup_is_enabled');
+            }
+        });
+    }
```
asbiin, line 80: This should go directly in the `routes/web.php` file, there is no need to "patch" the route

asbiin (changes requested):
Thank you for the work.
However it is a little bit more complicated than needed, please see my reviews.

app/Http/Middleware/EnsureSignupIsEnabled.php
```diff
@@ -0,0 +1,32 @@
...
+        }
```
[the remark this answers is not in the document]
davpsh: > We don't manage those kind of translation through lang php files.

@asbiin, maybe then it’s better to delete lang php files in a separate task so as not to get confused.

tests/Feature/Auth/RegistrationTest.php
davpsh: @asbiin, it is enough that the name of the method begins with the word “test” and the phpunit recognizes it as a test method. The attribute is not needed here. see https://docs.phpunit.de/en/11.1/attributes.html#test

app/Providers/FortifyServiceProvider.php
davpsh: @asbiin, no, it shouldn't be in the `routes/web.php`.

Routes and controller are part of the `fortify` package. I don’t want to redefine the entire route, we just need to add our custom middleware, that’s why we alter the routes in the service provider.

phpunit.xml
davpsh: @asbiin, Unit tests in Laravel are not a classic pure unit test, because they bootstrap the kernel, database, and much more. A pure unit test only works with abstractions and does not depend on anything, for this reason it is cleaner and faster. I can move the tests to the unit folder, but I would still suggest separating test types

davpsh: @asbiin, change reverted

davpsh:
> Thank you for the work.
> However it is a little bit more complicated than needed, please see my reviews.

@asbiin, can you give feedback on the current state of the code? Is there anything else that needs to be fixed?

tests/Feature/Auth/RegistrationTest.php
```diff
@@ -1,47 +1,68 @@
...
```
asbiin, line 18: Let's keep the new syntax
```suggestion
    #[Test]
    public function registration_screen_can_be_rendered(): void
```

tests/Feature/Auth/RegistrationTest.php
```diff
@@ -1,47 +1,68 @@
...
+
 namespace Tests\Feature\Auth;
 
+use App\Helpers\SignupHelper;
 use Illuminate\Foundation\Testing\DatabaseTransactions;
-use Laravel\Fortify\Features;
+use Illuminate\Http\Response;
 use Laravel\Jetstream\Jetstream;
-use PHPUnit\Framework\Attributes\Test;
+use Mockery;
```
asbiin, line 11:
```suggestion
use PHPUnit\Framework\Attributes\Test;
use Mockery;
```

tests/Feature/Auth/RegistrationTest.php
```diff
@@ -1,47 +1,68 @@
...
-        $response->assertStatus(200);
+        $isSignupEnabled = false;
+        $response = $this->get('/register');
+        $response->assertStatus(Response::HTTP_FORBIDDEN);
+        $response->assertSeeText('Registration is currently disabled');
     }
 
-    #[Test]
-    public function new_users_can_register()
+    public function testRegistration(): void
```
asbiin, line 40:
```suggestion
    #[Test]
    public function new_users_can_register()
```

asbiin (changes requested):
Thank you @davpsh I left some comments to fix the code, I've tested it but I got an issue running it ...

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -14,6 +15,11 @@
 
 class LoginController extends Controller
 {
+    public function __construct(
+        protected SignupHelper $signupHelper
+    ) {
+    }
+
```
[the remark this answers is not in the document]
davpsh: @asbiin, done

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -40,6 +46,7 @@ public function __invoke(Request $request): Response
         }
 
         return Inertia::render('Auth/Login', $data + [
+            'isSignupEnabled' => $this->signupHelper->isEnabled(),
```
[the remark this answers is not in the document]
davpsh: @asbiin, fixed

app/Http/Middleware/EnsureSignupIsEnabled.php
```diff
@@ -0,0 +1,25 @@
...
+class EnsureSignupIsEnabled
+{
+    public function __construct(
+        protected SignupHelper $signupHelper,
+    ) {
+    }
+
+    public function handle(Request $request, Closure $next): Response
+    {
+        abort_if(! $this->signupHelper->isEnabled(), 403, trans('Registration is currently disabled'));
```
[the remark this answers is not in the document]
davpsh: @asbiin, fixed

tests/Feature/Auth/RegistrationTest.php
davpsh: @asbiin, fixed

tests/Feature/Auth/RegistrationTest.php
davpsh: @asbiin, fixed

davpsh:
@asbiin, added fixes for lint errors.

asbiin:
> @asbiin, added fixes for lint errors. [bee8748](https://github.com/monicahq/monica/commit/bee87484afaf321406ac96e03fcf83ed49170133)
> 
> image

Thank you but that's not necessary, there is a misconfiguration between eslint and prettier, so prettier will revert that in the end...

asbiin (approved):

PR #7282, opened 2024-05-12
feat: add typesense connector

asbiin:
Adding `typesense` scout driver.
Also fixing the `database` driver, and the ability to switch between drivers. We need to create the full text indexes on database, unless you explicitly set the `SCOUT_FULL_TEXT_INDEX` variable to `false`.


app/Helpers/ScoutHelper.php
```diff
@@ -2,27 +2,84 @@
...
 {
     /**
      * When updating a model, this method determines if we should update the search index.
      *
-     * @return bool
-     *
      * @codeCoverageIgnore
      */
-    public static function activated()
+    public static function activated(): bool
```
djaiss, line 15: Technically, if it's a boolean, the function should be `isActivated` or something like that.

app/Helpers/ScoutHelper.php
```diff
@@ -2,27 +2,84 @@
...
                 return false;
         }
     }
+
+    /**
+     * Test if the driver requires indexes.
+     *
+     * @codeCoverageIgnore
+     */
+    public static function indexed(): bool
```
djaiss, line 37: Same comment as above. `isIndexed` or something.

djaiss (commented):
This ia a great PR. Congrats.

PR #7378, opened 2024-08-22
feat: Shell script for initializing sail

LukasRos:
A quick helper script to assist with setting up Docker Sail on a new system.


djaiss:
Thanks. This will be super useful indeed.

PR #7386, opened 2024-09-17
fix: join relationship query with undeleted contacts

phblj:
Currently if two contacts have a relationship and one is deleted, you'll get a server error when viewing the other.  This change just filters the relationships query to only return rows that reference two active contacts


phblj:
Sorry, this was intended for `beta`

Issue #7359, opened 2024-07-16
first_name single quote broke mariadb sql 


c0derpwner:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/monicahq/monica/issues?q=is%3Aopen+is%3Aissue+label%3Abug) _(I've searched it)_.
- [X] I agree to follow Monica's [Code of Conduct](https://github.com/monicahq/monica/blob/main/CODE_OF_CONDUCT.md).

### Bug description

single quotes brake application in first_name 

### Steps to reproduce

- /people/add endpoint can upload a new contact on our crm
- if i put in  first_name filed `giu ' lia` then this can be uploaded 
- when app redirects on the `people/h:id` the page is blank! 
- maybe there's a potentially blind sql injection
- version of monica is 4.1.2

### Expected behavior

first_name can be sanitized ?

### Environment

app.monicahq.com (monica v4)

### Version of Monica

4.1.2

### Installation method

None

### Web server

Apache

### Database engine version

MariaDB

### Additional info

Thanks in advice for every new patch!

jonathan-shahen:
Can confirm this bug.

If you get into this problem, just add `/edit` to the url and then you can remove the offending character and everything works again

c0derpwner:
Sent with [Proton Mail](https://proton.me/mail/home) secure email.
will you gave me an assign related to a cve for that?

PR #7400, opened 2024-10-10
fix: first_name single quote broke mariadb sql 

shreyashrpawar:
#7359


asbiin:
The bug seems to be related to monica v4.x, so it must target `4.x` branch.

Issue #7393, opened 2024-09-30
Overflow Menu Button for Mobile View


Masitulah06:
While using an Android device with a 5 inch display, page header is saturated with _username_, _search bar, settings and log out button_ being on almost the same line. 

Adding a **menu overflow button** which when tapped will display settings, account info, and logout buttons, help declutter the header ribbon.

PR #7402, opened 2024-10-10
Overflow Menu Button for Mobile View #7393

shreyashrpawar:
#7393


djaiss:
Thanks so much for this.
i went ahead and actually fix the entire UI instead of just the header in this one #7424

PR #7435, opened 2024-10-31
Improve .env.example with setup guidance and placeholders

AzeemSup:
Added helpful comments throughout the .env.example file to clarify how each environment variable is used. Also included placeholders for API keys and other important variables to guide users through the setup process.


asbiin:
Thank you, but it's easier, and more secure to left values empty actually, instead it will use the fake values if you don't edit the config file

Issue #7177, opened 2024-03-05
"Create Label" is not rendered if there are any partial matches


slifty:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/monicahq/monica/issues?q=is%3Aopen+is%3Aissue+label%3Abug) _(I've searched it)_.
- [X] I agree to follow Monica's [Code of Conduct](https://github.com/monicahq/monica/blob/main/CODE_OF_CONDUCT.md).

### Bug description

When adding a label to a contact it is possible to "create" a label, however if the label is a partial match to an existing label that option isn't rendered

### Steps to reproduce

1. Create a new contact
2. Attempt to add label "Pineapple" (and create the label "Pineapple")
3. Attempt to add the label "apple" -- see the lack of "create new label" button

### Expected behavior

I'd expect to see the list of potential matches but ALSO the "create new label" button in the event that there is no exact match.

### Environment

Your own self-hosted instance (v5 a.k.a chandler version or main branch)

### Version of Monica

5.0-beta3

### Installation method

Docker image

### Web server

Nginx

### Database engine version

PostgreSQL

PR #7454, opened 2024-11-11
fix: allow valid labels to be created on search

Laurell876:
This PR addresses the following bug: https://github.com/monicahq/monica/issues/7177

Users can create labels that form a substring of existing labels. Example: "pineapple" label already exists, now they can add the "apple" label.

I also moved the "create new label" button to be shown before the list to prevent having to scroll to see the button if the list of labels gets too long.

image
image


asbiin (approved):
Nice job!

Issue #5873, opened 2022-01-11
Implement DiceBear Avatars


asbiin:
Initials Avatars and Adorable avatars are limited.
I'd like to implement more kind of avatars, and let the user choose the style.

We could use https://avatars.dicebear.com/

Laurell876:
> Initials Avatars and Adorable avatars are limited. I'd like to implement more kind of avatars, and let the user choose the style.
> 
> We could use https://avatars.dicebear.com/

Hi, this is something I'd like to work on. Is this still a feature you'd like added?

PR #7457, opened 2024-11-14
Add dicebear avatars to app

Laurell876:
The user can select a dicebear style (hardcoded list) and that avatar style will be used acrossed the app for all contacts.

This can be done on the user preferences section of the settings page.

image
image
image
image
image

Related to issue: https://github.com/monicahq/monica/issues/5873


app/Helpers/AvatarHelper.php
```diff
@@ -11,21 +11,39 @@ class AvatarHelper
...
             $name = Faker::create()->name();
         } else {
             $name = $contact->first_name.' '.$contact->last_name;
         }
 
-        return $multiavatar($name, null, null);
+        if ($avatarStyle) {
+            return [
+                'type' => Contact::AVATAR_TYPE_URL,
+                'content' => self::getDicebearLink($avatarStyle, $name),
```
asbiin, line 33: What if we use the id (uuid) so it's always the same for one contact?
```suggestion
                'content' => self::getDicebearLink($avatarStyle, $contact->id),
```

asbiin (commented):
That's a really great job! thank you

PR #7535, opened 2025-02-04
ci: Analyze mail templates with Bladestan (PHPStan extension)

AJenbo:
[Bladestan](https://github.com/bladestan/bladestan) is a PHPStan extension that compiles and analyzes Blade templates used in a Laravel application. In this way it's similar to how Larastan helps PHPStan understand code unique to Laravel.

It was proposed to use Monica as an e2e test suite for Bladestan, so thought I might send you this PR while investigating things.

There where no issues discovered at the current run level for the project, but this is a good way to make sure it stays that way :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,8 @@
     "roave/security-advisories": "dev-master",
     "spatie/laravel-ignition": "^2.4",
     "stichoza/google-translate-php": "^5.1",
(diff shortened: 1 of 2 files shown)


AJenbo:
@asbiin I rebased

asbiin:
All set, thank you @AJenbo

PR #7538, opened 2025-02-09
feat: Mentions

Ali1:
This push add 2 main features:
1) Mention feature for journal posts (by incorporating the Tribute js package) which I think would be a great addition to monica with many possible further development potential.
2) Work on switching the main tab of the contact page into a broader activity timeline rather than an activity feed. In my mind, an activity timeline would as contain a broader set of contact activity feed more like a social media feed with all various updates (feed activity, posts, notes, docs, photos, other dated entries) into one timeline which would serve as the central place to look and scroll down when reviewing a contact.

image

image

image
(See the last item in the activity feed. This has nothing to do with mentions, it relies on which contacts are linked to the post in the existing way).

In this commit, mentions look like this in the edit box @"Tony Smith", and go into the database as a token {{{CONTACT-ID:xxxx-xxxx-xxxxxx|Tony Smith}}}. The reason the name is also in the token is to act as a fall back name only, should the contact get deleted, the user's intent does not get lost. The view will always attempt to retrieve the up to date name information from the databasae.

There are no breaking changes.

Currently the only contacts that can be selected are the ones already linked to the post in the submenu. My preference for monica is actually removal of contacts in this post form entirely and relying on mentions to connect the post to users only. That way, any contacts in the journal can be picked when making a mention. This can be introduced in a non-breaking way if needed (by adding all the mentions at the bottom to all previously linked contacts to all posts).

To do:
1) Removing a contact from a post should prompt a warning to remove mentions first.
2) Adding or messing a mention into a name that's not recognised as one of the contacts in the post (for example @"Tony Smi") should return an error
3) Disallow mentions for contacts where there are more than one contact attached to a post that have the same name and therefore the mention becomes ambiguious to which contact ID should be attached.
4) Develop the activity timeline further. I think it would be fantastic if many more various dated items relatated to the contact are incorporated (notes, docs etc.)
5) For journal posts in the timeline, I would like to see them summarised into an except and where the contact is @mentioned, those paragraphs and sentances are the ones that show in the excerpt that's generated.
6) Investigate why pre-existing contact's in the post look different to newly added contacts (for example contact prefix only appears for new contacts). This is causing some issues with saving tags.
7) If proceeding with a broader activity timeline, a way to filter which activities to display and which to hide. Would some of the other tabs become redundant if we do this? For example is a seperate section on posts needed if we can easily set the filter of the activity timeline to show posts only with one click?

I wanted the thoughts of @djaiss @asbiin and the community before proceeding further.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,5 +59,9 @@
       "vendor/bin/pint"
     ]
   },
-  "packageManager": "yarn@4.6.0"
+  "packageManager": "yarn@4.6.0",
+  "dependencies": {
+    "tributejs": "^5.1.3",
+    "vue-tribute": "^2.0.0"
+  }
 }
--- a/app/Domains/Contact/ManageContact/Web/ViewHelpers/ContactShowViewHelper.php
+++ b/app/Domains/Contact/ManageContact/Web/ViewHelpers/ContactShowViewHelper.php
@@ -47,6 +47,7 @@ class ContactShowViewHelper
         $templatesPagesCollection = self::getTemplatePagesList($templatePages, $contact, $firstPage);
 
         return [
+            'id' => $contact->id,
             'contact_name' => ModuleContactNameViewHelper::data($contact, $user),
             'listed' => $contact->listed,
             'template_pages' => $templatesPagesCollection,
@@ -101,6 +102,7 @@ class ContactShowViewHelper
         $templatePages = $contact->template->pages()->orderBy('position', 'asc')->get();
 
         return [
+            'id' => $contact->id,
             'contact_name' => ModuleContactNameViewHelper::data($contact, $user),
             'listed' => $contact->listed,
             'template_pages' => self::getTemplatePagesList($templatePages, $contact, $templatePage),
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -59,5 +59,9 @@
       "vendor/bin/pint"
     ]
   },
-  "packageManager": "yarn@4.6.0"
+  "packageManager": "yarn@4.6.0",
+  "dependencies": {
+    "tributejs": "^5.1.3",
+    "vue-tribute": "^2.0.0"
```
asbiin, line 65: This needs to go to the `devDependencies`

asbiin (commented):
This is outstanding!
I need to test it though, let me come back to you soon ..

Ali1:
Thanks asbiin
Having thought a bit about the contact view, what do you think of these changes.

Instead of the current tabs:
Activity feed, Ways to connect, Social, Life & Goals, Information
New tabs
Timeline, Activity feed, Ways to connect, Social, Information

Timeline would show all the posts, notes, documents, photos, calls, life events with a filtering system at the top to show/hide these. There would also be buttons at the top to create each of these different events.
Activity feed - like the current one, acts a bit like an audit trail
Social: unchanged
Life & Goals: life has gone to timeline, goals goes to the next tab Information
Information: This would have fewer components since all the dated items have moved to timeline

Issue #7320, opened 2024-06-05
Contact name misorder WHEN searching DURING 'Adding Relationship'


adrianmross:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/monicahq/monica/issues?q=is%3Aopen+is%3Aissue+label%3Abug) _(I've searched it)_.
- [X] I agree to follow Monica's [Code of Conduct](https://github.com/monicahq/monica/blob/main/CODE_OF_CONDUCT.md).

### Bug description

Contact name displays <First> <Last> <Middle> in the search results

### Steps to reproduce

1. SELECT any contact 
2. NAVIGATE to the 'Social' tab
3. CLICK the '+ Add a relationship' button to NAVIGATE to the 'Add a relationship' form
4. CHOOSE any relationship from the drop down
5. SELECT `Choose an existing contact`
6. CLICK the '+ add a contact' button to initialize the `input` search 'Find a contact in this vault'
7. TYPE target contact and the contact with the misordered name will show

### Expected behavior

Contact name should display <First> <Middle> <Last> as you type the name.

### Environment

beta.monicahq.com (v5 a.k.a chandler version or main branch)

### Version of Monica

v5.0.0-beta.4 — commit bad6967

### Installation method

Other

### Web server

Apache

### Database engine version

Other

### Additional info

Not running locally. Using the hosted [beta.monicahq.com](https://beta.monicahq.com/)

Mr1131:
i would like to fix this.

ivanDevPHP:
@mrsoneri  Are you still working on this issue? I noticed it's still open, and the bug is still present in the beta version. I'd like to work on it as well. 
@adrianmross Could you confirm the expected order for contact names in the search results? Should they be sorted alphabetically by first name, last name, or follow a different criteria?

hubikj:
@ivanfslDEV I am on v5.0.0 Beta 5 and my contact search in add relationship dialog did not work at all (giving 500 error in the dev console). I found out it happens because the attribute `first_name` is not sortable:
```json
{
"message": "Index `contacts`: Attribute `first_name` is not sortable. Available sortable attributes are: `updated_at`.",
"exception": "Meilisearch\\Exceptions\\ApiException",
"file": "/var/www/html/vendor/meilisearch/meilisearch-php/src/Http/Client.php",
"line": 186,
"trace": [
[...]

PR #7568, opened 2025-02-22
fix(contact-search): ensure contact names display correctly during search

ivanfslDEV:
Fix for #7320 

**What I did:**
- Fixed the contact search issue where names displayed incorrectly or only worked when typing the full name.
- Adjusted the search query to ensure contact names appear correctly as you type, even with partial matches.
- Sorted results by first and last name for consistency.

**Why I did it:**
- In my local environment (Apache and MySQL), the search only returned results when typing the full name. This fix improves the search functionality for partial name inputs.


app/Domains/Vault/Search/Web/ViewHelpers/VaultContactSearchViewHelper.php
```diff
@@ -11,8 +11,16 @@ class VaultContactSearchViewHelper
     public static function data(Vault $vault, string $term): Collection
     {
         /** @var Collection<int, Contact> */
-        $contacts = Contact::search($term)
-            ->where('vault_id', $vault->id)
+        $contacts = Contact::where('vault_id', $vault->id)
```
asbiin, line 14: We *want* to use `Contact::search` because it uses Laravel Scout
https://laravel.com/docs/12.x/scout#searching
Basically, the `where` here are all covered by this function: https://github.com/monicahq/monica/blob/3dce1d67bb12c4a5defa0fde5021b751749d8627/app/Models/Contact.php#L88

I suggest to keep the `orderBy`, which could be useful indeed

asbiin (changes requested):

app/Domains/Vault/Search/Web/ViewHelpers/VaultContactSearchViewHelper.php
ivanDevPHP: Done! Thanks for the feedback

PR #7581, opened 2025-02-25
feat(auth): added support for keycloak auth provider

jargij:
This feature adds Keycloak (OAuth2 / OIDC) as Authentication Provider.

Example `.env` settings:

```
LOGIN_PROVIDERS=keycloak

KEYCLOAK_CLIENT_ID=monica
KEYCLOAK_CLIENT_SECRET=XXXXXXXXXXXXXXXXx
KEYCLOAK_BASE_URL=https://keycloak.yourowndomain.com
KEYCLOAK_REALM=master

### CODE REVIEW

File: bootstrap/providers.php [hunk 1]
```diff
@@ -7,4 +7,5 @@
     App\Providers\FortifyServiceProvider::class,
     App\Providers\JetstreamServiceProvider::class,
     App\Providers\TelescopeServiceProvider::class,
+    \SocialiteProviders\Manager\ServiceProvider::class,
```
asbiin, line 10: We don't need that actually
```suggestion
```

asbiin (changes requested):
Great!
Just 1 change to fix, appart from that it's OK

bootstrap/providers.php
jargij: Oh wauw, not sure how it ended up there.

jargij:
@asbiin Also resolved the merge conflict

asbiin (approved):

PR #7591, opened 2025-03-06
refactor: Fixing some code smells AISS IS3-L5-5

HenriquePRebolho:


asbiin (changes requested):

database/factories/UserFactory.php
```diff
@@ -42,11 +42,7 @@ public function definition(): array
      */
     public function unverified()
     {
-        return $this->state(function (array $attributes) {
-            return [
-                'email_verified_at' => null,
-            ];
-        });
+        return $this->state(function () => ['email_verified_at' => null,]);
```
asbiin, line 45:
```suggestion
        return $this->state(fn () => ['email_verified_at' => null,]);
```

PR #7626, opened 2025-03-27
fix: update is past address for contact address

vnghia:
`is_past_address` is part of `contact_address` table so it should be `updateExistingPivot` instead of `update` which will update the `addresses` table and throw an exception


asbiin (approved):

vnghia:
Hi @asbiin, thanks for your review. Do I need to do anything since I see that one check is pending ?

asbiin:
> Hi @asbiin, thanks for your review. Do I need to do anything since I see that one check is pending ?

Thank you, that's ok, I've merged it ;)

Issue #7675, opened 2025-04-25
[Chandler] contact reminders showing up twice in dashboard => all planned reminders


cmpickle:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [x] This issue is **not** already reported on [Github](https://github.com/monicahq/monica/issues?q=is%3Aopen+is%3Aissue+label%3Abug) _(I've searched it)_.
- [x] I agree to follow Monica's [Code of Conduct](https://github.com/monicahq/monica/blob/main/CODE_OF_CONDUCT.md).

### Bug description

When I create a reminder date in Chandler it shows up twice in the reminder screen in Dashboard => all planned reminders. I checked in the DB and there is only a single contact_reminder entry for each.  ![Image](https://github.com/user-attachments/assets/1c432a73-8c97-46b1-b08c-76b2ceed656c)  ![Image](https://github.com/user-attachments/assets/ec544332-df0a-4399-b31d-8de3ebf53414)

### Steps to reproduce

1. Create a new Chandler instance
2. create a contact
3. add an important date with the yearly reminder checked
4. go to dashboard => all planned reminders
5. Each reminder date will show up twice

### Expected behavior

1. Create a new Chandler instance
2. create a contact
3. add an important date with the yearly reminder checked
4. go to dashboard => all planned reminders
5. Each reminder date should show up once

### Environment

Your own self-hosted instance (v5 a.k.a chandler version or main branch)

### Version of Monica

5.0.0-beta.4

### Installation method

Docker image

### Web server

Apache

### Database engine version

MySQL

cmpickle:
I see where the issue comes from. In the table `contact_reminder_scheduled` there is a reminder created per notification channel, because of this it show twice in that view. Instead of pulling from that table to show the reminders to the UI it should probably be pulling from `contact_reminders`

CodeLeom:
Can this issue be assigned to me?

PR #7697, opened 2025-05-09
fix: remove duplicate reminders displayed in vault reminders when multiple notification channels exist

cmpickle:
Fixes issue #7675 where multiple contact reminders show up in the vault view as well as the vault reminders page. I added the contact_reminders_id and de-duped based on that ID.

Before fix:
image
image

After fix:
image
image


asbiin (approved):
LGTM

PR #7705, opened 2025-05-16
feat(auth): added support for authentik auth provider

UndyingSoul:
Adds authentik as an auth provider. Based on the simplicity of https://github.com/monicahq/monica/pull/7625


vnghia:
I think you will need to add Authentik logo as well

Issue #7755, opened 2025-07-02
[Feature Request] Enhance journal input with a Markdown / WYSIWYG editor for families ✨


zunami:
Hi Monica team 👋

I self-host Monica for my family—my wife and our kids use the journal daily to capture thoughts, dreams, and everyday moments. We absolutely love Monica! 💜

However, plain text entries can feel flat and hard to read. We would greatly benefit from rich text formatting like:

- **Bold**, *italic*, _underlined_ text  
- 🔗 Clickable links  
- Tables  
- Embedded images  
- Indentation, lists, headers

We are *not* developers and don’t want to learn Markdown syntax. A user-friendly WYSIWYG editor or live Markdown preview would be ideal.

**Why it matters:**  
- My wife wants to format memories and recipes beautifully.  
- Our children want to illustrate their experiences with images and highlights—without copy-pasting or special markup.  
- A visually pleasing journal becomes a cherished family keepsake.

**User Story:**  
> As a family, we want to enrich our journal entries with formatting and images so they’re clearer, more expressive, and enjoyable to revisit later.

**Suggested implementation approaches:**  
- Integrate a WYSIWYG JS editor (e.g. Quill, Tiptap, TinyMCE)  
- Or support simple Markdown with real-time preview as an option

This feature would greatly enhance the experience for self-hosting families, and I’m confident many Monica users would appreciate it. Thank you for all your amazing work—let’s make this even more special! 😊

#6831 
#7674

jhedmendoza:
Hi @zunami 

I'm not sure how to assign the issue to myself, but I'm working on it now. I'll implement a WYSIWYG editor.

jhedmendoza:
https://github.com/monicahq/monica/pull/7912

jhedmendoza:
I'm done @zunami. Let’s wait for the main developer's approval of the changes I made to the code. Here are some screenshots from my local env.

<img width="1183" height="951" alt="Image" src="https://github.com/user-attachments/assets/97b4e1da-356b-4f2a-a2a9-826d168b1f0a" />

<img width="765" height="830" alt="Image" src="https://github.com/user-attachments/assets/6da8e9a2-52a9-450c-8daa-baeca1568419" />

zunami:
> I'm done [@zunami](https://github.com/zunami). Let’s wait for the maintainer's approval of the changes I made to the code. Here are some screenshots from my local env.


Then I look forward to testing it.

MrSuddenJoy:
[...]

PR #7912, opened 2025-12-11
feat: Add quill editor support (#7755)

jhedmendoza:
Created component for quill text editor support.


jhedmendoza:
@djaiss @asbiin

Issue #5631, opened 2021-10-20
"Stay in Touch" reminders don't remind me


JonTheNiceGuy:
**Is your feature request related to a problem? Please describe.**
I have found myself missing contacting people I set "stay in touch" reminders for. While that was largely due to email issues, it would be useful to see the "upcoming dates" of stay in touch reminders.

**Describe the solution you'd like**
I would find it useful to have these either as a separate box on the dashboard ("Upcoming contact reminders"), or in the "Events in the next 3 months" box.

**Additional context**
This SQL seems useful:
```
SELECT id, DATE_ADD(stay_in_touch_trigger_date, INTERVAL stay_in_touch_frequency DAY) AS next_contact_date FROM contacts WHERE stay_in_touch_frequency > 0 AND next_contact_date < DATE_ADD(CURRENT_DATE(), INTERVAL 7 DAY);
```

Issue #6640, opened 2023-05-16
php artisan send:stay_in_touch not working


ROODAY:
**Describe the bug**
I'm not receiving emails from Monica, specifically the stay_in_touch ones but I haven't gotten any other ones either. I'm running Monica via Docker (compose), and when I get into the server with `docker exec -it monica /bin/bash` I can successfully run `php artisan monica:test-email` and receive the email, but when I run `php artisan send:stay_in_touch`, nothing happens.

**Which version are you using:**
 - Self-hosting via docker, docker image ID is `926e6860e9b2`

Issue #4842, opened 2021-02-04
"Stay in Touch", cron, and database connection issues


arfar:
**Describe the bug**
I have set up a few different "Stay in Touch" reminders which all should have triggered (I've been using Monica for about a month now and all my reminders are <1 month). Only one has been triggered however.

**Which version are you using:**
 - A server you maintain yourself 
```
www-data@monica:~/monica$ git status
HEAD detached at v2.19.1
```

I'm not using docker or anything. If you need to know the version of anything else, please ask :)

**Additional context**
I don't think there's an issue with the emailing, I can send myself test emails fine, and one of the reminders to stay in touch did come through just fine.

One of the logs showing what I _think_ could be related to the issues: https://pastebin.com/7VE0K1NX
I don't understand why I would be getting a database connection issue though, all the rest of the application works fine. Perhaps something to do with the php/cron/mysql environment not working? My cron job is also below. When I run the command in the shell, I don't get any errors.

I was able to check the crons table myself manually using the username and password from the ```.env``` file.

```
www-data@monica:~/monica$ mysql -u homestead -p monica -e 'SELECT * FROM crons;'
Enter password: 
+----+----------------------------+---------------------+---------------------+---------------------+
| id | command                    | last_run            | created_at          | updated_at          |
+----+----------------------------+---------------------+---------------------+---------------------+
|  1 | send:reminders             | 2021-02-04 17:00:02 | 2021-01-20 14:14:01 | 2021-02-04 17:00:02 |
|  2 | send:stay_in_touch         | 2021-02-04 17:00:02 | 2021-01-20 14:14:01 | 2021-02-04 17:00:02 |
|  3 | monica:calculatestatistics | 2021-02-04 00:00:03 | 2021-01-20 14:14:01 | 2021-02-04 00:00:03 |
|  4 | monica:ping                | 2021-02-04 00:00:03 | 2021-01-20 14:14:02 | 2021-02-04 00:00:03 |
|  5 | monica:clean               | 2021-02-04 00:00:04 | 2021-01-20 14:14:03 | 2021-02-04 00:00:04 |
|  6 | monica:updategravatars     | 2021-02-04 00:00:05 | 2021-01-20 14:14:03 | 2021-02-04 00:00:05 |
+----+----------------------------+---------------------+---------------------+---------------------+
```
```
www-data@monica:~/monica$ crontab -l
.
.
.
[...]

PR #7937, opened 2026-04-28
fix: Fix bug in stay in touch functionality

charliewolf:
The time in the stay_in_touch trigger date column is based on when the trigger was created in the UI. It uses the same logic as reminders to send the notification at the users preferred time of day, however it also pushes them forward if it detects it missed a reminder. However, the logic is flawed because that time comparison is done in UTC (as it is not associated with a specific user) and so if the the UTC time happens to be before the user's reminder time, the notification will continuously get pushed forward and thus never triggered.


charliewolf:
I believe this fixes  #5631 #4842 and #6640

PR #7941, opened 2026-05-07
fix: PHP 8.4 support, migration back to NPM, address npm audit warnings

ecaron:
When creating a fresh instance of Monica on a new Linux install, and multiple errors related to everything from PHP 8.4 (I think 8.3 would also have these errors) to package changes prevent the install from completing. With this PR, it is possible to securely run Monica on a 2025 (or later) Linux set. Changes include:
- Migration to Laravel 12, which brings PHP 8.4 support (inc. 8.2 support, [which had active support end 18 months ago](https://www.php.net/supported-versions.php))
- Migration from `yarn` to `npm` (I tried to stick with `yarn`, but fixing the security audit complaints was easier moving back to `npm` than fixing in yarn 2, plus `npm` has resolved the issues that lead to Monica's 2021 adoption of `yarn` in the first place)
- Moving `composer` and `package` libraries to maintained releases
- Fixing the `CloudflareDriver` to leverage `Fluent`'s pattern
- Dropped the `laravel-adorable` package, since Adorable went out of business. I didn't drop all the code related to Adorable, but would like to do that if/when this PR is approved.
- Modified the `manifest.webmanifest` so it wouldn't throw console errors
- Adjusted the `app_version` CLI output to be less noisy when running on untagged repos

**Note** Although there are changes to the database migration files, the migrations themselves are left untouched. This felt like an appropriate pattern, since it will make the migrations work for new installs that require them - but existing installs won't need to re-run them since they've already worked.


brycehans:
Hey @ecaron I have a question for you: while I'm working through your PR as reference material for [my fork](https://github.com/brycehans/monica) (4.x line, Laravel 9 for the moment (upgrading that soon), walking dep upgrades one rung at a time).

I noticed you dropped `laravolt/avatar` and replaced `Avatar::create(...)` with a hand-rolled inline-SVG generator + `getInitials()` helper. The PR body explains the `laravel-adorable` drop (Adorable went out of business) but doesn't say what drove the laravolt swap. Was it driven by a concrete laravolt-on-Laravel-12 / PHP-8.4 constraint failure, or was there some other reason?

Reason I'm asking: my fork holds a "no UI redesigns" line, and the visible behaviour of generated avatars changes between approaches (raster JPEG with laravolt's font/sizing vs. inline SVG with system sans-serif). 

I'm trying to figure out whether I need to plan a coordinated avatar-regeneration migration when I eventually hit this on my Phase 3 (Laravel 9→12) ladder, or whether laravolt is still viable on Laravel 12 if I want to defer the swap.

ecaron:
@brycehans I dropped it because PHP 8.4 support didn't happen until [`v6.1.2`](https://github.com/laravolt/avatar/releases/tag/6.1.2), and Laravel 9 support was dropped at [`v6.0.0`](https://github.com/laravolt/avatar/releases/tag/6.0.0), which made my local step upgrades more complex than seemed warranted. Like you hinted at, the package didn't feel worth the effort for the 16 lines where it is utilized in [`GenerateDefaultAvatar.php`](https://github.com/monicahq/monica/pull/7941/changes#diff-c5bc0a4b4cadc69f703bc9b497991a96a737274dc5b842d341cdc4bdef1e5607).

If you've found a way to make Laravel 9 be happy with PHP 8.4, then you probably shouldn't think too much about my design decisions. I couldn't figure out how to do one without the other (short of ignoring a lot of PHP errors).

Hope this helps!

brycehans:
Thanks mate. I managed to partially side-step it with a different sequence:

* Laravel 9 → 11 on PHP 8.1
* That made it easier to go PHP 8.1 → 8.4
* Then the laravolt window was no longer in the way

Appreciate you taking the time to dig out the version specifics! 🙏

PR #7953, opened 2026-07-02
Fix date format in British English locale

qxzkjp:
Hello friendly devs,

I noticed a bug in how dates are displayed: when setting locale to English (United Kingdom), a lot of places that show dates don't actually change format. The root cause seems to be that you are using the Locale setting from "Carbon", but when Locale is set in Carbon it "normalises" it. So eg 'en-GB' becomes 'en_GB'. But Laravel's internationalisation expects the locale name to exactly match, so it can't find 'en_GB' and falls back to 'en'.

The fix is to use `App::Locale` instead of `Carbon::Locale`. The unit tests all pass (and I added a new one to check the en-GB case). I also updated the dockerfile in this project to match monicahq/docker, so that any future devs would easily be able to create a test image. I have made an image from the code in this repo, which can be pulled from `ghcr.io/qxzkjp/monica:date-fix`. I have used it myself (it is running on my personal server) and I have tested that the UI is fixed and that there are no obvious introduced errors.

I appreciate that you are focusing all of your efforts on v5 at the moment, but I was hoping that if someone did all the coding work for you you'd have time available to merge and tag it. It's a simple fix but it'll be useful to a lot of people, and it seems that Chandler is not expected to be out of beta for a year or more.

PS for disclosure, I did use Claude Code to help make this change. I have read the code myself, and I made sure to test it. The fix was simple enough that Claude was able to do the whole task, but I promise I am a professional who is used to using AI assistance and getting good-quality output from it, not a random slop coder. I have had PRs accepted before without use of AI (admittedly years ago, I'm an infrequent contributor).

### DIFF
--- a/app/Helpers/CountriesHelper.php
+++ b/app/Helpers/CountriesHelper.php
@@ -73,7 +73,12 @@ class CountriesHelper
         $locale = App::getLocale();
         $lang = LocaleHelper::getLocaleAlpha($locale);
 
-        return $country->getTranslation($lang)['common'];
+        $translations = $country->getTranslations();
+        if (! isset($translations[$lang])) {
+            return $country->getName();
+        }
+
+        return $translations[$lang]['common'];
     }
 
     /**
--- a/app/Helpers/DateHelper.php
+++ b/app/Helpers/DateHelper.php
@@ -6,6 +6,7 @@ use Carbon\Carbon;
 use function Safe\date;
 use function Safe\strtotime;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
 
 class DateHelper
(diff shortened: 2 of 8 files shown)


qxzkjp:
I actually found a second display bug after creating the PR: when locale is set to en-GB, the country list (eg for an address input) displays in German!

This is because `CountriesHelper::getCommonNameLocale` is calling `LocaleHelper::getLocaleAlpha`, which is the only `LocaleHelper` function that does _not_ split the locale on `/-_/` to get the two-letter code. It ends up with `''` as the locale, and when the Rinvex country model can't find the locale you ask for, it returns the "first" one, which in this case happened to be "first alphabetically by ISO code". So `deu` ended up as the default for most countries.

I added this as a separate commit because it seemed like a separate bug, but I am happy to squash if that's the procedure.

